PR #12, opened 2025-12-16
make: add build target

advait:
Adds a `build` target to build ./bin/gog and makes it the default `make` goal. README updated to point to `make`.


steipete:
Thank you!

PR #51, opened 2026-01-08
fix(secrets): detect headless Linux and avoid D-Bus hangs

### LINKED ISSUE #45: Headless mode does not accept redirect url
I'm using headless mode with `--manual` flag, but when I'm pasting redirect URL back to terminal nothing happens.

salmonumbrella:
## Summary
- Detect missing `DBUS_SESSION_BUS_ADDRESS` on Linux and automatically use file backend (avoids hang entirely)
- Add 5-second timeout fallback for edge cases where D-Bus is present but unresponsive
- Provides actionable error message guiding users to `GOG_KEYRING_BACKEND=file`

## Background
Issue #45 reported that `gog auth tokens import` hangs indefinitely on headless Raspberry Pi. The root cause is the keyring library trying to connect to D-Bus SecretService when gnome-keyring is installed but not running.

## Approach
Two-layer protection:
1. **D-Bus detection (primary)**: If `DBUS_SESSION_BUS_ADDRESS` is not set on Linux with "auto" backend, force file backend. This handles the common headless case cleanly with no goroutine leaks.
2. **Timeout fallback (safety net)**: If D-Bus is present but unresponsive, timeout after 5 seconds with guidance to use file backend.

## Test plan
- [x] `TestOpenKeyring_NoDBus_ForcesFileBackend` - verifies D-Bus detection
- [x] `TestOpenKeyring_ExplicitBackend_IgnoresDBusDetection` - explicit backend bypasses detection
- [x] `TestOpenKeyringWithTimeout_Success` - timeout wrapper works
- [x] `TestOpenKeyringWithTimeout_Timeout` - timeout error message is correct
- [x] All existing tests pass

Fixes #45

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - CLI: accept `--output` alias for `--out` and `--output-dir` for Gmail attachment downloads (#47) — thanks @salmonumbrella.
 - Auth: `gog auth list --check` validates refresh tokens by exchanging for an access token.
 - Auth: manual OAuth paste accepts EOF without newline (#46) — thanks @salmonumbrella.
-- Secrets: on headless Linux (no `DBUS_SESSION_BUS_ADDRESS`), automatically use file backend to avoid D-Bus hangs; timeout fallback for edge cases (#45).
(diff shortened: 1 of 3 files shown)


steipete:
Landed on main as 088bb6c.
- Linux-only D-Bus guard + timeout gating (auto only)
- Deterministic timeout test + Windows regression guard
- Changelog updated (#51, thanks @salmonumbrella)

Tests: pnpm lint && pnpm build && pnpm test (from internal/tracking/worker)

Thanks @salmonumbrella!

PR #53, opened 2026-01-09
fix(calendar): add timezone to EventDateTime for recurring events

visionik:
## Problem

When creating a recurring event via `gog calendar create` with `--rrule`, the Google Calendar API returns:

```
Google API error (400 required): Missing time zone definition for start time.
```

This happens because Google requires an explicit `TimeZone` field on the `EventDateTime` struct for recurring events.

## Solution

Extract the timezone from the RFC3339 offset in the `--from`/`--to` values and map common offsets to IANA timezone names.

## Example

This now works:

```bash
gog calendar create primary \
  --summary 'Publix Pharmacy App Order' \
  --from '2026-01-08T11:00:00-05:00' \
  --to '2026-01-08T11:30:00-05:00' \
  --rrule 'RRULE:FREQ=DAILY;INTERVAL=28' \
  --account user@example.com
```

## Notes

- Maps common US timezone offsets (-8 to -4) to IANA names
- UTC (Z suffix or +00:00) is handled
- Unknown offsets gracefully degrade (no timezone set, non-recurring events still work)
- Added unit tests for `extractTimezone()`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - DX: `make gogcli -- ...` passes args to the CLI; add `make gogcli-help` convenience target.
 - Calendar: `gog calendar update --add-attendee ...` adds attendees without replacing existing RSVP state (#24) — thanks @salmonumbrella.
 - Calendar: add `gog calendar create|update --rrule/--reminder` for recurrence rules and custom reminders (#34) — thanks @salmonumbrella.
+- Calendar: infer IANA `timeZone` from `--from/--to` offsets for recurring events (#53) — thanks @visionik.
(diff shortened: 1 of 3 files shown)


steipete:
Landed with small follow-ups:
- Rebased onto main @8fe47fc
- Improved timeZone inference by matching candidate IANA zones at the event instant (DST-safe; adds America/Phoenix)
- Updated unit tests + CHANGELOG
- Local gate: make lint/build/test (repo has no package.json, so pnpm scripts not available)

Commits:
- 6ec670d
- 00dcff3
Merge commit on main: 6c6a354e544f818e6a15e34eb656f20365336ba7

Thanks @visionik!

PR #59, opened 2026-01-10
made show-assigned for tasks default

tompson:
I know that this is default false in the API, but I find it very strange that not all open tasks show up when you list them, IMHO you should get all open tasks per default


steipete:
Thomas, servas, I agree!

steipete:
Landed via temp rebase onto main.\n\n- Gate: make ci && make worker-ci\n- Land commit: acc1c7833358314cacad136f5c7cbd354c93231d\n- Merge commit: 0fcffcafd496ee9f9fd1ec3449541a3979696d0c\n\nThanks @tompson!

PR #83, opened 2026-01-16
feat(gmail): show attachment info in get command output

jeanregisser:
## Summary

Display attachment filename, size, MIME type, and attachment_id in both plain text and JSON output for the `gmail get` command with `--format=full`.

This makes it easier to identify and download attachments without needing to manually parse the message payload structure.

## Problem

When using `gog gmail get <messageId>`, there's no indication of attachments in the output. Users need to use `--format=full --json` and then manually parse the nested `message.payload.parts` structure to find attachment IDs before they can use `gog gmail attachment` to download them.

## Solution

Reuse the existing `collectAttachments()` function from `gmail_thread.go` to extract and display attachment info in both output formats.

### Example plain text output (new section before body):
```
from	a@example.com
to	b@example.com
subject	Email with attachment
date	Fri, 26 Dec 2025 10:00:00 +0000

attachment	CV-Document.pdf	192123	application/pdf	ANGjdJ...

hello with attachment
```

### Example JSON output (new field):
```json
{
  "attachments": [
    {"filename": "CV-Document.pdf", "size": 192123, "mimeType": "application/pdf", "attachmentId": "ANGjdJ..."}
  ]
}
```

## AI-Assisted Development

This PR was created with AI assistance. The prompts used:

**Initial prompt:**
> I wonder if the gog cli email output should be improved to include info about attachments?

**Follow-up prompt:**
> Yes it's a repo from steipete, I think we could propose a PR directly with a proposed solution given the problem we had. Can you take care of it? Give a link to the PR once done. Include the prompts used in the PR too I think it's helpful for the maintainer

**Context:** The user was trying to download an email attachment and struggled because `gog gmail get` didn't show attachment info. They had to:
1. Use `gog gmail get <id> --format=full --json`
2. Parse the JSON to find `.message.payload.parts[].body.attachmentId`
3. Then use `gog gmail attachment <messageId> <attachmentId>`

With this change, the attachment ID is visible directly in the standard output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 0.7.0 - Unreleased
 
+### Added
+
+- Gmail: show attachment info (incl. humanized sizes) for `gmail get` full/metadata output, with JSON `sizeHuman`. (#83) — thanks @jeanregisser.
+
 ### Fixed
 
 - Gmail: include `gmail.settings.sharing` scope for filter operations to avoid 403 insufficientPermissions. (#69) — thanks @ryanh-ai.
--- a/internal/cmd/gmail_get.go
+++ b/internal/cmd/gmail_get.go
@@ -27,6 +27,7 @@ const (
 type attachmentOutput struct {
 	Filename     string `json:"filename"`
 	Size         int64  `json:"size"`
+	SizeHuman    string `json:"sizeHuman"`
 	MimeType     string `json:"mimeType"`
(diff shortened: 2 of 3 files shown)


steipete:

- Gate: pnpm lint && pnpm build && pnpm test
- Land commit: 67eb0d943394e250838f01d81980b83c49b3a046
- Merge commit: 6c374ea7f1a7486b1be82b36c42b484e10dff651

Thanks @jeanregisser!

PR #104, opened 2026-01-23
feat(calendar): add calendar aliases

zacharytamas:
## Summary

- Add support for calendar aliases, allowing users to map friendly names like `family` to full calendar IDs like `3656f8abc123@group.calendar.google.com`
- New CLI commands: `gog calendar alias {list,set,unset}`
- All calendar commands now resolve aliases before using calendar IDs

## Usage

```bash
# Set an alias
gog calendar alias set family 3656f8abc123@group.calendar.google.com

# Use the alias in any calendar command
gog calendar create family --summary "Dinner" --from "2026-01-24T18:00:00Z" --to "2026-01-24T20:00:00Z"
gog calendar events family --today
gog calendar search "birthday" --calendar family

# Aliases work in comma-separated lists too
gog calendar freebusy family,work --from "..." --to "..."

# List and remove aliases
gog calendar alias list
gog calendar alias unset family
```

## Config

Aliases are stored in `~/.config/gog/config.json`:

```json
{
  "calendar_aliases": {
    "family": "3656f8abc123@group.calendar.google.com"
  }
}
```

## Test plan

- [x] Unit tests for config CRUD operations
- [x] Unit tests for CLI commands
- [x] Unit tests for alias resolution helper
- [x] All existing tests pass

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -19,6 +19,7 @@
 
 - Formatting: `make fmt` (`goimports` local prefix `github.com/steipete/gogcli` + `gofumpt`).
 - Output: keep stdout parseable (`--json` / `--plain`); send human hints/progress to stderr.
+- Gmail labels: treat label IDs as case-sensitive opaque tokens; only case-fold label names for name lookup.
 
 ## Testing Guidelines
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,62 @@
 # Changelog
 
+## 0.12.0 - Unreleased
+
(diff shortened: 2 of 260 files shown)


visionik:
Hi Zachary - 

Great idea.

Two things:

1. Your PR has a CI failure check (see above).  Can you fix and update?

2. I am trying to decide if we should merge this as is, or create a more holistic alias system so you can use aliases anywhere in gog.  For eample

gog docs family-history sed ...

Thoughts?

PR #107, opened 2026-01-25
feat(drive): add --all flag to ls for global file listing

### LINKED ISSUE #108: feat(drive): add --all flag to ls for global file listing
## Problem

Currently `gog drive ls` always scopes results to a specific folder (defaulting to root via `'root' in parents` query). To list files globally across all of Drive, users must use `drive search` which requires a search query string.

This makes it cumbersome to simply list recent files across all accessible files without knowing a specific search term.

## Proposed Solution

Add an `--all` (or `--global`) flag to `drive ls` that omits the parent folder constraint, allowing users to list all accessible files.

## Example Usage

```bash
# List all accessible files (not just root folder)
gog drive ls --all --max 20

# Combine with query filter  
gog drive ls --all --query "mimeType='application/pdf'"
```

## Implementation

PR: #107

struong:
## Summary

Adds an `--all` (alias `--global`) flag to `gog drive ls` that lists all accessible files instead of being scoped to a specific folder.

Closes #108

## Motivation

Currently `drive ls` always filters by parent folder (defaulting to root). To list files globally, users must use `drive search` which requires a search query. This makes it cumbersome to simply list recent files across all of Drive.

## Changes

- Added `All bool` field to `DriveLsCmd` with `--all/--global` flags
- Added `buildDriveAllListQuery()` that builds a query without parent folder constraint
- Updated `DriveLsCmd.Run()` to use the appropriate query builder based on the flag
- Added unit tests for the new function

## Usage

```bash
# List all accessible files (not just root folder)
gog drive ls --all --max 20

# Combine with query filter
gog drive ls --all --query "mimeType='application/pdf'"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Sheets: add `sheets insert` to insert rows/columns into a sheet. (#203) — thanks @andybergon.
 - Gmail: add `watch serve --history-types` filtering (`messageAdded|messageDeleted|labelAdded|labelRemoved`) and include `deletedMessageIds` in webhook payloads. (#168) — thanks @salmonumbrella.
 - Contacts: support `--org`, `--title`, `--url`, `--note`, and `--custom` on create/update; include custom fields in get output with deterministic ordering. (#199) — thanks @phuctm97.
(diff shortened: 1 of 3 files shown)


steipete:

- Gate: make ci
- Land commit: ced43a5a705783a8296e5dd8e876d52e67d42849
- Merge commit: 9eb42121e9b26f2401f9ec3293c7783c2e8e1e90

Thanks @struong!

PR #173, opened 2026-02-03
feat(cmd): add --disable-commands flag for command denylist

spookyuser:
Add support for blocking specific commands using dot-notation (e.g., gmail.send, calendar.delete).
- Add GOG_DISABLE_COMMANDS env var
- Block commands by exact match or parent prefix
- Case-insensitive matching


steipete:
After comparing #173, #218, and #366: #173 and #218 overlap heavily in the same runtime policy layer, while #366 is the stronger compile-time safety track. If we keep one runtime PR, I would keep #218 over #173: it extends the existing `--enable-commands` path, stays least-privilege, and avoids extra denylist precedence/docs/test complexity. #173 feels superseded unless we explicitly want both allowlist and denylist semantics.

steipete:
Thanks @spookyuser! Landed command denylist support on main in d84148a, sharing the same dotted command-path matcher as the allowlist. Closing this PR as landed.

PR #203, opened 2026-02-08
feat(sheets): add insert subcommand for rows and columns

### LINKED ISSUE #202: Feature(sheets): Insert rows at a specific position
## Problem

There's no way to insert rows at a specific position in a sheet to shift existing data down.

`gog sheets append` creates new rows but always at the bottom of existing data. `gog sheets update` writes to existing cells in place without shifting anything. There's no command that pushes existing rows down to make room.

## Use case

When maintaining a sheet sorted by date (newest first), new entries need to go after the header row, not at the bottom. Currently the only option is to `append` at the bottom and manually re-sort, or orchestrate a read-shift-write cycle externally.

## Proposed solution

Add `gog sheets insert-rows <spreadsheetId> <sheet> <startRow> [--count N] [--after]` using the Sheets API `InsertDimensionRequest` via `batchUpdate`. The API already supports this — it's just not exposed in the CLI.

andybergon:
Closes #202

## Summary

Adds `gog sheets insert <spreadsheetId> <sheet> <rows|cols> <start> [--count N] [--after]` to insert empty rows or columns at a specific position in a sheet.

Uses the Sheets API `InsertDimensionRequest` via `batchUpdate`, following the same pattern as `format` and `copy-validation-from`.

## Usage

```bash
# Insert 3 empty rows before row 2
gog sheets insert <id> "Sheet1" rows 2 --count 3

# Insert 1 column after column 3
gog sheets insert <id> "Sheet1" cols 3 --after
```

## New flags

| Flag | Description | Default |
|------|-------------|---------|
| `--count` | Number of rows/columns to insert | `1` |
| `--after` | Insert after the specified position instead of before | `false` |

## Testing

- Unit tests with httptest mock server covering rows before/after, cols before/after, and invalid dimension
- `make ci` passes (fmt, lint, all tests)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 0.12.0 - Unreleased
 
+### Added
+- Sheets: add `sheets insert` to insert rows/columns into a sheet. (#203) — thanks @andybergon.
+
 ### Fixed
 - Calendar: respond patches only attendees to avoid custom reminders validation errors. (#265) — thanks @sebasrodriguez.
 
--- a/internal/cmd/sheets_insert.go
+++ b/internal/cmd/sheets_insert.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"context"
-	"fmt"
 	"os"
 	"strings"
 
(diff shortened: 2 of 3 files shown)


andybergon:
This PR is ready for review.

Note: I squashed and force-pushed — the initial version only had `insert-rows`, but since the Sheets API `InsertDimensionRequest` supports both rows and columns via the `Dimension` field, I consolidated it into a single `insert` subcommand that handles both with a `rows`/`cols` argument.

steipete:

- Gate: make ci
- Land commit: 039e2963244cebd36355b3737a7987dee12632e9
- Merge commit: 039e2963244cebd36355b3737a7987dee12632e9

Thanks @andybergon!

Issue #110, opened 2026-01-26
feat(gmail): Add --gmail-scope=readonly option for gmail.readonly scope


a-b13:
## Problem

Currently, the Gmail service always requests the `gmail.modify` scope, which grants read-write access. There's no way to request read-only access to Gmail.

For security-conscious setups where an AI assistant or automation tool needs to *read* emails but should not be able to *send* or *modify* them, this is a significant limitation.

## Desired Behavior

Add a `--gmail-scope` flag similar to `--drive-scope`:

```bash
gog auth add user@gmail.com --services=gmail --gmail-scope=readonly
```

This would request `gmail.readonly` scope instead of `gmail.modify`.

## Google's Available Scopes

Google supports these Gmail scopes:
- `gmail.readonly` — Read-only access to messages and settings
- `gmail.modify` — Read/write access (current default)
- `gmail.compose` — Send only
- `gmail.send` — Send only

Reference: https://developers.google.com/gmail/api/auth/scopes

## Use Case

I'm using gog with an AI assistant that needs to search and read emails but should never be able to send or delete messages. Currently I have to grant full read-write access, which is a security risk.

Thank you for this excellent tool!

PR #218, opened 2026-02-10
feat: subcommand-level allowlist for --enable-commands

EricYangTL:
Extend --enable-commands to support dotted subcommand paths (e.g., gmail.search, gmail.drafts.create) in addition to top-level commands.

Top-level entries like 'gmail' remain backward-compatible and allow all subcommands. Dotted entries restrict to specific subcommand paths. My specific use case here is to enable gmail **drafts** but not **sends**, which is not available with top-level allowlist.

Examples:
  --enable-commands='gmail.search,gmail.drafts.create,calendar'
  --enable-commands='gmail.search,gmail.get,gmail.labels'

Refs #110 (related: per-service scope control for agent sandboxing)


steipete:
Thanks @EricYangTL! Landed dotted command-path allowlist support on main in d84148a, with shared matcher coverage and docs. Closing this PR as landed.

PR #228, opened 2026-02-12
fix: recurring timezone fallback, attachment hardening, transport defaults

### LINKED ISSUE #216: Recurring Events Fail with "Missing time zone definition" Error
### Environment

• gog version: 0.9.0 (Homebrew 2026-01-22)
• OS: Linux (Ubuntu/Debian-based)
• Installation: Homebrew

### Issue

Creating recurring calendar events fails with Google API error: Missing time zone definition for start time, even though single (non-recurring) events work fine with the same datetime format.

### Steps to Reproduce

1. Single event (works fine):
export GOG_ACCOUNT="user@example.com"
gog calendar create primary \
  --summary "Test Event" \
  --from "2026-02-13T08:00:00+02:00" \
  --to "2026-02-13T09:00:00+02:00"
✅ Result: Event created successfully

2. Recurring event (fails):
export GOG_ACCOUNT="user@example.com"
gog calendar create primary \
  --summary "Recurring Test" \
  --from "2026-02-13T08:00:00+02:00" \
  --to "2026-02-13T09:00:00+02:00" \
  --rrule "FREQ=WEEKLY;BYDAY=TU,TH" \
  --reminder "popup:30m"
❌ Result: Google API error (400 required): Missing time zone definition for start time.

### Expected Behavior

Recurring events should work with the same datetime format that works for single events, especially since PR #53 (v0.5.0) fixed timezone handling for recurring events.

### Actual Behavior

Recurring events fail even though:

• The datetime includes explicit timezone offset (+02:00)
• Single events work with the same format
• The fix was supposedly applied in v0.5.0
Workaround Attempts

Tried multiple datetime formats, all failed:

• 2026-02-13T08:00:00+02:00 (with offset)
• 2026-02-13T08:00:00 (without offset, relying on system TZ)
• Setting TZ=Africa/Johannesburg environment variable
All single-event variants work; all recurring variants fail.

### Related

• PR #53 supposedly fixed this in v0.5.0
• Possible regression in later versions?
• Google Calendar API docs require explicit timeZone field for recurring events: https://developers.google.com/workspace/calendar/api/guides/recurringevents
Additional Context

The Google Calendar API example for recurring events shows:
{
  "start": {
    "dateTime": "2011-06-03T10:00:00.000-07:00",
    "timeZone": "America/Los_Angeles"
  },
  "recurrence": [
    "RRULE:FREQ=WEEKLY;UNTIL=20110701T170000Z"
  ]
}
It seems the API requires both the dateTime with offset AND an explicit timeZone field for recurring events, which might not be set when using --rrule.

### Request

Could you verify if the timezone inference logic from PR #53 is still being applied when --rrule is present? There might be a code path that bypasses the fix for recurring events.

### LINKED ISSUE #217: Bug: gog drive delete fails with 404 on Shared Drive files (missing supportsAllDrives=true)
# Bug Report: `gog drive delete` fails with 404 on Shared Drive files

## Environment
- gog version: v0.9.0
- Platform: Debian 12 (Linux)
- Google Account type: Google Workspace (with Shared Drives)

## Description

`gog drive delete <fileId>` fails with a 404 error when deleting files from Google Shared Drives, even when the user has `fileOrganizer` (Content Manager) permissions which include delete rights.

## Root Cause

The Google Drive API requires the `supportsAllDrives=true` query parameter for all operations on Shared Drive files. The `gog drive delete` command does not set this parameter, causing the API to only search in the user's personal "My Drive" and return a 404 error when the file exists only in a Shared Drive.

## Steps to Reproduce

1. Set up a Google Shared Drive with a test account having "Content Manager" role (fileOrganizer)
2. Upload a file to the Shared Drive root:
   ```bash
   gog drive upload test.txt --shared-drive "MySharedDrive" --account user@example.com
   ```
3. Note the returned file ID (e.g., `1jGHQ8uydpRQrJHSgthRyz9_9fMeDHdia`)
4. Verify the file exists:
   ```bash
   gog drive get 1jGHQ8uydpRQrJHSgthRyz9_9fMeDHdia --account user@example.com --shared-drive "MySharedDrive"
   # Returns: file metadata successfully
   ```
5. Try to delete it:
   ```bash
   gog drive delete 1jGHQ8uydpRQrJHSgthRyz9_9fMeDHdia --account user@example.com --shared-drive "MySharedDrive" --force
   ```

## Expected Behavior

File should be moved to trash (or permanently deleted if `--permanent` flag exists).

## Actual Behavior

```
Google API error (404 notFound): File not found: 1jGHQ8uydpRQrJHSgthRyz9_9fMeDHdia.
```

## Verification

The file DOES exist - confirmed via:
1. `gog drive get <fileId>` returns full metadata
2. Google Drive web UI shows the file
3. `gog drive permissions <fileId>` shows user has `fileOrganizer` role

## Workaround

Move the file to personal "My Drive" first, then delete:
```bash
# Move to My Drive (parentId: null = root of My Drive)
gog drive move 1jGHQ8uydpRQrJHSgthRyz9_9fMeDHdia --parent <myDriveRootId> --account user@example.com

# Delete from My Drive (works without supportsAllDrives)
gog drive delete 1jGHQ8uydpRQrJHSgthRyz9_9fMeDHdia --account user@example.com --force
```

## Fix Needed

The `delete` command should add `supportsAllDrives=true` query parameter when making the API call, similar to how other commands (upload, move, etc.) handle Shared Drives.

**API Documentation:**
https://developers.google.com/drive/api/guides/enable-shareddrives#make_an_api_request

Relevant excerpt:
> To access shared drives using the Drive API, add the `supportsAllDrives=true` query parameter to all requests to the Drive API that include a file ID parameter.

## Additional Context

Other `gog drive` commands correctly work with Shared Drives when `--shared-drive` flag is used:
- ✅ `gog drive upload` - works
- ✅ `gog drive move` - works
- ✅ `gog drive rename` - works
- ✅ `gog drive copy` - works
- ✅ `gog drive get` - works
- ❌ `gog drive delete` - fails with 404

This suggests the delete command is missing the `supportsAllDrives=true` parameter that other commands include.

### LINKED ISSUE #223: gmail attachment: caching returns false hits for directories and unrelated files
## Problem

The attachment download caching in `downloadAttachmentToPath` has two issues that combine to produce false cache hits:

### 1. Cache accepts directories as valid cached files

`os.Stat(outPath)` succeeds on directories, and since `expectedSize == -1` the check is just `st.Size() > 0`. A directory always has `Size() > 0` (typically 64-512 bytes depending on filesystem), so it passes the cache check and returns `cached: true`.

### 2. Cache doesn't validate content — any non-empty file is a "hit"

When `expectedSize == -1` (which is always the case for `GmailAttachmentCmd.Run`), the cache check is:
```go
if st, err := os.Stat(outPath); err == nil && st.Size() > 0 {
    return outPath, true, st.Size(), nil
}
```

This means if **any** file exists at the output path (from a previous failed download, a different attachment, or even an unrelated file), it's returned as a cached result without any content validation.

## Reproduction

```bash
# Write a small unrelated file to the output path
echo "not a PDF" > /tmp/invoice.pdf

# Try to download an attachment to the same path
gog gmail attachment <messageId> <attachmentId> --out /tmp/invoice.pdf
# Output: path=/tmp/invoice.pdf cached=true bytes=11
# Expected: should re-download the actual attachment (648KB PDF)
```

## Suggested fix

In `downloadAttachmentToPath`:

1. Check `st.Mode().IsRegular()` to reject directories
2. When `expectedSize > 0` is available, pass it through from the Gmail API metadata (the `sizeEstimate` field on the attachment) so the size check can reject wrong-sized files
3. Consider adding a `--no-cache` / `--force-download` flag for programmatic usage where cache correctness matters

At minimum, checking for regular files would prevent the most confusing case (directories being treated as cached attachments).

---
*Filed by an AI agent on behalf of [@schickling](https://github.com/schickling)*

salmonumbrella:
## Summary
- **Calendar**: Fall back to `Etc/GMT±N` timezone for recurring events when no regional IANA timezone matches the UTC offset (fixes offset datetimes losing timezone info)
- **Gmail**: Harden attachment output path resolution and cache validation — sanitize filenames against path traversal, require exact size match for cache hits, add `IsRegular()` check
- **Google API**: Clone `http.DefaultTransport` instead of bare `http.Transport{}` to preserve proxy settings and connection pooling; add `supportsAllDrives` to drive delete
- **Tests**: Add unit tests for `sanitizeAttachmentFilename` and `resolveAttachmentOutputPath` with adversarial inputs (traversal, dotdot, trailing separators)

Fixes #216
Fixes #217
Fixes #223

## Test plan
- [x] `go test ./...` passes (all packages green)
- [x] Calendar Etc/GMT fallback tested with +02:00, -04:00, and half-hour offsets
- [x] Attachment path traversal cases covered (`../../etc/passwd`, `../secret.txt`, absolute paths)
- [x] Cache validation tested: exact size match, stale file re-download, directory not treated as cache hit
- [x] Transport proxy propagation and TLS 1.2 minimum verified

### DIFF
--- a/internal/cmd/execute_gmail_attachment_test.go
+++ b/internal/cmd/execute_gmail_attachment_test.go
@@ -408,7 +408,7 @@ func TestExecute_GmailAttachment_StaleFileIsRedownloaded(t *testing.T) {
 	newGmailService = func(context.Context, string) (*gmail.Service, error) { return svc, nil }
 
 	outPath := filepath.Join(t.TempDir(), "invoice.pdf")
-	if err := os.WriteFile(outPath, []byte("stale"), 0o600); err != nil {
+	if writeErr := os.WriteFile(outPath, []byte("stale"), 0o600); writeErr != nil {
 		t.Fatalf("WriteFile: %v", err)
 	}
 
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
+	"strconv"
 
 	"github.com/alecthomas/kong"
 
@@ -162,10 +163,7 @@ func envOr(key, fallback string) string {
 }
 
 func boolString(v bool) string {
-	if v {
-		return "true"
-	}
-	return "false"
+	return strconv.FormatBool(v)
 }
 
 func newParser(description string) (*kong.Kong, *CLI, error) {
--- a/internal/googleapi/client_more_test.go
+++ b/internal/googleapi/client_more_test.go
@@ -268,24 +268,29 @@ func TestNewBaseTransport_RespectsProxyAndTLSMinimum(t *testing.T) {
 	if transport == nil {
 		t.Fatalf("expected transport")
 	}
+
 	if transport.Proxy == nil {
 		t.Fatalf("expected proxy func")
 	}
+
 	if transport.TLSClientConfig == nil {
 		t.Fatalf("expected TLS config")
 	}
+
 	if transport.TLSClientConfig.MinVersion < tls.VersionTLS12 {
 		t.Fatalf("expected TLS min version >= 1.2, got %d", transport.TLSClientConfig.MinVersion)
 	}
 
-	req, err := http.NewRequest(http.MethodGet, "https://www.googleapis.com", nil)
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodGet, "https://www.googleapis.com", nil)
 	if err != nil {
 		t.Fatalf("new request: %v", err)
 	}
+
 	proxyURL, err := transport.Proxy(req)
 	if err != nil {
 		t.Fatalf("proxy lookup: %v", err)
 	}
+
 	if proxyURL == nil || !strings.Contains(proxyURL.String(), "127.0.0.1:8888") {
 		t.Fatalf("expected HTTPS proxy to be honored, got: %v", proxyURL)
 	}


steipete:
Landed via #251.

steipete:
Landed via #251 (merge commit 1a125efbb3691d2b7c723f6e2ff527cf43ea46f0). Thanks! Closing this PR since it was merge-conflicting.

PR #258, opened 2026-02-14
feat(drive): add --no-all-drives to ls/search

steipete:
Adds `--no-all-drives` to `drive ls` and `drive search` (default stays all drives).

Also
- centralize List() shared-drive params
- validate `drive download --format` earlier
- docs/spec + README + changelog

Test plan
- make ci


steipete:

- Gate: make ci
- Land commit: 42cfe13376c89f2933037024170a918996f4e6fa
- Merge commit: ae103ced0792dcf04404ebad1cb96acae8a66c43

PR #260, opened 2026-02-14
fix(drive,gmail): pass-through filter queries and encode non-ASCII headers

### LINKED ISSUE #254: UTF-8 double-encoding in From header display name
## Description

When sending an email via `gog gmail send --from`, the display name from the Gmail send-as alias gets double-encoded in the From header if it contains non-ASCII characters.

## Steps to Reproduce

1. Have a send-as alias with a display name containing UTF-8 characters, e.g. `Sérgio Bastos • Importrust`
2. Send an email:
```bash
gog gmail send \
  --account user@gmail.com \
  --from alias@domain.com \
  --to recipient@example.com \
  --subject "Test" \
  --body "Test email"
```

## Expected

From header displays: `Sérgio Bastos • Importrust <alias@domain.com>`

## Actual

From header displays: `SÃƒÂ©rgio Bastos Ã¢Â€Â¢ Importrust <alias@domain.com>`

This is a classic UTF-8 double-encoding issue - the bytes of the UTF-8 string are being interpreted as Latin-1/ISO-8859-1 and then re-encoded as UTF-8.

## Environment

- gogcli version: v0.10.0
- OS: macOS (Darwin, arm64)

## Workaround

Changing the display name to ASCII-only characters via `gog gmail sendas update` is not viable as it requires domain-wide delegation (service account), returning a 403 error for regular OAuth tokens.

Currently there is no workaround other than changing the display name in Gmail web to remove non-ASCII characters.

### LINKED ISSUE #255: gog drive search returns folders but not files — gog drive ls works correctly
## Problem

`gog drive search` returns folders successfully but doesn't return any files, regardless of query syntax.

## Steps to Reproduce

```bash
# Works: returns folders
gog drive search "type:folder" --max 10

# Fails: returns "No results" (even though files exist)
gog drive search "mimeType != 'application/vnd.google-apps.folder'" --max 10
gog drive search "mimeType = 'application/vnd.google-apps.document'" --max 10

# Workaround: gog drive ls works correctly
gog drive ls --parent <folder-id> --max 10
# → Returns both folders AND files ✅
```

## Expected Behavior

`gog drive search` should return files when the query matches files.

## Actual Behavior

`gog drive search` always returns "No results" for file queries, even when files exist in Drive.

## Environment

- gog version: v0.9.0 (99d9575)
- macOS (Darwin 25.2.0 arm64)
- OAuth scopes: `drive`, `docs`, `calendar`, `gmail`, `sheets`, `contacts`
- Auth method: File keyring backend (`GOG_KEYRING_BACKEND=file`)

## Workaround

Use `gog drive ls --parent <folder-id>` instead of `gog drive search`.

## Additional Context

- `gog docs cat <doc-id>` works correctly for accessing files directly by ID
- `gog drive ls` correctly lists both folders and files
- The Drive API OAuth scope is correctly configured (full `drive` access)

Thanks for the excellent tool! 🙏

salmonumbrella:
## Summary
- **Drive search** now detects raw API filter syntax (field comparisons, `contains`, `in` membership, `has` property, `sharedWithMe`) and passes it through verbatim instead of wrapping in `fullText contains`. Plain-text searches always append `trashed = false` to prevent false positives when the search term contains the word "trashed".
- **Gmail MIME** now RFC 2047-encodes display names with non-ASCII characters in `From`/`To`/`Cc`/`Bcc`/`Reply-To` headers using `net/mail`, fixing garbled headers.

Fixes #254, fixes #255

## Test plan
- [x] Unit tests for `buildDriveSearchQuery` (plain text, filter pass-through, trashed already present, plain text containing "trashed")
- [x] Table-driven tests for `looksLikeDriveFilterQuery` (28 cases: filter syntax + natural language)
- [x] Integration test with httptest server for filter query pass-through
- [x] Tests for `formatAddressHeader` (UTF-8 display name, plain address, unparseable fallback)
- [x] Tests for `formatAddressHeaders` (mixed addresses, empty value filtering)
- [x] Full test suite passes (`go test ./...`)

### DIFF
--- a/internal/cmd/drive.go
+++ b/internal/cmd/drive.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"google.golang.org/api/drive/v3"
@@ -20,6 +21,14 @@ import (
 )
 
 var newDriveService = googleapi.NewDrive
+
+var (
+	driveSearchFieldComparisonPattern = regexp.MustCompile(`(?i)\b(?:mimeType|name|fullText|trashed|starred|modifiedTime|createdTime|viewedByMeTime|visibility)\b\s*(?:!=|<=|>=|=|<|>)`)
+	driveSearchContainsPattern        = regexp.MustCompile(`(?i)\b(?:name|fullText)\b\s+contains\s+'`)
+	driveSearchMembershipPattern      = regexp.MustCompile(`(?i)'[^']+'\s+in\s+(?:parents|owners|writers|readers)`)
+	driveSearchHasPattern             = regexp.MustCompile(`(?i)\b(?:properties|appProperties)\b\s+has\s+\{`)
+	driveTrashedPattern               = regexp.MustCompile(`(?i)\btrashed\b`)
+)
 
 const (
 	driveMimeGoogleDoc     = "application/vnd.google-apps.document"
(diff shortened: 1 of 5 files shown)


steipete:
Landed via #266 (had to resolve cross-repo conflicts). Thanks @salmonumbrella.

PR #264, opened 2026-02-14
feat: add Google Ads as an auth service

ufkhan97:
Adds `ads` as a new service for `gog auth add`, enabling OAuth tokens with the `https://www.googleapis.com/auth/adwords` scope.

## What

- New `ServiceAds` in the services list
- `gog auth add user@example.com --services ads` obtains a token with the Google Ads API scope
- Auth-only for now — no CLI commands, but the refresh token works with any Google Ads API client

## Why

Google Ads API requires OAuth with the adwords scope. Currently there's no way to get this through gog, so users managing Google Workspace + Ads for the same account need a separate auth flow. This lets them use gog's existing auth infrastructure.

## Changes

- `internal/googleauth/service.go`: Add `ServiceAds` constant, scope mapping, and switch case
- `internal/googleauth/service_test.go`: Update test expectations

14 lines added, 4 changed. Tests pass.

### DIFF
--- a/internal/googleauth/service.go
+++ b/internal/googleauth/service.go
@@ -25,6 +25,7 @@ const (
 	ServiceAppScript Service = "appscript"
 	ServiceGroups    Service = "groups"
 	ServiceKeep      Service = "keep"
+	ServiceAds       Service = "ads"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


steipete:
What is the point of this? Not quite sure I get it?

spa5k:
Hi @steipete can we get this please? need to do things with Google Ads and since i dont have any exp with it, Agents can do much better

steipete:
Thanks! Landed manually on main as c4665c4, with tests, docs, CHANGELOG thank-you, and co-author trailer. Closing this PR since the Google Ads auth service is now on main.

PR #265, opened 2026-02-15
fix: only patch Attendees in calendar respond to avoid reminders vali…

sebasrodriguez:
…dation error

When responding to a calendar event, the previous code was patching the entire event object including Reminders. This caused a Google API error 'Missing override reminder minutes' when the event had custom reminders.

The fix creates a minimal patch object containing only the Attendees field, avoiding the reminders validation issue.

Fixes events with custom reminders failing to accept RSVP responses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 0.12.0 - Unreleased
 
+### Fixed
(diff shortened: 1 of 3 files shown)


steipete:
Landed via local gate + squash merge onto main.

- Gate: make ci
- PR head commit: 4bf3829
- Merge commit: a2e4867

Thanks @sebasrodriguez!

PR #269, opened 2026-02-15
fix(secrets): recognize empty GOG_KEYRING_PASSWORD in non-TTY environments

### LINKED ISSUE #268: File backend: empty GOG_KEYRING_PASSWORD not recognized in non-TTY environments
## Bug Description

When using the file keyring backend with an **empty passphrase** (`GOG_KEYRING_PASSWORD=""`), gog fails in non-TTY environments with:

```
no TTY available for keyring file backend password prompt; set GOG_KEYRING_PASSWORD
```

This affects CI, cron jobs, bots, and any headless automation using encrypted tokens set up with an empty passphrase.

## Root Cause

In `internal/secrets/store.go`, the `fileKeyringPasswordFuncFrom()` function checks:

```go
if password != "" {
    return keyring.FixedStringPrompt(password)
}
```

When `GOG_KEYRING_PASSWORD=""` (empty string), this condition is `false`, so it falls through to require a TTY instead of using the empty password.

## Suggested Fix

Use `os.LookupEnv()` to check if the env var is **set** (regardless of value):

```go
func fileKeyringPasswordFunc() keyring.PromptFunc {
    password, isSet := os.LookupEnv(keyringPasswordEnv)
    if isSet {
        return keyring.FixedStringPrompt(password)  // works even if password is ""
    }
    
    if term.IsTerminal(int(os.Stdin.Fd())) {
        return keyring.TerminalPrompt
    }
    
    return func(_ string) (string, error) {
        return "", fmt.Errorf("%w; set %s", errNoTTY, keyringPasswordEnv)
    }
}
```

## Reproduction

1. Set up gog with file backend and empty passphrase (via TTY prompt, just press Enter)
2. In a non-TTY context: `GOG_KEYRING_PASSWORD="" gog auth list`
3. Error: `no TTY available for keyring file backend password prompt; set GOG_KEYRING_PASSWORD`

## Environment

- gog v0.11.0 (and v0.9.0)
- Linux arm64 (headless server)
- File keyring backend with empty passphrase

## Workaround

Use an `expect` wrapper to simulate TTY and send empty password:

```bash
#!/usr/bin/expect -f
spawn /usr/local/bin/gog-real {*}$argv
expect { "passphrase" { send "\r"; exp_continue } eof }
catch wait result
exit [lindex $result 3]
```

Thanks for the awesome tool! 🙏

zerone0x:
## Summary
Fixes #268

Use `os.LookupEnv` to distinguish between "env var not set" and "env var set to empty string" when resolving the file keyring password. Previously `fileKeyringPasswordFuncFrom` checked `password != ""`, which caused an empty passphrase (`GOG_KEYRING_PASSWORD=""`) to fall through to the TTY prompt path, failing in headless/CI environments.

## Changes
- `fileKeyringPasswordFuncFrom` now accepts an `isSet` bool parameter instead of relying on the password value
- `fileKeyringPasswordFunc` uses `os.LookupEnv` instead of `os.Getenv`
- Added test case for empty password with `isSet=true`

---
🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 - Calendar: respond patches only attendees to avoid custom reminders validation errors. (#265) — thanks @sebasrodriguez.
+- Secrets: respect empty `GOG_KEYRING_PASSWORD` (treat set-to-empty as intentional; avoids headless prompts). (#269) — thanks @zerone0x.
 
 ## 0.11.0 - 2026-02-15
 
--- a/internal/secrets/store.go
+++ b/internal/secrets/store.go
@@ -119,8 +119,9 @@ func wrapKeychainError(err error) error {
 	return err
 }
 
-func fileKeyringPasswordFuncFrom(password string, isSet bool, isTTY bool) keyring.PromptFunc {
-	if isSet {
+func fileKeyringPasswordFuncFrom(password string, passwordSet bool, isTTY bool) keyring.PromptFunc {
+	// Treat "set to empty string" as intentional; empty passphrase is valid.
+	if passwordSet {
 		return keyring.FixedStringPrompt(password)
 	}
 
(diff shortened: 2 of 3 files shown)


steipete:
Landed via temp merge + squash onto main.

- Gate: make ci
- Land commit: 7a4ca51589234f7061eb4206c7ffa0701eae662b
- Merge commit: 3a6dae30d49cf9fb121e188091bbd1390d056ea3

Thanks @zerone0x!

PR #295, opened 2026-02-16
feat(tracking): add key rotation and worker abuse protections

### LINKED ISSUE #293: Security: Add key rotation mechanism for email tracking encryption
## Summary

The email tracking system uses a single AES-256-GCM key (`TRACKING_KEY`) that is set once during deploy and has no rotation mechanism. A compromised key allows decrypting all past and future tracking pixel payloads (recipient email, subject hash, timestamp).

## Affected Files

- `internal/tracking/crypto.go`
- `internal/tracking/deploy.go` (lines 87-88)

## Current Code

**crypto.go** — Single key, no key ID or versioning:
```go
func Encrypt(payload *PixelPayload, keyBase64 string) (string, error) {
    key, err := base64.StdEncoding.DecodeString(keyBase64)
    // ... single key used for all encryptions
    ciphertext := aead.Seal(nonce, nonce, plaintext, nil)
    return base64.RawURLEncoding.EncodeToString(ciphertext), nil
}
```

**deploy.go** — Key set once as a Cloudflare secret with no rotation:
```go
runWranglerCommand(ctx, workerDir, 
    strings.NewReader(opts.TrackingKey+"\n"), 
    "secret", "put", "TRACKING_KEY", "--name", opts.WorkerName)
```

## Risk

1. **No forward secrecy**: A compromised key decrypts all historical tracking IDs (every pixel URL ever generated)
2. **No rotation path**: Rotating the key invalidates all outstanding tracking pixels in already-delivered emails
3. **Key exposure**: The key is passed via stdin to wrangler — it could appear in process listings or shell history

## Remediation

1. **Add a key version prefix** to the encrypted blob:
   ```
   <key-version-byte><nonce><ciphertext>
   ```

2. **Support multiple active keys** in the worker:
   ```typescript
   // Worker environment
   TRACKING_KEY_V1: "base64..."
   TRACKING_KEY_V2: "base64..."
   TRACKING_CURRENT_KEY_VERSION: "2"
   
   // Encrypt with current version, decrypt tries all versions
   ```

3. **Add a `tracking key-rotate` command** that:
   - Generates a new key
   - Deploys it as the next version
   - Updates the current version pointer
   - Old keys remain for decrypting historical tracking pixels

4. **Add key expiry warnings** in the CLI when keys are older than a configurable threshold.

## References

- [NIST Key Management Guidelines](https://csrc.nist.gov/publications/detail/sp/800-57-part-1/rev-5/final) — recommends periodic key rotation for symmetric keys

### LINKED ISSUE #294: Security: Add rate limiting to email tracking Cloudflare Worker
## Summary

The email tracking pixel endpoint (`/p/:blob.gif`) has no rate limiting, allowing enumeration attacks and resource abuse. The bot detection mechanism relies solely on User-Agent string matching and timing heuristics, which are trivially bypassed.

## Affected Files

- `internal/tracking/worker/src/index.ts` (lines 40-97)
- `internal/tracking/worker/src/bot.ts`

## Current Code

**index.ts** — No rate limiting on pixel endpoint:
```typescript
async function handlePixel(request: Request, env: Env, path: string): Promise<Response> {
  const blob = path.slice(3, -4);
  const key = await importKey(env.TRACKING_KEY);
  // ... decrypts and records open with no rate check
  // Every request hits D1 database INSERT
}
```

**bot.ts** — Detection based on User-Agent strings only:
```typescript
export function detectBot(userAgent: string, ip: string, timeSinceDeliveryMs: number | null): BotDetectionResult {
  if (userAgent.includes('GoogleImageProxy')) return { isBot: false, botType: 'gmail_proxy' };
  if (userAgent.includes('Outlook-iOS') || userAgent.includes('Microsoft Outlook')) return { isBot: true, botType: 'outlook_prefetch' };
  if (userAgent.includes('Barracuda') || userAgent.includes('Symantec') || userAgent.includes('Proofpoint')) return { isBot: true, botType: 'security_scanner' };
  // Trivially bypassed by setting any other User-Agent
  return { isBot: false, botType: null };
}
```

## Risk

1. **Tracking ID enumeration**: An attacker can brute-force or replay tracking pixel URLs to inflate open counts, polluting analytics
2. **D1 database abuse**: Each pixel request triggers a D1 INSERT — no deduplication means repeated requests fill the database
3. **Resource exhaustion**: High-volume requests against the worker endpoint consume Cloudflare Workers resources
4. **Bot detection bypass**: Setting `User-Agent: Mozilla/5.0` bypasses all bot detection

## Remediation

1. **Add Cloudflare Rate Limiting** via `wrangler.toml`:
   ```toml
   # Or use the Rate Limiting API in the worker
   ```

2. **Deduplicate opens** — Only record the first open per tracking_id per IP within a time window:
   ```typescript
   const existing = await env.DB.prepare(
     'SELECT 1 FROM opens WHERE tracking_id = ? AND ip = ? AND opened_at > datetime("now", "-1 hour")'
   ).bind(blob, ip).first();
   if (existing) return pixelResponse(); // Already recorded
   ```

3. **Add IP-based rate limiting** using Cloudflare's `request.cf` or a KV-based counter:
   ```typescript
   const rateKey = `rate:${ip}`;
   const count = parseInt(await env.RATE_KV.get(rateKey) || '0');
   if (count > 100) return pixelResponse(); // Silent rate limit
   await env.RATE_KV.put(rateKey, String(count + 1), { expirationTtl: 3600 });
   ```

4. **Improve bot detection** — Consider checking for:
   - Missing or suspicious headers (no `Accept`, no `Referer`)
   - Known datacenter IP ranges
   - TLS fingerprinting (via Cloudflare Bot Management if available)

salmonumbrella:
## Summary
- Add AES-256-GCM key rotation with versioned encryption blobs, backward-compatible decryption, and `gog gmail track key rotate` CLI command (closes #293)
- Add IP-based rate limiting via KV, D1-based open deduplication, and improved bot detection (header checks, timing heuristics, Cloudflare Bot Management) to the tracking pixel worker (closes #294)
- Fix Drive `--filter` query pass-through for `search` and `search-more` subcommands
- Fix Gmail MIME encoding for non-ASCII display names in From/Reply-To headers

## Test plan
- [x] Go build compiles cleanly
- [x] All Go tests pass (`go test ./internal/tracking/... ./internal/cmd/...`)
- [x] Key rotation: versioned encrypt → decrypt with multiple keys → legacy fallback
- [x] Bot detection: expanded UA matching, header presence checks, timing-based prefetch detection
- [x] `/q/` endpoint requires admin Bearer token (security fix from code review)
- [x] Drive filter pass-through with compound queries, case sensitivity
- [x] Gmail MIME non-ASCII header encoding edge cases

Closes #293
Closes #294

### DIFF
--- a/internal/cmd/drive.go
+++ b/internal/cmd/drive.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"google.golang.org/api/drive/v3"
@@ -20,6 +21,14 @@ import (
 )
 
 var newDriveService = googleapi.NewDrive
+
+var (
+	driveSearchFieldComparisonPattern = regexp.MustCompile(`(?i)\b(?:mimeType|name|fullText|trashed|starred|modifiedTime|createdTime|viewedByMeTime|visibility)\b\s*(?:!=|<=|>=|=|<|>)`)
+	driveSearchContainsPattern        = regexp.MustCompile(`(?i)\b(?:name|fullText)\b\s+contains\s+'`)
+	driveSearchMembershipPattern      = regexp.MustCompile(`(?i)'[^']+'\s+in\s+(?:parents|owners|writers|readers)`)
+	driveSearchHasPattern             = regexp.MustCompile(`(?i)\b(?:properties|appProperties)\b\s+has\s+\{`)
+	driveTrashedPattern               = regexp.MustCompile(`(?i)\btrashed\b`)
+)
 
 const (
 	driveMimeGoogleDoc     = "application/vnd.google-apps.document"
@@ -956,15 +965,38 @@ func buildDriveListQuery(folderID string, userQuery string) string {
 	} else {
 		q = parent
 	}
-	if !strings.Contains(q, "trashed") {
+	if !hasDriveTrashedPredicate(q) {
 		q += " and trashed = false"
 	}
 	return q
 }
 
 func buildDriveSearchQuery(text string) string {
-	q := fmt.Sprintf("fullText contains '%s'", escapeDriveQueryString(text))
-	return q + " and trashed = false"
+	q := strings.TrimSpace(text)
+	if q == "" {
+		return "trashed = false"
+	}
+	if !looksLikeDriveFilterQuery(q) {
+		return fmt.Sprintf("fullText contains '%s' and trashed = false", escapeDriveQueryString(q))
+	}
+	if !hasDriveTrashedPredicate(q) {
+		q += " and trashed = false"
+	}
+	return q
+}
+
+func looksLikeDriveFilterQuery(q string) bool {
+	if strings.EqualFold(q, "sharedWithMe") {
+		return true
+	}
+	return driveSearchFieldComparisonPattern.MatchString(q) ||
+		driveSearchContainsPattern.MatchString(q) ||
+		driveSearchMembershipPattern.MatchString(q) ||
+		driveSearchHasPattern.MatchString(q)
+}
+
+func hasDriveTrashedPredicate(q string) bool {
+	return driveTrashedPattern.MatchString(q)
 }
 
 func escapeDriveQueryString(s string) string {
--- a/internal/cmd/drive_search_more_test.go
+++ b/internal/cmd/drive_search_more_test.go
@@ -162,3 +162,62 @@ func TestDriveSearchCmd_NoResultsAndEmptyQuery(t *testing.T) {
 		t.Fatalf("expected empty query error")
 	}
 }
+
+func TestDriveSearchCmd_PassesThroughDriveFilterQueries(t *testing.T) {
+	origNew := newDriveService
+	t.Cleanup(func() { newDriveService = origNew })
+
+	const query = "mimeType = 'application/vnd.google-apps.document'"
+	const wantQ = query + " and trashed = false"
+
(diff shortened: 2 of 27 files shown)


steipete:
Thanks for the tracking security work. This is now superseded by main, so I’m closing the stale PR rather than trying to merge the dirty branch.\n\nLanded/current coverage:\n- #293 key rotation: fixed in e98f44d (versioned tracking blobs, `gmail track key rotate`, versioned Worker secrets, legacy fallback).\n- #294 abuse/rate limiting: already fixed/closed via dfc5b75 (dedupe repeated opens and cap per-IP writes).\n- retention/admin query hardening also exists on main via 9ce77ef and related tracking commits.\n- the unrelated Drive/Gmail fixes from the first PR commit are already on main via #266 / 4272c68, and issues #254/#255 are closed.\n\nVerified now:\n- `go test ./internal/tracking ./internal/cmd -run 'TestEncryptWithVersion|TestDecryptWithKeys|TestGmailTrackKeyRotate|TestGeneratePixelURL|TestGmailTrackSetupAndStatus'`\n- `go test ./internal/cmd -run 'TestBuildDriveSearchQuery|TestLooksLikeDriveFilterQuery|TestDriveSearchCmd_PassesThroughDriveFilterQueries|TestBuildRFC822UTF8FromDisplayName|TestFormatAddressHeaders|TestBuildRFC822PlainFromAddressStaysUnwrapped'`\n- `pnpm -C internal/tracking/worker test`\n\nClosing as superseded by landed main. Thanks again.

PR #314, opened 2026-02-18
feat(gmail): implement Safe Mode (drafts-only) via --safe flag

### LINKED ISSUE #239: Restrict Gmail to creating drafts, never send
Any ideas how to allow creating drafts but not sending messages? The Gmail API doesn't offer that level of control.

I always want a human in the loop, so creating drafts is the "most" I want the app to be able to do. 

Wondering if someone has thought about this.

Ishaan453:
# Description
This PR implements a "Safe Mode" for Gmail interactions, designed to ensure a "Human-in-the-Loop" workflow. When enabled, the CLI can create drafts but is cryptographically restricted from sending emails via the Google API.

# Key Changes
New Flag: Added --safe to gog auth add.
Restricted Scopes: Authentication with --safe requests only gmail.insert and gmail.readonly scopes (no gmail.send or gmail.compose).
Security Fix: Patched 

internal/googleauth/oauth_flow.go
to support disabling include_granted_scopes. This prevents new tokens from silently inheriting dangerous permissions from previous sessions.
Smart Fallback: Updated gog gmail drafts create to automatically fallback to the messages.insert API if the standard drafts.create permission is missing.
Verification

### Authenticate in Safe Mode:

`gog auth add <email> --services gmail --safe`

### Create a Draft (Should Succeed):

`gog gmail drafts create --to <email> --subject "Safe Test" --body "Hello"`

#### First, get the draft ID

`gog gmail drafts list`

#### Then try to send
`gog gmail drafts send <DRAFT_ID>`

Expected Result: Google API error (403 insufficientPermissions)

Fixes #239

### DIFF
--- a/internal/cmd/auth.go
+++ b/internal/cmd/auth.go
@@ -484,6 +484,7 @@ type AuthAddCmd struct {
 	ForceConsent bool          `name:"force-consent" help:"Force consent screen to obtain a refresh token"`
 	ServicesCSV  string        `name:"services" help:"Services to authorize: user|all or comma-separated ${auth_services} (Keep uses service account: gog auth service-account set)" default:"user"`
 	Readonly     bool          `name:"readonly" help:"Use read-only scopes where available (still includes OIDC identity scopes)"`
+	Safe         bool          `name:"safe" help:"Use safe/restricted scopes (eg. Gmail drafts only, no send)"`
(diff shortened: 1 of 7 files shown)


steipete:
This is a great idea, but incomplete and --safe would in my mind also limit things like deletion + needs to be thought of accoss the service. I close this for now so it doesn't get accidentally merged, but welcome you to rethink this in a bigger context.

PR #370, opened 2026-02-25
feat(docs): add read and edit commands for Google Docs

danhayman:
## Summary

  - **Add `gog docs read`** — reads a Google Doc as plain text with character-based `--offset` and `--limit`
  for paging through large documents
  - **Rewrite `gog docs edit`** — find-and-replace with `--old`/`--new` that requires the match to be unique by
   default, preventing accidental edits (use `--replace-all` to override). Adds `--dry-run`, escape sequences,
  and `--no-match-case`.

  ## Motivation

  My family uses Google Docs for shared lists and notes, and when I tried using an AI agent to help manage them
   via gog, it struggled — there was no structured way to read a doc with pagination or safely edit specific
  text. The existing `find-replace` replaces all occurrences with no uniqueness check, which is risky for
  automated use.

  These commands are modelled on [ox-cli](https://github.com/danhayman/ox-cli) (my CLI for editing local Office
   documents) so that agents can use a consistent interface across both local and Google Docs. `read` gives you
   plain text with character-based pagination for large documents, and `edit` requires a unique match so you
  don't accidentally change the wrong thing.

  This PR complements #359 (`docs sed`) rather than replacing it — `edit` is the simple "Find and Replace" for
  agents that just need `--old "X" --new "Y"`, while `sed` is the power tool for regex and multi-operation
  pipelines.

  ## New commands
```
## Read document content
gog docs read <docId>

## Page through a large doc (character-based offset/limit)
gog docs read <docId> --offset 5000 --limit 2000

## Read as JSON (includes totalChars for pagination logic)
gog docs read <docId> --json

## Safe edit — errors if "old text" isn't unique
gog docs edit <docId> --old "old text" --new "new text"

## Replace all occurrences
gog docs edit <docId> --old "foo" --new "bar" --replace-all

## Preview without modifying
gog docs edit <docId> --old "foo" --new "bar" --dry-run

## Case-insensitive matching
gog docs edit <docId> --old "Hello" --new "hi" --no-match-case --replace-all

## Escape sequences in --old/--new
gog docs edit <docId> --old "line1" --new "line1\nline2"

## Delete text (empty --new)
gog docs edit <docId> --old "remove this" --new ""
```
  ## Test plan

  - [x] Unit tests for helpers (`applyCharWindow`, `countOccurrences`, `unescapeString`)
  - [x] Integration tests for `read` (plain text, character offset/limit, offset-only, limit-only, offset past
  end, JSON output)
  - [x] Integration tests for `edit` (unique replacement, not-unique error, not-found error, `--replace-all`,
  JSON output, `--dry-run`, `--dry-run` JSON, deletion, escape sequences, `--no-match-case`)
  - [x] All existing docs tests pass (no regressions)
  - [x] Updated upstream `docs_edit_test.go` field references to match renamed struct fields

### DIFF
--- a/internal/cmd/docs.go
+++ b/internal/cmd/docs.go
@@ -22,7 +22,8 @@ type DocsCmd struct {
 	Info        DocsInfoCmd        `cmd:"" name:"info" aliases:"get,show" help:"Get Google Doc metadata"`
 	Create      DocsCreateCmd      `cmd:"" name:"create" aliases:"add,new" help:"Create a Google Doc"`
 	Copy        DocsCopyCmd        `cmd:"" name:"copy" aliases:"cp,duplicate" help:"Copy a Google Doc"`
-	Cat         DocsCatCmd         `cmd:"" name:"cat" aliases:"text,read" help:"Print a Google Doc as plain text"`
+	Cat         DocsCatCmd         `cmd:"" name:"cat" aliases:"text" help:"Print a Google Doc as plain text"`
+	Read        DocsReadCmd        `cmd:"" name:"read" help:"Read document content with character offset/limit for pagination. Use --offset and --limit to page through large documents."`
 	Comments    DocsCommentsCmd    `cmd:"" name:"comments" help:"Manage comments on files"`
 	ListTabs    DocsListTabsCmd    `cmd:"" name:"list-tabs" help:"List all tabs in a Google Doc"`
 	Write       DocsWriteCmd       `cmd:"" name:"write" help:"Write content to a Google Doc"`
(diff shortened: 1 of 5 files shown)


steipete:
Thanks for the contribution. I reviewed this against current `main`; the docs surface has moved on and now includes `docs cat` with `read`/`text` aliases plus `docs edit` and the richer `docs sed` path, all with broader tab/edit coverage.

Closing this as obsolete rather than keeping a conflicting duplicate open. If there is a specific remaining ox-cli-compatible behavior missing, a focused follow-up PR would be easier to review and land.

danhayman:
Cheers. Will check it out

PR #393, opened 2026-03-03
feat(calendar): add calendar aliases (supersedes #104)

salmonumbrella:
Supersedes #104.

This continues @zacharytamas's calendar alias work and incorporates CI/lint compatibility updates so it passes current checks.

## What’s included
- calendar alias config storage (`calendar_aliases`)
- `gog calendar alias set/list/unset` commands
- alias resolution integrated across calendar commands (create/update/delete/search/specialized commands)
- lint/format updates required by current repo rules

## Verification
- `make ci` passes locally
- `CGO_ENABLED=1 go build ./cmd/gog` passes locally

Thanks to @zacharytamas for the original implementation.


visionik:
Thank you @salmonumbrella and @zacharytamas (for the OG).

I need one day to think about the pattern of 

gog calendar alias
gog docs alias
etc

vs 

gog alias alias_name any_google_identifier

Any thoughts?

PR #439, opened 2026-03-11
fix(contacts): guard --from-file against all conflicting flags

klodr:
Flags --org, --title, --url, --note, --address and --custom were missing from the conflict check that rejects mixing --from-file with individual update flags. Omitting them meant the guard could be silently bypassed, letting users pass contradictory inputs without an error.

Added all six flags to the existing conditional, and extended TestContactsUpdate_FromFile_CantCombineWithFlags with a table-driven sub-test covering each previously unguarded flag.


steipete:
Landed on main.

- Local gate: `make ci`
- CI: run 24666775462 green
- Merge commit: 9e407bde7425e45c883da72df616427896de2758

Thanks @klodr!

PR #447, opened 2026-03-13
feat(gmail): add --full flag to messages search to disable body truncation

GodsBoy:
## Problem

`gog gmail messages search "query" --include-body` truncates message bodies to 200 characters in plain-text output. For longer emails this cuts off important content, forcing users to pipe through `--json` and manually extract the body field.

The `gmail thread get` command (`gog gmail read`) already has a `--full` flag that disables its 500-character truncation, but the messages search command has no equivalent.

## Solution

Add a `--full` flag to `GmailMessagesSearchCmd` that:

- Skips the 200-character truncation in `sanitizeMessageBody`
- Implies `--include-body` (so `--full` alone is sufficient)
- Preserves existing behaviour when `--full` is not passed

### Usage

```bash
# Before: body truncated at 200 chars
gog gmail messages search "invoice" --include-body

# After: full body in plain-text output
gog gmail messages search "invoice" --full
```

## Testing

- Added `TestSanitizeMessageBody_FullSkipsTruncation` test verifying full bodies are preserved
- Updated existing tests for the new function signature
- All tests pass (`make test`)
- Build passes (`make build`)
- Lint has one pre-existing `nilnil` issue in `gmail_filters_helpers.go` (present on main, not introduced by this PR)

## Changes

- `internal/cmd/gmail_messages.go`: Add `Full` field, thread it through to `sanitizeMessageBody`
- `internal/cmd/gmail_messages_test.go`: Add full-body test, update existing test signatures

### DIFF
--- a/internal/cmd/gmail_messages.go
+++ b/internal/cmd/gmail_messages.go
@@ -28,9 +28,13 @@ type GmailMessagesSearchCmd struct {
 	Timezone    string   `name:"timezone" short:"z" help:"Output timezone (IANA name, e.g. America/New_York, UTC). Default: local"`
 	Local       bool     `name:"local" help:"Use local timezone (default behavior, useful to override --timezone)"`
 	IncludeBody bool     `name:"include-body" help:"Include decoded message body (JSON is full; text output is truncated)"`
+	Full        bool     `name:"full" help:"Show full message bodies without truncation (implies --include-body)"`
 }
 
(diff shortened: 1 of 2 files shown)


steipete:
Landed manually on main with regression coverage for --full avoiding body truncation.

- Commit: fe06c3c feat(gmail): add full message search bodies (#447)
- Gate: make ci
- Live: sent a long self-test message with clawdbot@gmail.com, verified --include-body truncates and --full returns the full body, then trashed the test message.

Thanks @GodsBoy!

PR #454, opened 2026-03-16
feat(gmail): add --gmail-no-send flag to block send operations

### LINKED ISSUE #453: feat(gmail): add --gmail-no-send flag for agent safety
## Problem

`gmail.modify` is the right OAuth scope for an AI agent that needs to read, label, and create drafts — but it also permits send. Google provides no scope that grants modify-level access without send capability. This means there is currently no way to give an AI agent read/label/draft access while maintaining a hard block on send.

A prompt-level guardrail is not sufficient: if safety instructions are compacted away during a long session, an agent operating on `gmail.modify` could send email on behalf of the user. The Summer Yue incident (February 2026) demonstrated what happens when an AI agent loses its instructions mid-execution due to context window compaction and defaults to its last understood objective — her problem was deletion, but the equivalent risk here is send.

## Proposal

Add a `--gmail-no-send` global flag and `GOG_GMAIL_NO_SEND` env var that blocks **all** send operations in the gmail command group at the CLI layer, regardless of what OAuth scopes the token has.

### Blocked paths

| Command path | How it sends |
|---|---|
| `gog send` (top-level alias) | Direct compose + send |
| `gog gmail send` | Direct compose + send |
| `gog gmail drafts send <id>` | Sends existing draft |
| `gog gmail autoreply <query>` | Bulk auto-reply |

### Non-send commands (unaffected)

`watch serve`, `autoforward`, `vacation`, `filters`, `forwarding`, `sendas`, `delegates`, `drafts create`, `drafts update`, and all read/organize commands continue to work normally.

### Design

Follow the same centralized enforcement pattern as `enforceEnabledCommands`:
- Post-parse, pre-`Run()` check in `Execute()`
- Inspects `kctx.Command()` to identify send commands
- Returns `usagef(...)` error (exit code 2) if blocked
- No auth/API calls needed — pure CLI-layer gate

### Usage

```bash
# Flag
gog --gmail-no-send gmail send --to x@y.com --subject S --body B
# → "send blocked by --gmail-no-send (GOG_GMAIL_NO_SEND): use "gog gmail drafts create" to compose without sending"

# Env var (preferred for agent deployments)
GOG_GMAIL_NO_SEND=1 gog gmail send --to x@y.com --subject S --body B
# → same

# Non-send commands pass through
GOG_GMAIL_NO_SEND=1 gog gmail search "from:boss"
# → works normally
```

This gives agent operators a hard invariant they can rely on: deploy gog with `GOG_GMAIL_NO_SEND=1` and send is impossible, full stop, regardless of what the agent is told or forgets.

veteranbv:
## Motivation

`gmail.modify` is the right OAuth scope for an agent that needs to read, label,
and create drafts, but it also permits send. Google provides no scope that
grants modify-level access without send capability. This means there is no way
to give an AI agent read/label/draft access while maintaining a hard block on
send.

A prompt-level guardrail is not sufficient: if safety instructions are compacted
away during a long session, an agent operating on `gmail.modify` could send
email on behalf of the user. The block needs to live in the binary itself, below
the instruction layer, where context compaction cannot reach it.

Closes #453

## Summary

- Add `--gmail-no-send` global flag and `GOG_GMAIL_NO_SEND` env var
- Add `gmail_no_send` config key (`gog config set gmail_no_send true`) for
  persistent global blocking without env vars
- Add per-account send blocking via `gog config no-send set/remove/list`,
  stored as `no_send_accounts` in config.json
- Block all four send paths: `send`, `gmail send`, `gmail drafts send`,
  `gmail autoreply`
- Global enforcement in `Execute()` using the same pattern as
  `enforceEnabledCommands` (post-parse, pre-`Run()`, exit code 2)
- Per-account enforcement in each send command's `Run()` after account
  resolution via `requireAccount()`
- Error messages identify the blocking source (flag, config key, or account)

Precedence: CLI flag > env var > config key > per-account config > default (allow).

## Test plan

- [x] `TestIsGmailSendCommand` -- unit coverage for all send and non-send
      command strings
- [x] `TestGmailNoSendBlocksViaCLI` -- integration test: all four send paths
      blocked via `--gmail-no-send`
- [x] `TestGmailNoSendAllowsNonSendCommands` -- non-send commands pass through
- [x] `TestGmailNoSendEnvVar` -- `GOG_GMAIL_NO_SEND=1` blocks send
- [x] `TestGmailNoSendNotSetAllowsSend` -- without flag, send is not blocked
- [x] `TestGmailNoSendConfigGlobal` -- `gmail_no_send: true` in config blocks
- [x] `TestGmailNoSendConfigGlobalAllowsNonSend` -- config global does not
      block non-send commands
- [x] `TestGmailNoSendGlobalOverridesPerAccount` -- flag wins over per-account
- [x] `TestCheckAccountNoSend_Blocked` -- per-account blocks matching account
- [x] `TestCheckAccountNoSend_Allowed` -- per-account allows non-matching
- [x] `TestCheckAccountNoSend_NoConfig` -- absent config does not block
- [x] Config unit tests: `IsNoSendAccount`, `SetNoSendAccount`,
      `ListNoSendAccounts` (with nil map, case insensitivity, empty email)
- [x] `go test ./...` -- full suite passes, no regressions
- [x] `make lint` -- clean

### DIFF
--- a/internal/cmd/config_cmd.go
+++ b/internal/cmd/config_cmd.go
@@ -10,12 +10,13 @@ import (
 )
 
 type ConfigCmd struct {
-	Get   ConfigGetCmd   `cmd:"" aliases:"show" help:"Get a config value"`
-	Keys  ConfigKeysCmd  `cmd:"" aliases:"list-keys,names" help:"List available config keys"`
-	Set   ConfigSetCmd   `cmd:"" aliases:"add,update" help:"Set a config value"`
-	Unset ConfigUnsetCmd `cmd:"" aliases:"rm,del,remove" help:"Unset a config value"`
-	List  ConfigListCmd  `cmd:"" aliases:"ls,all" help:"List all config values"`
-	Path  ConfigPathCmd  `cmd:"" aliases:"where" help:"Print config file path"`
+	Get    ConfigGetCmd    `cmd:"" aliases:"show" help:"Get a config value"`
+	Keys   ConfigKeysCmd   `cmd:"" aliases:"list-keys,names" help:"List available config keys"`
+	Set    ConfigSetCmd    `cmd:"" aliases:"add,update" help:"Set a config value"`
+	Unset  ConfigUnsetCmd  `cmd:"" aliases:"rm,del,remove" help:"Unset a config value"`
+	List   ConfigListCmd   `cmd:"" aliases:"ls,all" help:"List all config values"`
+	Path   ConfigPathCmd   `cmd:"" aliases:"where" help:"Print config file path"`
+	NoSend ConfigNoSendCmd `cmd:"" name:"no-send" aliases:"nosend" help:"Manage per-account send blocking"`
 }
 
 type ConfigGetCmd struct {
--- a/internal/cmd/config_no_send.go
+++ b/internal/cmd/config_no_send.go
@@ -1,0 +1,96 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/steipete/gogcli/internal/config"
+	"github.com/steipete/gogcli/internal/outfmt"
+)
+
+type ConfigNoSendCmd struct {
+	Set    ConfigNoSendSetCmd    `cmd:"" aliases:"add,block" help:"Block send for an account"`
+	Remove ConfigNoSendRemoveCmd `cmd:"" aliases:"rm,unblock,allow" help:"Allow send for an account"`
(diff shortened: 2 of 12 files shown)


steipete:
Thanks @veteranbv! Landed Gmail no-send safety on main in d84148a, including --gmail-no-send, GOG_GMAIL_NO_SEND, config gmail_no_send, and per-account config no-send guards. Closing this PR as landed.

PR #463, opened 2026-03-19
feat(secrets): allow overriding keyring service name

mkurz:
## Summary

Add `GOG_KEYRING_SERVICE_NAME` as an environment variable override for the keyring `ServiceName`.

This keeps the default behavior unchanged (`gogcli`), but allows Linux/KWallet users to point `gog` at a different wallet/collection name when needed, for example:

```bash
export GOG_KEYRING_SERVICE_NAME=kdewallet
```

## Changes

- add `GOG_KEYRING_SERVICE_NAME` support in internal/secrets/store.go
- keep the default service name as `gogcli` when the env var is unset
- add tests covering the default and override behavior
- document the new env var in `README.md` and `docs/spec.md`

## Test plan

- [x] GOCACHE=/tmp/go-build go test ./internal/secrets

## User-facing changes

- new env var: `GOG_KEYRING_SERVICE_NAME`
- no behavior change unless the env var is set

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,6 +286,14 @@ export GOG_KEYRING_BACKEND=file
 
 Precedence: `GOG_KEYRING_BACKEND` env var overrides `config.json`.
 
+Override the OS keyring service/collection/wallet name via env:
+
+```bash
+export GOG_KEYRING_SERVICE_NAME=kdewallet
+```
+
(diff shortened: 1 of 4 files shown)


steipete:
Thanks! Landed manually on main as c4665c4, with a follow-up lint cleanup in a8531f2, plus tests/docs/CHANGELOG and co-author trailer. Closing this PR since the keyring service-name override is now on main.

PR #473, opened 2026-03-24
feat: add `auth credentials remove` command

yamagucci:
## Summary

- Add `gog auth credentials remove` subcommand to remove stored OAuth client credentials
- Previously only `set` and `list` were available — there was no way to remove credentials via the CLI
- **Also removes associated refresh tokens** for all accounts authenticated under the deleted client
- Cleans up domain mappings (`--domain`) that reference the removed client

## Usage

```bash
# Remove default client credentials (+ its tokens)
gog auth credentials remove

# Remove a named client
gog auth credentials remove work

# Remove all stored credentials
gog auth credentials remove all

# Preview what would be removed
gog auth credentials remove --dry-run

# Skip confirmation prompt
gog auth credentials remove --force
```

## Confirmation prompt example

```
Proceed to remove OAuth credentials for client "work" and 2 associated token(s) (alice@co.com, bob@co.com)? [y/N]:
```

## Implementation

- **`config.DeleteClientCredentialsFor(client)`** — removes the credentials JSON file
- **`AuthCredentialsRemoveCmd`** — CLI command following the same patterns as `auth remove`
  - Accepts optional positional `[<client>]` argument
  - Supports `all` to remove every stored credential
  - Falls back to `--client` flag or default when no arg is given
  - Finds and deletes all tokens stored under the target client
  - Cleans up domain mappings referencing the deleted client
  - Confirmation prompt shows affected accounts, `--dry-run`, `--force`, `--json` output

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/config/... ./internal/cmd/...` passes
- [x] `gog auth credentials remove --help` shows correct usage
- [ ] Manual test: set + add accounts + remove + list cycle

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 ### Added
+- Auth: add `auth credentials remove` to delete stored OAuth client credentials and associated refresh tokens. (#473) — thanks @yamagucci.
 - Gmail: add `gmail autoreply` to reply once to matching messages, label the thread for dedupe, and optionally archive/mark read. Includes docs and regression coverage for skip/reply flows.
 - Gmail: add `gmail messages search --full` to print complete message bodies instead of truncating text output. (#447) — thanks @GodsBoy.
 - Drive: allow `drive share --role commenter` for comment-only sharing. (#443) — thanks @pavelzak.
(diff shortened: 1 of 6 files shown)


steipete:
Landed in c73bb23 after a polish pass.

Changes added while landing:
- propagated credentials/config/keyring errors instead of swallowing them
- added dry-run JSON behavior for remove/remove-all
- removed associated tokens and domain mappings deterministically
- added regression coverage and README/spec/CHANGELOG entries

Verified locally with `make fmt`, `make lint`, `make test`, `make build`, and `make ci`.

Thanks again @yamagucci.

PR #476, opened 2026-03-25
fix(gmail): stop hard-wrapping plain text email bodies at ~72 chars

### LINKED ISSUE #475: bug(gmail): --body and --body-file silently hard-wrap text at ~72 chars, breaking email formatting
## Description

When sending or creating drafts with `--body` or `--body-file`, the CLI hard-wraps plain text at ~72 characters (RFC 2822 line folding) before sending to the Gmail API. This silently destroys paragraph formatting in sent emails.

## Steps to Reproduce

1. Create a file with a single long paragraph (e.g., 200+ chars on one line):
   ```
   echo "This is a long paragraph that should reflow naturally in the recipient's email client but instead gets chopped into narrow lines by the CLI before it even reaches Gmail." > /tmp/test.txt
   ```
2. Send or create a draft:
   ```
   gog gmail drafts create --to "test@example.com" --subject "Test" --body-file /tmp/test.txt -a me@example.com --client myclient
   ```
3. Send the draft from Gmail

## Expected Behavior

The paragraph renders as a single reflowing block in the recipient's email client.

## Actual Behavior

The paragraph is chopped into ~72-char lines with hard `\r\n` breaks:

```
This is a long paragraph that should reflow naturally in the
recipient's email client but instead gets chopped into narrow lines
by the CLI before it even reaches Gmail.
```

This is **invisible in Gmail's compose/draft view** (which reflows text while editing) but appears as broken formatting once the email is actually sent and viewed by recipients.

## Root Cause

The Go MIME library used to construct the `text/plain` MIME part applies RFC 2822 line folding at ~72 characters. While this was historically needed for SMTP transport, modern email clients and the Gmail API handle long lines fine. The wrapping is applied after reading the input file, so even perfectly formatted single-line paragraphs get wrapped.

## Suggested Fix

**Option A (preferred):** Stop hard-wrapping `text/plain` bodies by default. The Gmail API and modern email infrastructure do not require 72-char folding for message bodies.

**Option B:** Add a `--no-wrap` flag to disable line folding when the user wants to preserve their original formatting.

**Option C (minimum):** Document this behavior so users know `--body` content will be reformatted, and recommend `--body-html` as a workaround.

## Workaround

Use `--body-html` with `<p>` tags instead of `--body` / `--body-file`. HTML bodies are not subject to line folding.

## Environment

- gogcli v0.12.0 (Homebrew, macOS)
- Sending via Gmail API

shashankkr9:
## Summary

- Fixes #475 — plain text bodies sent via `gog gmail send --body` / `--body-file` were hard-wrapped at ~72 characters, destroying paragraph formatting
- **Root cause**: `text/plain` MIME parts used `Content-Transfer-Encoding: 7bit`, so long lines violated RFC 5322 limits and Gmail's SMTP transport hard-wrapped them
- **Fix**: Switch `text/plain` parts to `quoted-printable` encoding — the QP encoder inserts soft line breaks (`=\r\n`) that mail clients strip, preserving original paragraph structure
- HTML bodies remain `7bit` (unaffected since HTML ignores whitespace)

## Changes

- `internal/cmd/gmail_mime.go` — use `mime/quotedprintable.Writer` for all plain text body paths (standalone, multipart/alternative, multipart/mixed)
- `internal/cmd/gmail_mime_test.go` — 3 new tests: long line QP round-trip, multi-paragraph round-trip, HTML stays 7bit

## Test plan

- [x] All existing tests pass (`go test ./...`)
- [x] `go vet ./...` clean
- [x] Live test: 325-char single paragraph arrives as 1 line (no wrapping)
- [x] Live test: 3 paragraphs with blank lines — structure preserved
- [x] Live test: HTML body still works (regression check)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/internal/cmd/gmail_mime.go
+++ b/internal/cmd/gmail_mime.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"mime"
+	"mime/quotedprintable"
 	"net/mail"
 	"net/url"
 	"os"
@@ -142,9 +143,9 @@ func buildRFC822(opts mailOptions, cfg *rfc822Config) ([]byte, error) {
 			return b.Bytes(), nil
 		default:
 			writeHeader(&b, "Content-Type", "text/plain; charset=\"utf-8\"")
-			writeHeader(&b, "Content-Transfer-Encoding", "7bit")
+			writeHeader(&b, "Content-Transfer-Encoding", "quoted-printable")
 			b.WriteString("\r\n")
-			writeBodyWithTrailingCRLF(&b, plainBody)
+			writeQuotedPrintableBody(&b, plainBody)
 			return b.Bytes(), nil
 		}
 	}
@@ -175,8 +176,8 @@ func buildRFC822(opts mailOptions, cfg *rfc822Config) ([]byte, error) {
 		writeBodyWithTrailingCRLF(&b, htmlBody)
 	default:
 		b.WriteString("Content-Type: text/plain; charset=\"utf-8\"\r\n")
-		b.WriteString("Content-Transfer-Encoding: 7bit\r\n\r\n")
-		writeBodyWithTrailingCRLF(&b, plainBody)
+		b.WriteString("Content-Transfer-Encoding: quoted-printable\r\n\r\n")
+		writeQuotedPrintableBody(&b, plainBody)
 	}
 
 	// Attachments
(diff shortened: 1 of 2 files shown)


steipete:
Landed on main.

- Local gate: `make ci`
- CI: run 24666775462 green
- Merge commit: e787894cdf371f3e247738d87c4d56105acd44c9

Thanks @shashankkr9!

PR #477, opened 2026-03-25
fix: use 8bit Content-Transfer-Encoding for non-ASCII text bodies

yeager:
## Problem
When sending emails with non-ASCII body text (e.g. Swedish, German) AND attachments, the MIME `Content-Transfer-Encoding` header was set to `7bit` for text parts. Per RFC 2045 §2.7, `7bit` encoding must not contain octets with the high bit set.

This caused some mail servers/clients to either mangle the message body or strip attachments from `multipart/mixed` messages.

## Fix
Introduce `transferEncodingForText()` helper that:
- Returns `"7bit"` for pure ASCII text (no change for English-only emails)
- Returns `"8bit"` for text containing non-ASCII bytes (UTF-8)

Applied to all text body emission paths:
- Standalone text/plain and text/html
- Parts within `multipart/alternative`
- Inline text parts within `multipart/mixed` (emails with attachments)

## Testing
All existing tests pass. The change is backwards-compatible — ASCII-only emails continue to use `7bit`.

### DIFF
--- a/internal/cmd/gmail_mime.go
+++ b/internal/cmd/gmail_mime.go
@@ -136,13 +136,13 @@ func buildRFC822(opts mailOptions, cfg *rfc822Config) ([]byte, error) {
 			return b.Bytes(), nil
 		case hasHTML && !hasPlain:
 			writeHeader(&b, "Content-Type", "text/html; charset=\"utf-8\"")
-			writeHeader(&b, "Content-Transfer-Encoding", "7bit")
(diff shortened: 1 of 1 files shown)


steipete:
Thanks! Landed manually on main as 3d63ebc with regression coverage and a CHANGELOG thank-you. I kept plain text quoted-printable, and fixed the non-ASCII HTML path so it no longer declares 7bit. Closing this PR since the fix is now on main.

PR #481, opened 2026-03-27
fix(cli): generate native zsh completions

piiq:
## Summary
- replace the zsh completion output with a native `_gog` zsh completer
- keep the implementation small by calling `gog __complete` directly from zsh
- update the existing completion test to expect the native zsh marker

## Why
The current `gog completion zsh` output is a bash compatibility shim using `bashcompinit`, which is awkward in zsh and does not install as a normal native zsh completion script.

## Validation
- `go test ./internal/cmd`
- `go test ./...`
- `go run ./cmd/gog completion zsh | sed -n '1,40p'`

### DIFF
--- a/internal/cmd/completion_scripts.go
+++ b/internal/cmd/completion_scripts.go
@@ -37,9 +37,24 @@ complete -F _gog_complete gog
 func zshCompletionScript() string {
 	return `#compdef gog
 
-autoload -Uz bashcompinit
-bashcompinit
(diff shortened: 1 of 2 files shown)


steipete:
Thanks @piiq! Landed the native zsh completion fix on main in 4f5e474. I kept it focused on the __complete bridge and added coverage. Closing this PR as landed.

piiq:
yay!

PR #482, opened 2026-03-27
feat(gmail): add forward command

spencer-c-reed:
## Summary

- Adds `gog gmail forward <messageId> --to <recipients>` (alias `fwd`) to forward a Gmail message to new recipients
- Fetches the original message in full format, constructs a proper RFC 2822 MIME forward with original headers (From, Date, Subject, To, Cc), plain text + HTML body, and re-downloaded attachments
- Sets `threadId`, `In-Reply-To`, and `References` headers so the forward stays in the sender's Gmail thread
- Supports `--note` / `--note-file` for introductory text above the forwarded content
- `--skip-attachments` flag to omit original attachments
- `--from` for send-as alias support
- Full `--dry-run`, `--json`, `--plain` output mode support
- Deduplicates forward subject prefixes (`Fwd: Fwd:` → `Fwd:`)

## Motivation

The Gmail API has no native forward endpoint — forwarding is done by constructing a new message via `messages.send`. gog currently supports replies (`--reply-to-message-id`, `--quote`) but not forwards, which is a common workflow gap. This command fills that gap using the same `buildRFC822()` and `replyInfo` infrastructure that replies already use.

## Test plan

- [x] `TestExecute_GmailForward_Basic` — verifies subject, forward separator, original headers in body, note text, original body content, threadId, In-Reply-To header
- [x] `TestExecute_GmailForward_WithAttachments` — verifies attachments are fetched and included in MIME
- [x] `TestExecute_GmailForward_SkipAttachments` — verifies `--skip-attachments` prevents attachment fetch
- [x] `TestBuildForwardSubject` — subject prefix handling (new, existing Fwd:, double Fwd:, Fw:, FWD:, empty, Re:)
- [x] `TestStripForwardPrefix` — prefix stripping edge cases
- [x] `TestFormatForwardedMessage` — plain text body construction with all fields
- [x] `TestFormatForwardedMessage_NoNote` — no leading blank lines when note is empty
- [x] `TestFormatForwardedMessageHTML` — HTML body construction
- [x] All 10 tests passing
- [x] Tested live against real Gmail account — verified proper threading and content


steipete:
Landed in `13a1aaea13482fdd4f826abedd250a6ef7a97dbb`. Thanks @spencer-c-reed.

What changed in the landing pass:
- kept `gmail forward` / `gmail fwd` and attachment forwarding
- added forward/fwd to Gmail no-send guard paths
- added per-account no-send enforcement before any Gmail fetch/send work
- de-duplicated forwarded header formatting and test service setup
- updated README, `docs/spec.md`, and `CHANGELOG.md`

Verified:
- `make fmt`
- `go test ./internal/cmd -run Forward\|NoSend`
- `make lint && make test && make build`
- `make ci`
- clawdbot live smoke: read-only search/thread get, forward dry-run with a real message id, and `--gmail-no-send` blocked forward before send.

PR #483, opened 2026-03-27
fix(docs): use UTF-16 sed offsets and preserve & replacements

bill492:
## Summary
- fix Google Docs API range calculations in sed formatting paths to use UTF-16 code unit lengths instead of Go byte lengths
- preserve whole-match `&` replacements in addressed substitutions after regex expansion
- add regression coverage for UTF-16 style ranges and `s/foo/&/`

## Testing
- `go build ./...`
- `go test ./internal/cmd/... -run Sed -v -count=1`

## Notes
- This PR intentionally does **not** refactor paragraph-addressed brace formatting (`runAddressedSubstitute`) to emit formatting requests. That bug is real, but it is a larger behavior change and felt too risky to mix into the UTF-16/data-corruption fix.

### DIFF
--- a/internal/cmd/docs_sed.go
+++ b/internal/cmd/docs_sed.go
@@ -102,14 +102,11 @@ type indexedExpr struct {
 	expr  sedExpr
 }
 
-// literalReplacement returns the replacement string with Go regex escaping undone,
+// literalReplacement returns the replacement string with Go regex dollar escaping undone,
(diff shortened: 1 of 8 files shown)


steipete:
Landed on main.

- Local gate: `make ci`
- CI: run 24666775462 green
- Merge commit: 4185e02a54127c0205ea6072d393f0a35ea28f9b

Thanks @bill492!

Issue #423, opened 2026-03-06
Calendar event timezone bug - events created at wrong time


Lyndon2066:
## Bug Description

When creating calendar events with timezone-aware timestamps, the time is incorrectly interpreted.

## Steps to Reproduce

1. Run: `gog calendar create primary --summary "会议" --from 2026-03-06T19:00:00+08:00 --to 2026-03-06T20:00:00+08:00 --force`
2. Check the created event time

## Expected Behavior

- Input: 2026-03-06T19:00:00+08:00 (19:00 Beijing Time)
- Should create event at 19:00 Beijing / 11:00 UTC

## Actual Behavior

- Event created at 06:00 local time (shows as 06:00 America/New_York or UTC-5)
- The timezone in +08:00 is ignored, defaults to UTC-5

## Environment

- gog version: v0.11.0
- OS: macOS
- Calendar: Google Calendar (primary)

## Additional Notes

- Using RFC3339 format with timezone (e.g., `+08:00`) does not work
- The tool seems to default to America/New_York or UTC-5 timezone regardless of input

steipete:
Fixed on current main.

Current calendar create/update payload construction preserves RFC3339 timestamps with their numeric offset and also supplies Calendar `timeZone` metadata derived from that offset (for example `+08:00` maps to fixed-offset `Etc/GMT-8`). That prevents the older behavior where offset-aware input could be interpreted in the local/default calendar timezone.

Regression proof on current main:

```bash
go test ./internal/cmd -run TestCalendarCreateCmd_RecurringOffsetTimezoneFallback -count=1
```

That test exercises offset-aware calendar creation and verifies the generated event payload carries explicit timezone metadata instead of relying on a local/default timezone. For named IANA zones, current main also has `--start-timezone` / `--end-timezone` from #422.

Issue #422, opened 2026-03-06
calendar: preserve named IANA time zones (Europe/Rome) instead of Etc/GMT offsets


nberardi:
## Summary
`gog calendar update` writes events with timezone labels normalized to `Etc/GMT-2` instead of preserving named IANA zones like `Europe/Rome`.

The actual instant is correct, but timezone semantics are degraded (important for travel events and editor UX consistency).

## Environment
- gog version: `0.11.0` (Homebrew)
- OS: macOS Darwin 25.3.0 (arm64)
- Account/calendar: Google Calendar shared calendar

## Repro
1. Create/update a calendar event with explicit cross-timezone start/end via `gog`:

```bash
gog calendar update <calendarId> <eventId> \
  --from 2026-08-13T13:40:00+02:00 \
  --to   2026-08-13T17:00:00-04:00 \
  --no-input --json
```

2. Fetch event:

```bash
gog calendar event <calendarId> <eventId> --json
```

## Actual
Returned event timezone fields show:
- start.timeZone: `Etc/GMT-2`
- end.timeZone: `America/New_York`

## Expected
When a named IANA zone is intended (e.g. `Europe/Rome`), preserve that zone name in event timezone metadata rather than canonicalizing to `Etc/GMT-2`.

## Why this matters
- Calendar UI consistency: users expect `Europe/Rome` in timezone editor.
- DST and human readability are clearer with named zones.

## Notes
- Creating/editing same event via Google Calendar web UI can preserve named zones as expected.
- This appears to be a gog write/update behavior, not a Google Calendar rendering issue.

steipete:
Fixed on main in 28d9e98.

`calendar create` and `calendar update` now accept explicit named timezone metadata via `--start-timezone` / `--end-timezone` (aliases: `--from-timezone` / `--to-timezone`). That keeps RFC3339 dateTime values intact while sending the intended IANA `timeZone`, e.g.:

```bash
gog calendar update <calendarId> <eventId> \
  --from 2026-08-13T13:40:00+02:00 --start-timezone Europe/Rome \
  --to   2026-08-13T17:00:00-04:00 --end-timezone America/New_York
```

Also added regression coverage for create/update payloads and validation for invalid/all-day timezone usage. Thanks for the detailed repro notes.

PR #492, opened 2026-04-07
calendar: avoid ambiguous timezone inference from offsets

RaphaelRUzan:
## Summary
Avoid ambiguous regional timezone inference from offset-only RFC3339 inputs.

This changes calendar event timezone extraction to use fixed-offset `Etc/GMT±N` zones instead of guessing named regions like `America/Phoenix` or `America/Los_Angeles` from a numeric offset alone.

## Why
Offsets like `-07:00` are ambiguous at a given instant and can match multiple regional IANA zones. The previous logic returned the first matching US timezone, which could mislabel Pacific events as `America/Phoenix`.

Using a fixed-offset timezone is less pretty, but safer and non-deceptive when the input does not carry a true named timezone.

## Changes
- remove ambiguous US region guessing from `extractTimezone()`
- keep `UTC` for zero offset
- keep `Etc/GMT±N` fallback for whole-hour offsets
- update tests accordingly

## Tests
Ran:

```bash
go test ./internal/cmd -run 'TestExtractTimezone|TestCalendarCreateCmd_RecurringIncludesTimezone|TestCalendarUpdateCmd_RecurrenceFillsMissingTimezone'
```

## Related
- #422
- #423


steipete:

- Calendar commit: df3e97f
- Changelog thanks: 89916a5
- Gate: `make ci`

Thanks!

PR #506, opened 2026-04-19
chat spaces find: substring match by default, --exact for legacy behavior

### LINKED ISSUE #503: chat spaces find: support partial/substring matching on displayName
## Description

`gog chat spaces find` currently requires an **exact match** on the full `displayName` to return results. For example, if a space is named "My Project Team", searching for "Project" returns no results — only the full string "My Project Team" works.

## Expected behavior

It would be very useful if `gog chat spaces find` supported **partial/substring matching** (case-insensitive), similar to a SQL `LIKE '%keyword%'` or a simple `contains` check.

For example:
```bash
# Currently: no results
gog chat spaces find "Project"

# Expected: returns all spaces whose displayName contains "Project"
# e.g. "My Project Team", "Project Alpha", "Old Project Archive"
```

## Current workaround

The only way to do a partial search right now is:
```bash
gog chat spaces list --all -p | grep -i "project"
```

This works but requires fetching the entire list every time, which is slower and less ergonomic.

## Suggestion

Either:
- Make the existing `find` command do substring matching by default (since exact match is a rare use case)
- Or add a flag like `--exact` for exact matching, with substring as the default

mvanhorn:
Closes #503

\`gog chat spaces find\` previously required an exact, case-insensitive match on the full \`displayName\`. Searching \`Project\` against a space named \`My Project Team\` returned no results, and the documented workaround was \`gog chat spaces list --all -p | grep\`, which is slower and less ergonomic.

This change makes \`find\` do substring matching (case-insensitive) by default and adds an \`--exact\` flag to opt into the previous behavior. The underlying Google Chat \`Spaces.List\` API has no server-side display name filter, so both modes iterate the same paginated result set; substring matching is strictly more permissive. Scripts that relied on the exact-match behavior can add \`--exact\` to keep the old semantics.

Examples:

\`\`\`console
# default (substring, case-insensitive)
gog chat spaces find Project
# returns: My Project Team, Project Alpha, Old Project Archive

# legacy exact match
gog chat spaces find --exact \"Project Alpha\"
# returns: Project Alpha
\`\`\`

### Tests

- \`TestExecute_ChatSpacesFind_Substring\` - default search \`project\` matches \`My Project Team\`, \`Project Alpha\`, and \`Old Project Archive\`, and excludes \`Random Channel\`.
- \`TestExecute_ChatSpacesFind_Exact\` - \`--exact \"project alpha\"\` returns only the space whose \`displayName\` equals \`Project Alpha\` (case-insensitive).
- Existing \`TestExecute_ChatSpacesFind_JSON\` still passes (exact-named \`Engineering\` remains matched by the default substring behavior because it is a substring of itself).

\`go test ./internal/cmd/ -run TestExecute_ChatSpacesFind\` and \`go build ./...\` both clean locally.

### DIFF
--- a/internal/cmd/chat_spaces.go
+++ b/internal/cmd/chat_spaces.go
@@ -124,8 +124,9 @@ func (c *ChatSpacesListCmd) Run(ctx context.Context, flags *RootFlags) error {
 }
 
 type ChatSpacesFindCmd struct {
-	DisplayName string `arg:"" name:"displayName" help:"Space display name"`
+	DisplayName string `arg:"" name:"displayName" help:"Space display name (substring match, case-insensitive)"`
 	Max         int64  `name:"max" aliases:"limit" help:"Max results per page" default:"100"`
+	Exact       bool   `name:"exact" help:"Require an exact, case-insensitive match on displayName instead of substring match"`
 }
 
 func (c *ChatSpacesFindCmd) Run(ctx context.Context, flags *RootFlags) error {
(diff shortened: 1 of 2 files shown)


steipete:
Landed manually on main with regression coverage for substring and --exact matching.

- Commit: f1cef0d feat(chat): support substring space lookup (#506)
- Gate: make ci
- Live: attempted against clawdbot@gmail.com; command correctly requires a Workspace account for Chat, so live API coverage was blocked by account type.

Thanks @mvanhorn!

mvanhorn:
Thanks for landing this @steipete and for adding the regression coverage. Appreciate the ping on the Workspace account blocker too.

PR #507, opened 2026-04-20
fix(tasks): clear due date when --due= is set to empty

### LINKED ISSUE #437: tasks update --due= does not clear the due date
## Description

`gog tasks update <listId> <taskId> --due= --force` does not clear the due date on a task, despite the help text stating "set empty to clear".

## Steps to Reproduce

1. Have a task with a due date set
2. Run: `gog tasks update <listId> <taskId> --due= --account <email> --force`
3. Observe the due date is unchanged

## Expected Behavior

The due date should be cleared (set to null via the Google Tasks API PATCH with `{"due": null}`).

## Actual Behavior

The due date remains unchanged. The command returns successfully and shows the old due date.

When using `--json` output on the update response, the `due` field shows `null` — but subsequent `gog tasks list` confirms the original due date is still set. This suggests the empty value may be omitted from the PATCH request body rather than being sent as an explicit `null`.

## Environment

- gog version: v0.12.0 (c18c58c 2026-03-09)
- OS: Linux 6.8.0 x86_64
- Installed via: brew (steipete/tap/gogcli)

## Workaround

Currently setting `--due 2099-12-31` as a workaround.

dinakars777:
## Summary

When `gog tasks update --due=` was used with an empty value, the due date was not actually cleared despite the help text stating "set empty to clear". The Google Tasks API PATCH treats an empty string the same as an omitted field due to `omitempty` JSON tagging, so the due date remained unchanged.

This fix adds `"Due"` to `NullFields` when the due value is empty, which forces the API to send an explicit `null` in the PATCH request body, properly clearing the due date.

## Steps to Reproduce (before fix)

1. Have a task with a due date set
2. Run: `gog tasks update <listId> <taskId> --due= --force`
3. Observe the due date is unchanged

## Expected Behavior

The due date should be cleared (set to null).

## Changes

- When `--due=` is provided with an empty value, add `"Due"` to `patch.NullFields` instead of setting `patch.Due = ""`
- Move the `warnTasksDueTime` call inside the non-empty branch since it's only relevant when a due date is being set

## Test plan

- [x] `go test ./internal/cmd/... -run TestTasksValidation` passes
- [ ] Manual test: `gog tasks update <listId> <taskId> --due= --force` clears the due date
- [ ] Manual test: `gog tasks update <listId> <taskId> --due=2026-05-01` still sets the due date normally

Closes #437

### DIFF
--- a/internal/cmd/tasks_items.go
+++ b/internal/cmd/tasks_items.go
@@ -491,14 +491,18 @@ func (c *TasksUpdateCmd) Run(ctx context.Context, kctx *kong.Context, flags *Roo
 		changed = true
 	}
 	if flagProvided(kctx, "due") {
-		if !outfmt.IsJSON(ctx) {
-			warnTasksDueTime(u, c.Due)
-		}
 		dueValue, dueErr := normalizeTaskDue(c.Due)
 		if dueErr != nil {
 			return dueErr
 		}
-		patch.Due = dueValue
+		if dueValue == "" {
+			patch.NullFields = append(patch.NullFields, "Due")
+		} else {
+			if !outfmt.IsJSON(ctx) {
+				warnTasksDueTime(u, c.Due)
+			}
+			patch.Due = dueValue
+		}
 		changed = true
 	}
 	if flagProvided(kctx, "status") {


steipete:
Landed manually on `main` as part of a grouped Tasks fix commit.

- Tasks commit: 570294c
- Changelog thanks: 89916a5
- Gate: `make ci`

Thanks!

PR #508, opened 2026-04-20
fix(calendar): add Context and ShowDeleted to search command

dinakars777:
## Summary

The `calendar search` command was missing `.Context(ctx)` and `.ShowDeleted(false)` on its `Events.List` call, unlike the regular events listing path in `calendar_list.go`.

- **Missing `Context(ctx)`**: Pressing Ctrl+C during a search query won't cancel the in-flight HTTP request, leaving the process hanging.
- **Missing `ShowDeleted(false)`**: Cancelled/deleted events may appear in search results, producing confusing output.

## Changes

Add `.ShowDeleted(false).Context(ctx)` to the search call to match the pattern used in the regular events listing path.

## Test plan

- [ ] Manual test: `gog calendar search "meeting"` still returns results
- [ ] Manual test: Ctrl+C during search cancels the request
- [ ] Verify deleted/cancelled events no longer appear in search results

### DIFF
--- a/internal/cmd/calendar_search.go
+++ b/internal/cmd/calendar_search.go
@@ -50,7 +50,9 @@ func (c *CalendarSearchCmd) Run(ctx context.Context, flags *RootFlags) error {
 		TimeMax(to).
 		MaxResults(c.Max).
 		SingleEvents(true).
-		OrderBy("startTime")
+		OrderBy("startTime").
+		ShowDeleted(false).
+		Context(ctx)
 
 	resp, err := call.Do()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


steipete:
Thanks for the PR. I’m closing this one because the patch is syntactically broken as submitted: it leaves an extra closing brace plus a stray `return nil` after `CalendarSearchCmd.Run`, so the package would not compile. The underlying idea is reasonable (`ShowDeleted(false)` + request context), but this PR needs to be recreated from a clean branch with the duplicate tail removed and a quick compile/test pass.

Issue #493, opened 2026-04-07
Calendar events display in wrong timezone (Pacific/Apia instead of system timezone)


ondohotola:
## Description

   When listing calendar events, `gog calendar events` displays times in Pacific/Asia (UTC+13) instead of the user's local timezone or the event's
 configured timezone.

   ## Steps to Reproduce

   1. Create a calendar event with Africa/Windhoek timezone (UTC+2):
      ```bash
      gog calendar create primary \
        --summary "Dr XXXXX Followup" \
        --from "2026-04-08T09:00:00+02:00" \
        --to "2026-04-08T09:20:00+02:00" \
        --location "Windhoek, Namibia"
 ```

 2. List events:
   ```bash
     gog calendar events --account <email> --from 2026-04-08 --to 2026-04-09
   ```

 Expected Behavior

 Event should display as:

 ```
   09:00–09:20  Dr XXXXX Followup
 ```

 Actual Behavior

 Event displays as:

 ```
   20:00–20:20+13:00  Dr XXXXX Followup
 ```

 Note: 09:00+02:00 and 20:00+13:00 represent the same instant, but the display timezone is incorrect.

 Environment

 - gog version: v0.12.0 (c18c58c 2026-03-09T05:53:14Z)
 - OS: macOS Darwin 25.4.0 (arm64)
 - System timezone: Africa/Windhoek (UTC+2)
 - gog calendar time output:
   ```
     timezone       Africa/Windhoek
     current_time   2026-04-07T22:22:21+02:00
   ```

 Additional Context

 - Event JSON from API shows correct timezone: "timeZone": "Etc/GMT-2"
 - Google Calendar web UI displays correct local time (09:00)
 - Only gog calendar events plain text output is wrong
 - JSON output (-j) may also be affected

 The issue appears to be in how gog converts and displays event times, not in how events are stored.

 Workaround

 Edit events directly in Google Calendar web UI, which displays times correctly, but this is counterproductive.

steipete:
Fixed on `main` in c4df6c3.

What changed:
- `calendar events` now resolves the calendar timezone for list output and formats table times using the calendar-local instant instead of preserving arbitrary API offsets like `+13:00`.
- JSON list output now includes the same timezone-aware `startLocal` / `endLocal` fields as `calendar event`; wrapper fields are explicitly marshaled so the Google API struct marshaler cannot drop them.
[...]

PR #509, opened 2026-04-20
fix(calendar): use buildEventDateTime in focus time command

dinakars777:
## Summary

The `calendar focus-time` command was constructing `EventDateTime` directly without calling `buildEventDateTime()`, which extracts IANA timezone names from RFC3339 offsets. This meant recurring focus-time events received no IANA timezone — only a numeric offset — causing DST-transition errors.

The `calendar out-of-office` command already uses `buildEventDateTime` correctly.

## Impact

Without IANA timezone info, Google Calendar uses a fixed offset for recurring focus-time events. When DST transitions occur, these events appear at the wrong time. For example, a recurring focus time at 9:00 AM CET would shift to 10:00 AM after a DST change because `Etc/GMT-1` doesn't account for CET/CEST.

## Changes

- Replace `&calendar.EventDateTime{DateTime: strings.TrimSpace(c.From)}` with `buildEventDateTime(c.From, false)` (and same for `c.To`)
- This brings focus time in line with the out-of-office command, which already uses `buildEventDateTime`

## Test plan

- [ ] Create a recurring focus time event and verify the IANA timezone is included in the API request
- [ ] Verify DST transitions are handled correctly for recurring focus time events

Related to #422, #423, #493

### DIFF
--- a/internal/cmd/calendar_focus_time.go
+++ b/internal/cmd/calendar_focus_time.go
@@ -36,8 +36,8 @@ func (c *CalendarFocusTimeCmd) Run(ctx context.Context, flags *RootFlags) error 
 
 	event := &calendar.Event{
 		Summary:      strings.TrimSpace(c.Summary),
-		Start:        &calendar.EventDateTime{DateTime: strings.TrimSpace(c.From)},
-		End:          &calendar.EventDateTime{DateTime: strings.TrimSpace(c.To)},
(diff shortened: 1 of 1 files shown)


steipete:

- Calendar commit: df3e97f
- Changelog thanks: 89916a5
- Gate: `make ci`

Thanks!

PR #510, opened 2026-04-20
fix(calendar): use midnight of next day as endOfDay to avoid truncation

### LINKED ISSUE #367: Bug: calendar events list returns 404 when using --today, --from, or --days flags
## Description

`gog calendar events list` returns `Google API error (404 notFound): Not Found` when any date-filtering flag is used (`--today`, `--from`, `--to`, `--days`), but works correctly without date flags.

## Repro

```bash
# Works - returns events
gog calendar events list --account oskar@firstmovr.com

# 404 - any date filter breaks it
gog calendar events list --account oskar@firstmovr.com --today
gog calendar events list --account oskar@firstmovr.com --from today --to tomorrow
gog calendar events list --account oskar@firstmovr.com --days 1
gog calendar events list --account oskar@firstmovr.com --from "2026-02-25T00:00:00-05:00"
```

## Environment

- gog version: 0.11.0 (Homebrew)
- macOS 15.x (arm64)
- Google Workspace account
- Calendar ID is the account email (not `primary` alias)
- `gog calendar calendars` shows only one calendar with ID = account email

## Notes

- The `--verbose` flag shows auth succeeds; the 404 comes from the Google Calendar API itself.
- Suspect the date flags cause a different API call path that uses `primary` as the calendarId, which doesn't resolve for this Workspace account.
- The non-date-filtered path presumably uses a different endpoint or defaults correctly.

dinakars777:
## Summary

`endOfDay()` previously returned `23:59:59.999999999`, but `FormatRFC3339()` uses `time.RFC3339` which truncates to second precision, producing `23:59:59`. Since the Google Calendar API's `timeMax` parameter is **exclusive**, events starting at exactly 23:59:59 are silently dropped from `--today`, `--week`, and `--days` queries.

## Impact

This is the root cause of issue #367 where `calendar events list` returns 404 or missing results when using `--today`, `--from`, or `--days` flags. Events near midnight are silently excluded.

## Changes

- Change `endOfDay()` to return midnight of the next day (`startOfDay(t.AddDate(0, 0, 1))`) instead of `23:59:59.999999999`
- This ensures RFC3339 second-precision formatting still includes the entire target day when used as an exclusive `timeMax`
- Update existing tests to match the new expected values

## Test plan

- [x] `go test ./internal/cmd/... -run TimeRange -v` — all pass
- [ ] Manual test: `gog calendar events list --today` includes events at 23:59
- [ ] Manual test: `gog calendar events list --week` covers the full week

Closes #367

### DIFF
--- a/internal/cmd/time_helpers.go
+++ b/internal/cmd/time_helpers.go
@@ -305,9 +305,12 @@ func startOfDay(t time.Time) time.Time {
 	return time.Date(t.Year(), t.Month(), t.Day(), 0, 0, 0, 0, t.Location())
 }
 
-// endOfDay returns the end of the day (23:59:59.999) in the given time's location.
+// endOfDay returns the start of the next day so that when formatted as
+// RFC3339 (second precision) and used as an exclusive timeMax, the entire
+// target day is included. Previously, 23:59:59.999999999 was truncated to
+// 23:59:59 by time.RFC3339, which excluded events starting at that second.
 func endOfDay(t time.Time) time.Time {
-	return time.Date(t.Year(), t.Month(), t.Day(), 23, 59, 59, 999999999, t.Location())
+	return startOfDay(t.AddDate(0, 0, 1))
 }
 
 // startOfWeek returns the start of the week for the given time.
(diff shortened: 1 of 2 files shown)


steipete:

- Calendar commit: df3e97f
- Changelog thanks: 89916a5
- Gate: `make ci`

Thanks!

PR #511, opened 2026-04-20
fix(gmail): skip charset re-decoding when body is already valid UTF-8

### LINKED ISSUE #446: bug(gmail): ISO-2022-JP emails garbled — decodeBodyCharset re-decodes UTF-8 bytes
## Summary

`gog gmail read` garbles emails whose MIME header declares `charset="iso-2022-jp"` (and likely other non-UTF-8 charsets).

**Environment:** gog v0.12.0, Linux arm64

## Steps to Reproduce

1. Receive an email with `Content-Type: text/plain; charset="iso-2022-jp"`.
2. Run `gog gmail read <threadId> -a <account>`.
3. Body text is garbled (`\ufffd` replacement characters).

## Root Cause

In `internal/cmd/gmail_thread.go`, `decodeBodyCharset()` checks the MIME `charset` and, if it's not UTF-8, re-decodes the bytes using that charset.

However, the Gmail API (`format=full`) **always normalizes `body.data` to UTF-8** before base64url-encoding, while preserving the original MIME headers verbatim. So after `decodeBase64URLBytes()` the bytes are already valid UTF-8, but `decodeBodyCharset` re-interprets them as ISO-2022-JP via `golang.org/x/text/encoding/ianaindex`, producing garbage.

## Proposed Fix

Add a `utf8.Valid()` guard before charset conversion:

```go
import "unicode/utf8"

func decodeBodyCharset(data []byte, contentType string) []byte {
    charsetLabel := charsetLabelFromContentType(contentType)
    normalized := strings.ToLower(strings.ReplaceAll(strings.TrimSpace(charsetLabel), "_", "-"))
    if charsetLabel == "" || normalized == "utf-8" || normalized == "utf8" {
        return data
    }
    if utf8.Valid(data) {
        return data
    }
    if decoded, ok := decodeWithCharsetLabel(data, charsetLabel); ok {
        return decoded
    }
    return data
}
```

This is safe because genuine ISO-2022-JP / Shift-JIS / EUC-JP raw bytes are almost never valid UTF-8, so the guard won't fire on non-API paths. For Gmail API JSON responses (the normal path), the bytes are always valid UTF-8 and the re-encoding is correctly skipped.

## Affected Encodings

Confirmed: `iso-2022-jp`. Likely also: `shift_jis`, `euc-jp`, `gb2312`, `gbk`, `euc-kr`, `windows-1252`, `iso-8859-1`.

<details><summary>Workaround (Python script)</summary>

Use `gog gmail read -j <threadId>` (JSON mode) and decode the base64 body as UTF-8 manually:

```python
#!/usr/bin/env python3
import json, sys, subprocess, base64

cmd = ['gog', 'gmail', 'read', '-j', sys.argv[1]]
if len(sys.argv) > 2:
    cmd += ['-a', sys.argv[2]]

data = json.loads(subprocess.run(cmd, capture_output=True).stdout)
for msg in data.get('thread', data).get('messages', []):
    hdrs = {h['name']: h['value'] for h in msg['payload'].get('headers', [])}
    print(f"From: {hdrs.get('From','')}  Date: {hdrs.get('Date','')}")
    print(f"Subject: {hdrs.get('Subject','')}\n")
    def find_text(p):
        if 'text/plain' in p.get('mimeType',''):
            return p
        for sub in p.get('parts', []):
            r = find_text(sub)
            if r: return r
    tp = find_text(msg['payload'])
    if tp and tp.get('body',{}).get('data'):
        print(base64.urlsafe_b64decode(tp['body']['data'] + '==').decode('utf-8', errors='replace'))
    print('-' * 40)
```

</details>

dinakars777:
## Summary

`decodeBodyCharset` re-decodes already-UTF-8 data as the original charset (e.g., ISO-2022-JP, Shift-JIS, GBK) when the MIME header declares a non-UTF-8 charset, producing garbled output. The Gmail API normalizes `body.data` to UTF-8 before base64url-encoding but preserves the original MIME headers verbatim.

## Steps to Reproduce (before fix)

1. Receive an email with `Content-Type: text/plain; charset="iso-2022-jp"`
2. Run `gog gmail read <threadId>`
3. Body text is garbled (replacement characters `\ufffd`)

## Changes

Add a `utf8.Valid(data)` guard before charset conversion. If the data is already valid UTF-8, skip re-decoding. Genuine non-UTF-8 raw bytes (like ISO-2022-JP or Shift-JIS) are almost never valid UTF-8, so this guard is safe.

## Affected Encodings

Confirmed: `iso-2022-jp`. Likely also: `shift_jis`, `euc-jp`, `gb2312`, `gbk`, `euc-kr`, `windows-1252`, `iso-8859-1`.

## Test plan

- [x] `go test ./internal/cmd/... -run GmailGet -v` passes
- [ ] Manual test: read an ISO-2022-JP encoded email and verify it displays correctly

Closes #446

### DIFF
--- a/internal/cmd/gmail_thread.go
+++ b/internal/cmd/gmail_thread.go
@@ -14,6 +14,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"unicode/utf8"
 
 	"golang.org/x/net/html/charset"
 	"golang.org/x/text/encoding/ianaindex"
@@ -493,6 +494,15 @@ func decodeBodyCharset(data []byte, contentType string) []byte {
 	charsetLabel := charsetLabelFromContentType(contentType)
 	normalized := strings.ToLower(strings.ReplaceAll(strings.TrimSpace(charsetLabel), "_", "-"))
 	if charsetLabel == "" || normalized == "utf-8" || normalized == "utf8" {
+		return data
+	}
+	// The Gmail API normalizes body.data to UTF-8 before base64url-encoding,
+	// but preserves the original MIME Content-Type headers verbatim. If the
+	// data is already valid UTF-8, skip charset re-decoding to avoid garbling
+	// content (e.g., ISO-2022-JP or Shift-JIS messages already converted by
+	// the API). Genuine non-UTF-8 raw bytes are almost never valid UTF-8, so
+	// this guard is safe.
+	if utf8.Valid(data) {
 		return data
 	}
 	if decoded, ok := decodeWithCharsetLabel(data, charsetLabel); ok {


steipete:
Landed manually on `main` as part of a grouped Gmail fix commit.

- Gmail commit: 0aa06b4
- Changelog thanks: 89916a5
- Gate: `make ci`

Note: review kept the UTF-8 guard but preserved ISO-2022-JP escape-sequence decoding, with regression coverage.

Thanks!

PR #514, opened 2026-04-21
fix(gmail): use configured timezone for Date header (#472)

### LINKED ISSUE #472: [Bug] gmail send --reply-to sets wrong timezone (+0900) in Date header causing delivery failure
## Bug Description

When using `gog gmail send --reply-to <messageId>`, the resulting email has an incorrect timezone in the `Date` header (+0900, Tokyo timezone), which causes Google to reject the message with:

```
An error occurred. Your message was not sent.
```

## Steps to Reproduce

1. Have an existing email thread with a message ID
2. Run: `gog gmail send --to recipient@example.com --subject "Re: Subject" --body "..." --reply-to <messageId>`
3. The sent email will have a Date header like: `Tue, 24 Mar 2026 19:13:39 +0900` (Tokyo time) instead of the correct local timezone or UTC

## Expected Behavior

The `Date` header should use the system timezone (or UTC) — not hardcoded +0900.

## Actual Behavior

Email Date header is set to +0900 (Tokyo/JST), causing Google's delivery to fail with a generic error message.

## Workaround

Do **not** use `--reply-to` flag. Send as a new email with a manually prefixed subject (e.g. `Re: Original Subject`). The email sends successfully without this flag.

## Environment

- gog version: v0.12.0 (c18c58c 2026-03-09T05:53:14Z)
- OS: Linux (Ubuntu 24.04)
- Using OAuth credentials

## Impact

The `--reply-to` flag is effectively broken — using it causes email delivery to silently fail with a non-descriptive error message.

dinakars777:
Fixes #472.

Instead of directly invoking `time.Now().Format()` using the system's runtime timezone, we now properly fetch the globally configured timezone through `getConfiguredTimezone`. This ensures the `Date` header correctly aligns with the application-wide explicit configuration and falls back to `time.Local` correctly if not set, avoiding the `+0900` timezone delivery misconfiguration.

### DIFF
--- a/internal/cmd/gmail_mime.go
+++ b/internal/cmd/gmail_mime.go
@@ -86,7 +86,16 @@ func buildRFC822(opts mailOptions, cfg *rfc822Config) ([]byte, error) {
 		return nil, fmt.Errorf("invalid Subject: %w", err)
 	}
 	writeHeader(&b, "Subject", encodeHeaderIfNeeded(opts.Subject))
-	writeHeader(&b, "Date", time.Now().Format(time.RFC1123Z))
+
+	loc, err := getConfiguredTimezone("")
+	if err != nil {
+		return nil, err
+	}
+	if loc == nil {
+		loc = time.Local
+	}
+	writeHeader(&b, "Date", time.Now().In(loc).Format(time.RFC1123Z))
+
 	if !hasHeader(opts.AdditionalHeaders, "Message-ID") && !hasHeader(opts.AdditionalHeaders, "Message-Id") {
 		messageID, err := randomMessageID(opts.From)
 		if err != nil {


steipete:
Landed manually on `main` in 2c48bd8, with lint cleanup in 6bdc215.

I kept the intended fix: outbound RFC822 `Date` headers now use the configured timezone (`GOG_TIMEZONE` / `default_timezone`) and fall back to local time when none is configured.

I also added regression coverage in `internal/cmd/gmail_mime_more_test.go` for a host-local JST process with `GOG_TIMEZONE=UTC`, asserting the generated `Date` header uses `+0000`.

Verified locally:
- `go test ./internal/cmd -run 'TestBuildRFC822|TestGmailSend|TestGmailDraft' -count=1`
- `make ci`

Thanks @dinakars777.

Issue #86, opened 2026-01-19
OAuth tokens not persisted across upgrades (keyring empty after brew upgrade)


alsoft-alex:
Description:

After upgrading gogcli via brew upgrade, the OAuth tokens are lost and require re-authentication (I use clawdbot).

Steps to reproduce:
Authenticate with gog auth add user@gmail.com --services calendar,gmail
Verify auth works: gog auth list shows the account
Run brew upgrade steipete/tap/gogcli
Run gog auth list — hangs or shows no accounts

Observed behavior:
~/Library/Application Support/gogcli/keyring/ directory is empty after upgrade
gog auth list hangs indefinitely after upgrade (requires re-auth)

Expected behavior:
OAuth tokens should persist across version upgrades

Environment:
macOS 14.x (Sonoma)
gogcli 0.7.0 → 0.8.0 (also happened on earlier upgrades)
Installed via Homebrew

Workaround:
Re-authenticate after each upgrade:
gog auth add user@gmail.com --services calendar,contacts,drive,gmail

PR #515, opened 2026-04-21
fix(secrets): timeout keyring.Open() on macOS to prevent indefinite hang (#513)

### LINKED ISSUE #513: v0.13.0: keyring backend hangs indefinitely on macOS — all auth-requiring commands SIGKILL'd (rollback to v0.12.0 restores)
## Bug: v0.13.0 keyring backend hangs indefinitely on macOS (commands SIGKILL'd)

After upgrading from v0.12.0 → v0.13.0 via `brew upgrade steipete/tap/gogcli`, **every command that touches the keyring hangs indefinitely** and is eventually killed with SIGKILL. This includes `gog auth list`, `gog gmail search`, `gog gmail autoreply`, etc. — any command that needs to read stored OAuth tokens.

Rolling back to v0.12.0 (same OAuth tokens in the macOS Keychain) restores full functionality instantly.

## Reproduction

1. Upgrade: `brew upgrade steipete/tap/gogcli` (from v0.12.0 → v0.13.0)
2. Run any auth-requiring command:
   ```
   gog auth list
   gog gmail search -a me@example.com "newer_than:1h" --max 5
   ```
3. Command hangs. After 10–60s, exits with SIGKILL (no output, no prompt, no error message). No macOS Keychain prompt appears.
4. `gog --version` works fine (no keyring access needed).
5. Also tested `--no-input` flag — same hang.

## What I tried

- `gog auth list --no-input` → hangs, SIGKILL
- `GOG_KEYRING_BACKEND=file gog auth list` → returns **instantly** with "No tokens stored" (file backend works, but the tokens are in macOS Keychain, so this isn't a fix)
- Flag `--keyring-namespace` rejected with "unknown flag" (release notes mention keyring namespace overrides landed in v0.13.0 — the flag name may have changed?)
- Checked release notes for breaking changes around keyring/auth — didn't find anything explicit for macOS

## Rollback

Installing v0.12.0 from tap git history restored full functionality:
```bash
cd $(brew --repository)/Library/Taps/steipete/homebrew-tap
git show 7a44daf:Formula/gogcli.rb > /tmp/gogcli-0.12.rb
brew uninstall gogcli
cp /tmp/gogcli-0.12.rb Formula/gogcli.rb  # temporarily pin
brew install steipete/tap/gogcli
```

Exact same OAuth tokens in Keychain, exact same accounts (`info@booklocal.in`, `lease@membnb.com`) — `gog auth list` on v0.12.0 returns in < 1 second. Upgrade to v0.13.0 → hang. Revert to v0.12.0 → works again.

## Environment

- **macOS:** 15.3.1 (build 24D70) — Apple Silicon (arm64)
- **gogcli:** v0.13.0 (5cd913e 2026-04-20T22:14:47Z) — installed via `brew install steipete/tap/gogcli`
- **Shell:** zsh
- **Keyring backend setting:** `auto` (default, per `gog --help` output)

## Hypothesis

The "keyring namespace overrides" refactor in v0.13.0 (#463) may have changed how the `auto` keyring backend probes macOS Keychain, possibly waiting on a prompt that never surfaces in non-TTY contexts (or maybe always, given this fails in interactive shell too).

## What would help

1. A `--keyring-backend file` flag or `GOG_KEYRING_BACKEND=file` support that gracefully migrates existing tokens from macOS Keychain → file on first use
2. Or a debug/verbose flag that surfaces the keyring operation being attempted (so we can see where the hang is)
3. Or fix the `auto` backend to not hang on macOS when Keychain interaction is needed non-interactively

Happy to provide additional logs/traces if useful — just let me know what would help.

sardoru:
Fixes #513.

## The bug

After upgrading gogcli via Homebrew, every command that reads stored tokens (`gog auth list`, `gog gmail search`, etc.) hangs indefinitely on macOS and is eventually SIGKILL'd — no output, no prompt, no actionable error.

The root cause is that the `keyring.Open()` timeout safety net introduced for headless Linux (#XXX) only fires when `goos == "linux"`. On macOS, the Security framework waits for a Keychain GUI permission prompt that can't surface in non-interactive contexts (cron jobs, agents, subprocesses) — or, per the existing comment referencing #86, after a Homebrew upgrade installs a new binary hash and the previous "Always Allow" grant no longer applies. The process just hangs.

## The fix

Extend `shouldUseKeyringTimeout` to include `darwin` when the backend is `auto` or `keychain`. If the Keychain doesn't respond within the timeout, return a clear error with platform-specific guidance instead of hanging forever.

Also bump the timeout from 5s → 10s to tolerate slower macOS Keychain responses on first post-upgrade access (still short enough to feel like a real error, long enough to not false-positive on a cold keychain).

## Error message before

```
$ gog auth list
(hangs forever, then SIGKILL with no output)
```

## Error message after

```
$ gog auth list
Error: keyring open timed out after 10s (macOS Keychain may be waiting for a
permission prompt — run `gog auth list` from a terminal and click "Always Allow"
when prompted (common after Homebrew upgrades; see
https://github.com/steipete/gogcli/issues/86)); set GOG_KEYRING_BACKEND=file
and GOG_KEYRING_PASSWORD=<password> to use encrypted file storage instead
```

## What's covered

| Platform / backend       | Pre-PR timeout | Post-PR timeout |
|--------------------------|----------------|-----------------|
| Linux auto + D-Bus       | ✅ 5s          | ✅ 10s          |
| Linux auto, no D-Bus     | n/a (forced to file) | n/a    |
| Linux explicit file      | n/a            | n/a             |
| **macOS auto**           | ❌ hangs        | ✅ 10s          |
| **macOS keychain**       | ❌ hangs        | ✅ 10s          |
| macOS explicit file      | n/a            | n/a             |
| Windows auto             | n/a            | n/a             |

## Tests

- Extends `TestKeyringDbusGuards` with three darwin cases: `auto` → timeout, `keychain` → timeout, `file` → no timeout
- New `TestKeyringTimeoutHint` verifies platform-specific hint text for darwin/linux/windows
- Existing `TestOpenKeyringWithTimeout_Timeout` still passes (my change kept the `GOG_KEYRING_BACKEND=file` fallback text the test asserts on)

## Why 10s instead of 5s

On macOS, Keychain can briefly block when it's cold (first access after login, or first access after a binary-hash change). 5s was occasionally false-positive even when the prompt would eventually surface. 10s is comfortably long enough for a real prompt to complete while still failing fast on a deadlocked Security framework.

## Caveat / follow-up

The hang is really in `keyring.Keys()` on subsequent calls if the permission grant isn't cached — but in practice `keyring.Open()` probes the backend during construction, which is where the GUI prompt is triggered, so timing out `Open()` catches almost all real-world cases. A full-coverage fix would also wrap `Keys()`/`Get()` with a timeout, but that's more invasive and deserves a separate PR.

## Discovery

I hit this the night after #513 — upgraded from v0.12.0 to v0.13.0 via `brew upgrade`, and every `gog` command that touched auth hung. Rolled back to v0.12.0 (same OAuth tokens in Keychain), everything worked instantly again. Didn't re-upgrade until I could trace it.

Tested manually on:
- macOS 15.3.1 (build 24D70) — Apple Silicon (arm64)
- Existing keychain items from a prior v0.12.0 install

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+- Secrets: extend the `keyring.Open()` timeout safety net to macOS when the backend is `auto` or `keychain`. Previously only Linux used a timeout, so a macOS Security framework prompt that couldn't surface (e.g., non-interactive runs, post-Homebrew-upgrade re-authorization) would hang every auth-requiring command indefinitely until the process was killed. The timeout now also covers macOS and returns a clear error with guidance to run from a terminal and click "Always Allow", or switch to `GOG_KEYRING_BACKEND=file`. Bumps the timeout from 5s to 10s to tolerate slower macOS Keychain responses. Fixes #513.
+
 ## 0.13.0 - 2026-04-20
 
 ### Highlights
--- a/internal/secrets/store.go
+++ b/internal/secrets/store.go
@@ -156,14 +156,33 @@ func keyringServiceName() string {
 // keyringOpenTimeout is the maximum time to wait for keyring.Open() to complete.
 // On headless Linux, D-Bus SecretService can hang indefinitely if gnome-keyring
 // is installed but not running.
-const keyringOpenTimeout = 5 * time.Second
+//
+// On macOS, keyring.Open() can hang when the Security framework waits for a GUI
+// permission prompt that cannot surface (e.g., non-interactive runs after a
+// Homebrew upgrade installs a new binary hash, breaking the previous
+// "Always Allow" grant).
+const keyringOpenTimeout = 10 * time.Second
 
 func shouldForceFileBackend(goos string, backendInfo KeyringBackendInfo, dbusAddr string) bool {
 	return goos == "linux" && backendInfo.Value == keyringBackendAuto && dbusAddr == ""
 }
 
+// shouldUseKeyringTimeout returns true when keyring.Open() should be wrapped
+// in a timeout to prevent indefinite hangs.
+//
+//   - Linux with "auto" backend and a D-Bus session: SecretService may be
+//     installed but unresponsive (e.g., gnome-keyring not running).
+//   - macOS with "auto" or "keychain" backend: Security framework may block
+//     on a GUI permission prompt that cannot surface in non-interactive
+//     contexts or after a Homebrew upgrade.
 func shouldUseKeyringTimeout(goos string, backendInfo KeyringBackendInfo, dbusAddr string) bool {
-	return goos == "linux" && backendInfo.Value == "auto" && dbusAddr != ""
+	if goos == "linux" && backendInfo.Value == "auto" && dbusAddr != "" {
+		return true
+	}
+	if goos == "darwin" && (backendInfo.Value == keyringBackendAuto || backendInfo.Value == "keychain") {
+		return true
+	}
+	return false
 }
 
 func openKeyring() (keyring.Keyring, error) {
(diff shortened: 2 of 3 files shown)


steipete:
Thanks @sardoru. I landed this as a maintainer rewrite on main in 6430dd1, with lint/test cleanup in c2ea4f5.

I changed the implementation from an `Open()` timeout to operation-level timeouts for macOS keyring reads/writes/lists. The macOS keyring backend constructs the object in `Open()`; the blocking Security framework calls happen on `Keys` / `Get` / `Set` / `Remove`, so this covers the actual hang point. Added tests, docs, and changelog.

Closing this as landed.

PR #517, opened 2026-04-21
fix(docs): add TabId support to markdown replace and improve error logging

### LINKED ISSUE #494: Bug: docs find-replace --format markdown silently does nothing
## Summary

`docs find-replace --format markdown` silently does nothing — no error, no replacement, no output. The same command works correctly with `--format plain`.

## Steps to Reproduce

```bash
# 1. Write a placeholder to an existing doc
gog docs write <docId> --text "REPLACE_ME"

# 2. Try to replace with markdown formatting
gog docs find-replace <docId> "REPLACE_ME" "## Heading" --format markdown --first -v

# Output: only debug lines, no replacement count, no error
# time=... level=DEBUG msg="creating client options with custom scopes" ...
# time=... level=DEBUG msg="client options with custom scopes created successfully" ...

# 3. Verify — placeholder is still there
gog docs cat <docId>
# Output: REPLACE_ME
```

## Expected Behavior

The placeholder text should be replaced with the markdown content, rendered as rich text (heading formatting applied).

## Actual Behavior

- No replacement occurs
- No error message
- No output (no `replacements` count shown)
- The document remains unchanged

## Working Comparison

```bash
# This works correctly:
gog docs find-replace <docId> "REPLACE_ME" "replaced" --format plain --first
# Output: replacements  1
```

## Also tested with `--content-file`

```bash
gog docs find-replace <docId> "REPLACE_ME" --format markdown --first --content-file ./content.md
```

Same result — silently does nothing.

## Environment

- gogcli version: v0.12.0 (c18c58c 2026-03-09)
- OS: macOS (arm64)

## Related

- #272 — `docs write` also lacks markdown rendering support

dinakars777:
Fixes #494

## Bug
`gog docs find-replace <docId> "find" "replace" --format markdown` silently does nothing when using a tab.

## Root Cause
The markdown replace path (`replaceDocsMarkdownRange`) was missing TabId support:
1. `InsertTextRequest` did not have TabId set
2. `cleanupDocsImagePlaceholders` did not use TabsCriteria
3. No warning was shown when tab targeting would fail

Additionally, `DeleteContentRangeRequest.Range` does not support TabId in the Google Docs API - this is a fundamental limitation.

## Fix
1. Add TabId to `InsertTextRequest.Location` in `replaceDocsMarkdownRange`
2. Add TabsCriteria to `ReplaceAllText` in `cleanupDocsImagePlaceholders` 
3. Pass TabId through the call chain: `runMarkdown` → `replaceDocsMarkdownRange`
4. Add warning when using `--format markdown` with `--tab-id` (explaining the DeleteContentRange limitation)
5. Improve error messages to include match index on failure

## Testing
- Lint passes (0 issues)
- Existing tests pass (pre-existing 404 failures in mock server are unrelated)

### DIFF
--- a/internal/cmd/docs_edit.go
+++ b/internal/cmd/docs_edit.go
@@ -526,9 +526,28 @@ func (c *DocsFindReplaceCmd) Run(ctx context.Context, flags *RootFlags) error {
 	}
 
 	matches := findTextMatches(targetDoc, c.Find, c.MatchCase)
+	if len(matches) == 0 {
+		fmt.Fprintf(os.Stderr, "warning: no matches found for %q in locally parsed document (--format markdown uses local parsing); the text may be in a nested structure or have encoding differences\n", c.Find)
+		if outfmt.IsJSON(ctx) {
+			return outfmt.WriteJSON(ctx, os.Stdout, map[string]any{
+				"documentId":   docID,
+				"find":         c.Find,
+				"replace":      replaceText,
+				"replacements": 0,
+			})
+		}
+		u.Out().Printf("documentId\t%s", docID)
+		u.Out().Printf("find\t%s", c.Find)
+		u.Out().Printf("replace\t%s", replaceText)
+		u.Out().Printf("replacements\t0")
+		return nil
+	}
+	if c.TabID != "" && format == docsContentFormatMarkdown {
+		fmt.Fprintf(os.Stderr, "warning: --format markdown with --tab-id uses local parsing; DeleteContentRange does not support tab targeting, replacements may not apply to the correct tab\n")
+	}
 	for i := len(matches) - 1; i >= 0; i-- {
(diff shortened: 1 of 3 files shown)


steipete:
Thanks for the PR. I reviewed this against current `main`; I'm going to close this one rather than land it as-is.

Two blockers:

- The Gmail draft commit is unrelated here and current `main` already has the #512/#528 threading fix via the newer reply-target path.
- The Docs tab change is stale after the markdown-image work and is still partial: it adds warnings for tabbed markdown replacement instead of proving the range/image cleanup path is correctly tab-scoped. This needs a fresh, focused PR with only the Docs tab fix and regression coverage for `docs find-replace --format markdown --tab-id`.

Happy to review a rebased/split version, but this mixed patch is no longer a good landing unit.

steipete:
Closing per review above: mixed stale patch; needs a fresh focused Docs-tab PR.

PR #518, opened 2026-04-21
fix(docs): size markdown images on write

vinothd-oai:
## Summary
- extract Markdown image references before `gog docs write --replace --markdown` uploads content through Drive
- upload cleaned Markdown with placeholders, then reuse the existing Docs image insertion path
- preserve remote image support, local image resolution relative to `--file`, default sizing, custom width/height attrs, pageless behavior, and existing output shape

## Why
`gog docs create --file` already handles Markdown images in two passes: it strips image Markdown into placeholders before Drive import, then reads the converted Google Doc and inserts images through the Docs API with explicit sizing. That keeps large source images from being imported at their natural dimensions and gives users consistent defaults such as the 468pt max width.

`gog docs write --replace --markdown` was taking a different path. It sent the raw Markdown body directly to Drive update, which let Drive import image references itself. That bypassed the existing sizing logic, so large images could become oversized in the resulting Google Doc.

This change makes the write path follow the same pattern as document creation: extract images first, let Drive convert the cleaned Markdown, then reuse `insertImagesIntoDocs` to place remote or local images with the same default and explicit width/height handling.

## Validation
- `go test ./internal/cmd -run "TestDocsWrite_Markdown"`
- `go test ./internal/cmd -run "TestDocsWrite"`
- `git diff --check`

Note: `go test ./internal/cmd` still fails locally on an unrelated auth-state-sensitive `TestRequireAccount_Missing` because this environment has `actual@example.com` configured.

### DIFF
--- a/internal/cmd/docs_edit.go
+++ b/internal/cmd/docs_edit.go
@@ -11,6 +11,7 @@ import (
 	"google.golang.org/api/drive/v3"
 	gapi "google.golang.org/api/googleapi"
 
+	"github.com/steipete/gogcli/internal/config"
 	"github.com/steipete/gogcli/internal/outfmt"
 	"github.com/steipete/gogcli/internal/ui"
 )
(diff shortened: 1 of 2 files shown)


steipete:
Thanks @vinothd-oai, landed this on `main` via `e97cad6` with follow-up live-test hardening in `acb05f6`.

I kept the core fix: `docs write --replace --markdown` now extracts remote Markdown images before Drive conversion and reinserts them through Docs so sizing is applied consistently. During live testing, Google Docs rejected temporary Drive-hosted local images even after sharing/retry, so I adjusted the landed version to return a clear error for local image paths and document public HTTPS image URLs as the supported path.

Validation performed:
- `make ci`
- live reauth for `steipete@gmail.com` with docs/drive/gmail/people scopes
- live Docs remote-image write smoke with placeholder removal + inline object check
- live local-image clear-error smoke
- live Gmail draft create/get/delete smoke.

PR #520, opened 2026-04-22
feat(gmail): add --body-format flag to messages search

alexknowshtml:
## Summary

Adds `--body-format` to `gmail messages search` for controlling which MIME part is returned when `--include-body` is used.

- `--body-format text` (default) — preserves existing behavior, returns plain text part (falls back to HTML if no plain text)
- `--body-format html` — prefers the HTML part (falls back to plain text if no HTML)

Adds `bestBodyHTML()` helper to complement `bestBodyText()`.

## Motivation

Newsletters and rich emails often have a minimal plain text part (or just "view in browser") while all meaningful content is in the HTML part. This flag makes it possible to retrieve that content without custom workarounds.

## Usage

```
gog gmail messages search "from:newsletter@example.com" --include-body --body-format html
```

## Test plan

- [ ] `--body-format text` returns existing behavior (plain text, HTML fallback)
- [ ] `--body-format html` returns HTML part when present
- [ ] `--body-format html` falls back to plain text when no HTML part exists
- [ ] Omitting `--body-format` defaults to `text` behavior
- [ ] Invalid value is rejected by kong enum validation


steipete:
Thanks @alexknowshtml, landed on main as d08771e plus maintainer follow-up 6e0af2c.

Before landing I kept the feature, removed the accidental byline-cache churn via your second commit, and added:
- explicit body-format constants/helper plumbing instead of a variadic test dodge
- regression coverage for default text body selection, HTML preference, and plain fallback
- README, spec, and changelog coverage

Verified:
- focused Gmail body-format tests
- make ci
- live Gmail smoke on steipete@gmail.com with --body-format text and html; only counts/body lengths were printed, and the HTML path returned an HTML-looking body.

PR #523, opened 2026-04-23
fix: support client aliases for subdomain email accounts

DrBaher:
## Problem

When using `gog auth add` with a client alias for an account whose email has a subdomain (e.g. `user@sub.company.ai`), auth fails with:

```
authorized as user@sub.company.ai, expected company.ai
```

This affects any Google Workspace account on a subdomain domain (e.g. `baher@medicus.ai` using client alias `baher.ai`). Two bugs combine to cause this:

1. **`auth_add.go`**: The OAuth-authorized email is compared against the alias name, which never matches a bare client name, so the token is never saved.
2. **`clients.go`**: `ResolveClientForAccount` treats any input without `@` as a malformed email with no domain, falling through to `DefaultClientName` instead of treating it as a direct client name.

## Fix

- **`auth_add.go`**: Skip the email match check when `c.Email` has no `@` — the caller is intentionally using a client alias, not an email address.
- **`clients.go`**: Treat any input without `@` as a direct client name rather than attempting email/domain parsing.

## Usage after fix

```bash
# One-time setup in config.json:
# { "account_clients": { "user@sub.company.ai": "company.ai" } }

gog auth add company.ai --services gmail,calendar,drive,...
# sign in as user@sub.company.ai in the browser

gog gmail search "newer_than:1d" --account user@sub.company.ai  # works
```

## Tests

Added two new test cases:
- `TestAuthAddCmd_ClientAliasSkipsEmailCheck` — verifies auth succeeds when a client alias is passed
- `ResolveClientForAccount/client name without @ is used as-is` — verifies bare client name is returned directly
- `ResolveClientForAccount/subdomain email routes via account_clients mapping` — verifies subdomain email routing via config

All existing tests pass.

### DIFF
--- a/internal/cmd/auth_add.go
+++ b/internal/cmd/auth_add.go
@@ -248,8 +248,12 @@ func (c *AuthAddCmd) Run(ctx context.Context, flags *RootFlags) error {
 	if err != nil {
 		return fmt.Errorf("fetch authorized email: %w", err)
 	}
-	if normalizeEmail(authorizedEmail) != normalizeEmail(c.Email) {
-		return fmt.Errorf("authorized as %s, expected %s", authorizedEmail, c.Email)
+	// If c.Email looks like a client name (no "@"), skip the email match check —
+	// the user intentionally authorized with a client alias.
+	if strings.Contains(normalizeEmail(c.Email), "@") {
+		if normalizeEmail(authorizedEmail) != normalizeEmail(c.Email) {
+			return fmt.Errorf("authorized as %s, expected %s", authorizedEmail, c.Email)
+		}
 	}
 
 	store, err := openSecretsStore()
--- a/internal/cmd/auth_add_test.go
+++ b/internal/cmd/auth_add_test.go
@@ -200,6 +200,38 @@ func TestAuthAddCmd_KeepRejected(t *testing.T) {
 	}
 }
 
+func TestAuthAddCmd_ClientAliasSkipsEmailCheck(t *testing.T) {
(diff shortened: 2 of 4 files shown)


steipete:
Thanks for the PR. I reviewed this against current `main`, and I don't think we should land this shape as-is.

The risky bit is overloading the positional `auth add <email>` value: changing `ResolveClientForAccount` so any non-email account string is treated as a client alias broadens behavior outside auth-add and can mask malformed account input. It also does not fully solve the later command path unless the actual authorized email gets an account-client mapping.

The safer supported route today is explicit client selection, e.g. `gog --client company.ai auth add user@sub.company.ai ...`, which keeps the browser login email and stored account mapping unambiguous. If we want this UX, I would rather see it reworked around explicit `--client` / alias handling (or a dedicated flag) plus tests that prove the authorized email is mapped and that generic account resolution does not reinterpret arbitrary non-email account strings.

steipete:
Thanks again for the PR. I’m going to close this shape rather than keep it hanging.

The bug report is understandable, but overloading `auth add <email>` so a non-email positional value becomes a client alias changes generic account/client resolution in a risky way and can mask malformed account inputs. The supported safe path is explicit client selection with the real account email, for example:

```bash
gog --client company.ai auth add user@sub.company.ai --services gmail,calendar,drive
```

That keeps the browser login email, stored token key, and client bucket unambiguous. If we add an alias-first auth UX later, I’d want it behind an explicit flag or command path that stores/maps the authorized email deliberately rather than changing `ResolveClientForAccount` globally.

PR #524, opened 2026-04-24
fix(drive): include driveId in field masks for list/search/get

LeanSheng:
## Summary
- Adds `driveId` to the shared field mask in `driveFileListFields` (drives `drive ls`, `drive search`) and to the inline `Fields(...)` call in `DriveGetCmd.Run` (`drive get`).
- The Drive API only returns fields explicitly requested in `fields`; with `driveId` omitted, JSON output never included it even for files living in a Shared Drive, making it impossible to distinguish My Drive vs. Shared Drive files from `gog` output alone.
- For My Drive files the API omits the field, so existing consumers are unaffected. Shared Drive files now carry `driveId` through.
- Adds `internal/cmd/drive_driveid_field_test.go` pinning `driveId` in the shared mask constant.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/cmd/... -count=1 -timeout 60s`
- [x] `go test ./internal/cmd/... -run TestDriveFileListFields -count=1 -v` (new test passes)
- [x] Manual spot check: `gog drive ls --all-drives --format=json` on an account with Shared Drive access shows `driveId` for Shared Drive files and omits it for My Drive files
- [x] Manual spot check: `gog drive get <sharedDriveFileId> --format=json` includes `driveId`

### DIFF
--- a/internal/cmd/drive.go
+++ b/internal/cmd/drive.go
@@ -123,7 +123,7 @@ func (c *DriveGetCmd) Run(ctx context.Context, flags *RootFlags) error {
 
 	f, err := svc.Files.Get(fileID).
 		SupportsAllDrives(true).
-		Fields("id, name, mimeType, size, modifiedTime, createdTime, parents, webViewLink, description, starred").
+		Fields("id, name, mimeType, size, modifiedTime, createdTime, parents, webViewLink, description, starred, driveId").
 		Context(ctx).
 		Do()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


steipete:
Landed directly on main as 3f46ddf (with changelog entry in ee9b552). Thanks @LeanSheng.

steipete:
Closing as landed on main.

PR #525, opened 2026-04-24
feat(drive): add --drive and --parent flags to drive search

LeanSheng:
## Summary
- Adds two flags to `gog drive search`:
  - `--drive <driveId>` — scopes to a specific shared drive via `corpora=drive&driveId=X` (the Drive API's proper pattern for recursive within-drive search). Mutually exclusive with `--no-all-drives`.
  - `--parent <folderId>` — wraps the query with `'<folderId>' in parents`, matching `drive ls --parent` semantics (direct children only, since the Drive API doesn't support recursive parent queries). Mutually exclusive with `--raw-query`.
- Extends `driveFilesListCallWithDriveSupport` with a `driveID` arg (sole non-test caller, `listDriveFiles`, passes it through from the new `driveFileListOptions.driveID`).
- `DriveSearchCmd.Run` validates flag combinations up front, before any API call.
- Default behavior unchanged — no flags means same corpora=allDrives request as before.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/cmd/... -count=1 -timeout 60s` (full suite green)
- [x] Focused run: `TestDriveSearchCmd_WithDrive`, `..._WithParent`, `..._DriveAndParent_Combine`, `..._DriveAndNoAllDrives_Conflicts`, `..._ParentAndRawQuery_Conflicts` — all pass
- [x] Manual: `gog drive search "report" --drive $(gog drive drives --format=json | jq -r '.drives[0].id')` returns results from that shared drive only
- [x] Manual: `gog drive search "report" --parent <folderId>` returns only direct children of that folder
- [x] Manual: `gog drive search foo --drive X --no-all-drives` fails fast with a clear error

### DIFF
--- a/internal/cmd/drive.go
+++ b/internal/cmd/drive.go
@@ -99,6 +99,8 @@ type DriveSearchCmd struct {
 	Max       int64    `name:"max" aliases:"limit" help:"Max results" default:"20"`
 	Page      string   `name:"page" aliases:"cursor" help:"Page token"`
 	AllDrives bool     `name:"all-drives" help:"Include shared drives (default: true; use --no-all-drives for My Drive only)" default:"true" negatable:"_"`
+	Drive     string   `name:"drive" aliases:"drive-id" help:"Scope search to a specific shared drive (uses corpora=drive with driveId). Mutually exclusive with --no-all-drives. Pass the driveId from 'gog drive drives'."`
+	Parent    string   `name:"parent" help:"Scope search to direct children of a specific folder or shared drive. Wraps the query with \"'<parentId>' in parents\"."`
 }
 
 type DriveGetCmd struct {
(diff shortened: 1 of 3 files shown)


steipete:
Thanks @LeanSheng. Landed on main as 80039c8, with maintainer cleanup in 617293d.

What changed before landing:
- kept --drive and --parent for drive search
- fixed the shared Drive list helper callers used by backup code
- trimmed/escaped the new ID inputs before passing them into Drive query/API params
- added changelog credit

Verified:
- go test ./internal/cmd focused docs/drive suite
- make ci
- live smoke on steipete@gmail.com: drive search --parent root found the temp doc, --parent + --raw-query rejected locally, and drive search --drive 0AD1UuQS84YLYUk9PVA exercised the real shared-drive API path

Issue #395, opened 2026-03-03
fix: Windows test failures in keyring, config, secrets, and path expansion


visionik:
## Summary

9 tests are currently skipped on Windows CI (added in #388) because the underlying code assumes macOS/Linux:

### Keyring / Auth (3 tests)
- `TestAuthAddCmd_JSON`
- `TestAuthListRemoveTokensListDelete_JSON`
- `TestAuthServicesCmd_JSON`

These fail because the keyring backend expects macOS Keychain or Linux keyring. Windows Credential Manager support is needed.

### Config / Credentials (3 tests)
- `TestListClientCredentials`
- `TestConfigExists`
- `TestReadClientCredentials_Errors`

Fail due to config path or environment variable assumptions that don't hold on Windows.

### Path Expansion (1 test)
- `TestExpandPath` (subtests: `tilde_only`, `tilde_with_subpath`)

`~` expansion uses `$HOME` which isn't set by default on Windows (should fall back to `$USERPROFILE`).

### Secrets (1 test)
- `TestResolveKeyringBackendInfo_Default` — expects `auto` backend, gets `file` on Windows.

### Tracking (1 test)
- `TestLoadSecrets_LegacyFallback` — legacy secret path resolution differs on Windows.

## Acceptance Criteria

- All 9 tests pass on `windows-latest` in CI
- Remove the `-skip` list from the `windows` job in `.github/workflows/ci.yml`
- No regressions on Linux or macOS

## References

- PR #388 (added Windows CI + skip list)
- Windows CI workflow: `.github/workflows/ci.yml`

steipete:
Fixed on current `main`.

I rechecked the exact Windows-test issue against `origin/main`:

- the named tests are present without Windows-specific skips (`TestAuthAddCmd_JSON`, `TestAuthListRemoveTokensListDelete_JSON`, `TestAuthServicesCmd_JSON`, `TestListClientCredentials`, `TestConfigExists`, `TestReadClientCredentials_Errors`, `TestExpandPath`, etc.)
- the only remaining Windows mentions are backend-selection/unit-table cases, not skips for these failures
- recent `main` CI includes the Windows job and is green, including format, tests, lint, and build

Closing this as fixed on main.

PR #527, opened 2026-04-25
fix(secrets): make file keyring filenames Windows-safe (#502)

### LINKED ISSUE #502: File keyring backend unusable on Windows: `:` in key name rejected by NTFS
## Summary

On Windows, `keyring_backend: file` is unusable because the backend writes files whose names contain colons (e.g. `token:default:user@gmail.com`). NTFS treats `:` as the Alternate Data Stream separator, so `os.OpenFile` fails with `ERROR_INVALID_NAME` before any token can be stored.

Any command that needs to read/write the keyring produces:

```
store token: open C:\Users\<user>\AppData\Roaming\gogcli\keyring\token:default:<email>: The filename, directory name, or volume label syntax is incorrect.
```

## Reproduction

1. Fresh Windows 11 install (happens on 10 as well).
2. `gog config set keyring_backend file`
3. `gog auth add --account you@gmail.com`

Result: OAuth flow completes, token is fetched from Google, then `gog` errors while storing it. `AppData\Roaming\gogcli\keyring\` remains empty.

Same error appears on any subsequent `gog auth list`, `gog auth tokens export`, etc.

## Environment

- OS: Windows 11 Home 26200 (also reproducible on Win 10)
- Shell: Git Bash (MSYS2) — also reproduces from `cmd.exe` and PowerShell 7
- gog: v0.12.0 (c18c58c)
- Filesystem: NTFS (default)

## Root cause

gog uses `github.com/99designs/keyring`. The `file` backend builds its on-disk path by concatenating the item's `Key` into the filename without sanitizing characters illegal on NTFS. gog stores refresh tokens under keys shaped like `token:<client>:<account-email>`, which contains two colons — both invalid in NTFS filenames. The backend has the same issue on macOS/Linux only if the key contains `/`, which it doesn't, so this is Windows-specific.

This is distinct from #395 (CI test skips) — this is a runtime crash on real user machines. It's the most likely reason a Windows user hits `auto` backend's fallback; except `auto` isn't the default everywhere, so anyone who ran `gog config set keyring_backend file` (including anyone who followed docs written with Linux/macOS in mind) is stuck with no tokens.

## Suggested fix

In the keying code that derives the on-disk filename, sanitize characters illegal on Windows before passing to the file backend. Minimal patch: replace `:`, `<`, `>`, `"`, `|`, `?`, `*` with `_` (or URL-encode them). Colon is the one that actually bites; the others are covered here for completeness.

Alternatively, hash the key (e.g. SHA-256 hex) for the filename and keep the original key inside the file body — this is what most keyring implementations do. 99designs/keyring may need patching upstream, or gog can wrap the backend with a key-sanitizing shim.

## Workaround

`gog config set keyring_backend auto` — this picks `wincred` (Windows Credential Manager / DPAPI) on Windows and works correctly. Only viable as a workaround because `wincred` is accepted as valid by `auto` resolution but rejected as a literal value in `keyring_backend` (the error message says "expected auto, keychain, or file" — `wincred` is not listed). So `auto` is the only path.

## Related

- 99designs/keyring — upstream library.
- #395 — Windows CI test failures (broader, different scope).
- #377 — file-backend token corruption (possibly related if the token got partially written; haven't verified).

solomonneas:
Closes #502.

## Problem

The `99designs/keyring` file backend writes one file per item, escaping only `/` in filenames via `percent.Encode`. NTFS rejects `:`, `<`, `>`, `"`, `|`, `?`, `*`, and `\`, so token keys like `token:default:user@gmail.com` cannot be persisted on Windows. Any auth-requiring command exits with:

```
store token: open C:\Users\<user>\AppData\Roaming\gogcli\keyring\token:default:<email>: The filename, directory name, or volume label syntax is incorrect.
```

## Fix

Wrap the keyring with a new `fileSafeKeyring` shim that:

- Percent-encodes Windows-illegal characters (`:`, `<`, `>`, `"`, `|`, `?`, `*`, `\`, plus `%` for self-inverse encoding) in item keys before they reach the on-disk filename builder.
- Prepends a `_v1_` sentinel to encoded keys so `Keys()` can unambiguously distinguish wrapper-written entries from legacy raw-form files. Avoids corrupting any pre-existing item whose key happened to contain a literal `%XX` triplet.
- Falls back to the raw key form on read so existing `~/.local/share/gogcli/keyring/` directories on macOS/Linux keep working without manual migration.
- Best-effort removes the legacy raw file once the new encoded form is persisted, so `Keys()` does not return duplicates.
- Translates the platform's "filename invalid" error (`ERROR_INVALID_NAME` on Windows, surfaced as a `*fs.PathError` wrapping `syscall.Errno(0x7B)`) to `keyring.ErrKeyNotFound`. Detection is build-tagged: `filename_safe_windows.go` for the real check, `filename_safe_other.go` returning `false` everywhere else. This is what lets the raw-form fallback safely run on every platform without leaking a confusing error from a path the underlying filesystem could never have stored.

The shim is applied unconditionally rather than only when the file backend is explicitly selected. That covers two cases at once:

1. **Auto-fallback to file backend** — `99designs/keyring`'s opener tries native backends in priority order and silently lands on the file backend when they fail (e.g. Windows + WinCred unavailable, Linux + secret-service+kwallet unavailable). The wrapper now catches those too.
2. **Native backends with legacy raw keys** — for example Windows WinCred, which stores opaque credential targets and has always accepted colon-bearing keys. The raw-key fallback finds those, so existing credentials survive the upgrade. (For non-file backends the encoder is a near no-op for keys without Windows-illegal characters.)

## Tests

`internal/secrets/filename_safe_test.go`, `_other_test.go`, `_windows_test.go` add 25 tests:

**Encoder unit tests**

- Round-trip across primary, legacy single-colon, scoped, all-illegal-chars, percent-bearing, plain, and empty inputs.
- Fast path: keys with no encodable character pass through unchanged.
- Sentinel reservation: keys legitimately starting with `_v1_` are still encoded, preserving the round trip.
- Backslash is handled.
- `decodeFilenameKey` does not corrupt legacy `%XX`-bearing names (no `_v1_` prefix → returned as-is).

**End-to-end fixture against a real `99designs/keyring` file backend in a temp dir**

- On-disk filenames contain no NTFS-illegal characters.
- `Keys()` returns the original (decoded) key set with duplicates collapsed.
- `Get` falls back to legacy raw-form files (`Skipf`-guarded for filesystems that reject the legacy filename, so the test stays portable).
- `Set` migrates a legacy raw file to the encoded form.
- `Remove` cleans both encoded and legacy forms.
- `Get`, `GetMetadata`, `Remove` of missing keys return `keyring.ErrKeyNotFound`.

**Cross-platform error-translation behaviour via a `fakeInnerKeyring` plus a swappable `isInvalidFilenameError`**

- `Get` / `GetMetadata` / `Remove` translate the simulated "invalid filename" error to `keyring.ErrKeyNotFound`.
- `Set` and `Remove` survive the inner backend rejecting the raw-form cleanup attempt with that error.
- `acceptingFakeKeyring` covers the WinCred upgrade scenario: a non-file backend that accepts colon-bearing keys still serves legacy items via the raw-key fallback.

**Platform-specific defaults**

- `filename_safe_windows_test.go` (build-tagged `windows`) verifies `defaultIsInvalidFilenameError` matches a `*fs.PathError` wrapping `syscall.Errno(0x7B)` and rejects unrelated errnos.
- `filename_safe_other_test.go` (build-tagged `!windows`) verifies the default returns `false`.

## Verification

- `make fmt` clean, `make lint` 0 issues, `make test` green across all packages including `cmd/gog` and `internal/cmd`.
- `-race` clean on the 25 new tests. (The pre-existing race in `TestOpenKeyringWithTimeout_Timeout` is upstream/main behaviour, untouched here.)
- Cross-compile: `GOOS=windows GOARCH=amd64` and `GOOS=darwin GOARCH=arm64` both build the package and tests cleanly.
- Coverage: 100% on `encodeFilenameKey` / `decodeFilenameKey` / `needsFilenameEncoding` / `isNotFound`; 79.5% package overall.

## Notes

Out of scope on purpose, called out in a comment on `safeFilenameMap` so a future contributor knows where to extend:

- Windows reserved device names (`CON`, `PRN`, `AUX`, `NUL`, `COM1-9`, `LPT1-9`).
- Trailing spaces / dots.
- ASCII control characters.

The current gogcli key namespace (`token:<client>:<email>` and `default_account[:<client>]`) cannot produce any of those forms.

Distinct from #395 (CI test skips on Windows). This is the runtime crash path for real user machines whether they explicitly set `keyring_backend: file` or auto-fall through to it.

### DIFF
--- a/internal/secrets/filename_safe.go
+++ b/internal/secrets/filename_safe.go
@@ -1,0 +1,276 @@
+package secrets
+
+import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"strings"
+
+	"github.com/99designs/keyring"
+)
+
+// isInvalidFilenameError reports whether err is the platform-specific signal
+// that the underlying filesystem rejected a path because the filename
+// contains characters it considers illegal (e.g. NTFS rejecting ":" via
+// ERROR_INVALID_NAME on Windows). Tests swap this to simulate other
+// platforms; the runtime default is wired up per-OS via
+// filename_safe_{windows,other}.go.
+var isInvalidFilenameError = defaultIsInvalidFilenameError
+
+// isNotFound treats keyring.ErrKeyNotFound, fs.ErrNotExist, and the
+// platform-specific "filename invalid" signal as "not found". The third case
+// matters on Windows: the 99designs/keyring file backend forwards
+// ERROR_INVALID_NAME from os.OpenFile/os.Remove rather than translating to
+// the keyring sentinel, so a raw-key fallback for a path with NTFS-illegal
+// characters would otherwise look like a hard error to callers branching on
+// ErrKeyNotFound.
+func isNotFound(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	if errors.Is(err, keyring.ErrKeyNotFound) || errors.Is(err, fs.ErrNotExist) {
+		return true
+	}
+
+	return isInvalidFilenameError(err)
+}
+
+// safeFilenameMap lists characters that are illegal in NTFS / Windows
+// filenames plus "%" itself, so percent-style encoding is self-inverse. The
+// 99designs/keyring file backend already escapes "/" via percent.Encode, so
+// it is omitted.
+//
+// Order matters: "%" -> "%25" must run first on encode and last on decode.
+//
+// Not handled here, on purpose: Windows reserved device names (CON, PRN,
+// AUX, NUL, COM1-9, LPT1-9), trailing spaces / dots, and ASCII control
+// characters. The gogcli key namespace ("token:<client>:<email>",
+// "default_account[:<client>]") cannot produce any of those forms; if a
+// future caller wires arbitrary user input through SetSecret, this list is
+// the place to extend.
+var safeFilenameMap = []struct {
+	raw     string
+	encoded string
+}{
+	{`%`, "%25"},
+	{`:`, "%3A"},
+	{`<`, "%3C"},
+	{`>`, "%3E"},
+	{`"`, "%22"},
+	{`|`, "%7C"},
+	{`?`, "%3F"},
+	{`*`, "%2A"},
+	{`\`, "%5C"},
+}
+
+// encodedKeyPrefix marks item keys that fileSafeKeyring rewrote so Keys()
+// can distinguish wrapper-written entries from legacy raw-form files. Older
+// gogcli versions wrote keys directly (e.g. "token:default:foo@bar.com"),
+// none of which start with this sentinel.
+//
+// Choosing a printable, filesystem-safe sentinel (vs. percent-encoded NULs
+// or high-bit Unicode) keeps the on-disk filenames human-debuggable on every
+// platform.
+const encodedKeyPrefix = "_v1_"
+
+// needsFilenameEncoding reports whether s contains any character that the
+// file keyring backend would persist as an illegal NTFS filename byte, or
+// whether it collides with the encodedKeyPrefix sentinel and therefore must
+// be encoded to round-trip cleanly through Keys() / decodeFilenameKey.
+func needsFilenameEncoding(s string) bool {
+	if strings.HasPrefix(s, encodedKeyPrefix) {
+		return true
+	}
+
+	for _, p := range safeFilenameMap {
+		if strings.Contains(s, p.raw) {
+			return true
+		}
+	}
+
+	return false
+}
+
+// encodeFilenameKey percent-encodes Windows-illegal characters in key and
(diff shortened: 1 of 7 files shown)


steipete:
Thanks @solomonneas. I landed a narrower maintainer version on main as 8135cc3.

What changed from the PR version:
- only the file backend gets portable encoded filenames; native keychain/WinCred-style backends keep their existing key names
- legacy raw file-backend entries are still read and removed
- added focused regression coverage plus README/spec/changelog notes

Closing this as landed.

PR #533, opened 2026-04-27
refactor(cli): deprecate --tab-id in favor of --tab

johnbenjaminlewis:
### Summary

The docs editing commands (`write`, `update`, `insert`, `delete`, `find-replace`) used `--tab-id` (ID only), while the read commands (`cat`, `structure`, `sed`) already used `--tab` (title or ID). 

- Renames `--tab-id` to `--tab` on all 5 editing commands, accepting both tab titles (case-insensitive) and tab IDs
- Keeps `--tab-id` as a hidden flag that still works but emits a deprecation warning to stderr
- Rejects passing both `--tab` and `--tab-id` simultaneously with a clear error
- Switches internal helpers (`docsTargetEndIndex`, `loadDocsTargetDocument`) from the ID-only `findTabByID` to the title-aware `findTab`, so title resolution actually works for mutations — not just reads
- Removes the now-unused `findTabByID`

### Test plan

- Existing tab editing tests updated to use `--tab` and still pass
- New `TestDocsWriteCmd_DeprecatedTabIDFlag` verifies `--tab-id` still works and emits the deprecation warning
- Help snapshot test updated to check for `--tab` in `docs update --help` output
- `make test` passes

### DIFF
--- a/internal/cmd/docs_edit.go
+++ b/internal/cmd/docs_edit.go
@@ -15,6 +15,23 @@ import (
 	"github.com/steipete/gogcli/internal/ui"
 )
 
+// resolveTabArg returns the effective tab value from --tab or the deprecated
+// --tab-id flag. It rejects supplying both and emits a deprecation warning
+// when --tab-id is used.
+func resolveTabArg(ctx context.Context, tab, tabID string) (string, error) {
+	tab = strings.TrimSpace(tab)
+	tabID = strings.TrimSpace(tabID)
+	if tab != "" && tabID != "" {
+		return "", usage("--tab and --tab-id are mutually exclusive (--tab-id is deprecated; use --tab)")
+	}
+	if tabID != "" {
+		u := ui.FromContext(ctx)
+		u.Err().Printf("Warning: --tab-id is deprecated; use --tab instead")
+		return tabID, nil
+	}
+	return tab, nil
+}
+
 type DocsWriteCmd struct {
 	DocID    string `arg:"" name:"docId" help:"Doc ID"`
 	Text     string `name:"text" help:"Text to write"`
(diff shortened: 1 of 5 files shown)


internal/cmd/docs_edit.go
```diff
@@ -15,6 +15,23 @@ import (
 	"github.com/steipete/gogcli/internal/ui"
 )
 
+// resolveTabArg returns the effective tab value from --tab or the deprecated
+// --tab-id flag. It rejects supplying both and emits a deprecation warning
+// when --tab-id is used.
+func resolveTabArg(ctx context.Context, tab, tabID string) (string, error) {
```
johnbenjaminlewis, line 21: This function + its [five] call sites are the logical changes of this PR. The rest of the changes are
- `s/c.TabID/c.Tab/`
- `s/--tab-id/--tab/`
- `s/findTabByID/findTab/`
- and removing the unused `findTabByID`

internal/cmd/docs_helpers.go
```diff
@@ -114,16 +114,6 @@ func docsDocumentEndIndex(doc *docs.Document) int64 {
 	return end
 }
 
-func findTabByID(tabs []*docs.Tab, tabID string) *docs.Tab {
```
johnbenjaminlewis, line 117: [func findTab](https://github.com/steipete/gogcli/blob/0337dcc572c1d55f090dbaedaca34012c32fb407/internal/cmd/docs_read.go#L366-L380) which duplicates this logic

steipete:
Thanks @johnbenjaminlewis. Landed on main as 50f506d, with a maintainer follow-up in 617293d.

What changed before landing:
- kept the hidden deprecated --tab-id alias and the --tab spelling
- fixed the mutation path so tab titles resolve to the canonical Docs tab ID before batchUpdate requests
- added/updated regression coverage for write/update/insert/delete/find-replace with --tab titles

Verified:
- go test ./internal/cmd focused docs/drive suite
- make ci
- live smoke on steipete@gmail.com: created a temp Doc, resolved "Tab 1" via --tab, wrote content, read it back, and trashed the doc

PR #534, opened 2026-04-27
refactor(googleapi): extract authenticatedTransport and add NewHTTPCl…

johnbenjaminlewis:
refactor(googleapi): extract authenticatedTransport and add NewHTTPClient

> Dependency for https://github.com/steipete/gogcli/pull/535  ⚠️ only merge this if you plan to merge #535 ⚠️ 

### Summary

- Extracts the OAuth transport setup from `optionsForAccountScopes` into a reusable `authenticatedTransport` helper, removing duplication between the existing google-api-go client path and callers that need a raw `*http.Client`
- Adds `NewHTTPClient(ctx, service, email)` for callers that need to set their own policies (e.g. custom `CheckRedirect`) on the HTTP client rather than going through `option.WithHTTPClient`
- Minor test cleanup: switches `newDocsServiceForTest` to use `t.Cleanup` instead of returning a manual close func

### Test plan

- `TestNewHTTPClient_NoRedirectPolicy` verifies the returned client has no redirect policy set (callers can add their own)
- `TestOptionsForAccountScopes_NoClientTimeout` continues to pass, confirming no regression in the existing path
- `make test` passes

### DIFF
--- a/internal/cmd/docs_testutil_test.go
+++ b/internal/cmd/docs_testutil_test.go
@@ -19,16 +19,16 @@ func newDocsServiceForTest(t *testing.T, h http.HandlerFunc) (*docs.Service, fun
 	t.Helper()
 
 	srv := httptest.NewServer(h)
+	t.Cleanup(srv.Close)
 	docSvc, err := docs.NewService(context.Background(),
 		option.WithoutAuthentication(),
 		option.WithHTTPClient(srv.Client()),
 		option.WithEndpoint(srv.URL+"/"),
 	)
 	if err != nil {
-		srv.Close()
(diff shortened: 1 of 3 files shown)


steipete:
Landed on `main` in `8f6791f` with the changelog credit added.

Kept the transport extraction/NewHTTPClient shape from the PR, applied it on top of the current auth subject-backfill changes, and preserved co-author credit.

Verification:
- `go test ./internal/googleapi ./internal/cmd -run 'TestNewHTTPClient|TestOptionsForAccountScopes_NoClientTimeout|TestDocs' -count=1`\n- `make lint`\n- `make test`\n- `make ci`\n\nThanks @johnbenjaminlewis.

Issue #544, opened 2026-05-01
go install v0.14.0 binary reports stale dev version


joshavant:
## Summary

Installing `gogcli` from a tagged module version with `go install` can produce a binary that reports a stale hardcoded development version instead of the installed tag.

Observed with `v0.14.0`:

```sh
go install github.com/steipete/gogcli/cmd/gog@v0.14.0
$(go env GOPATH)/bin/gog --version
```

Actual output:

```text
0.13.0-dev
```

Expected output:

```text
v0.14.0
```

## Comparison With Release Artifact

The prebuilt GitHub release artifact for `v0.14.0` reports the expected version:

```sh
curl -fsSL "https://github.com/steipete/gogcli/releases/download/v0.14.0/gogcli_0.14.0_linux_arm64.tar.gz" \
  | tar xz -C /usr/local/bin gog

gog --version
```

Output:

```text
v0.14.0 (469f4b4 2026-04-28T09:39:03Z)
```

## Likely Cause

The release artifacts are correct because GoReleaser injects version metadata through linker flags:

```yaml
-X github.com/steipete/gogcli/internal/cmd.version={{ .Tag }}
-X github.com/steipete/gogcli/internal/cmd.commit={{ .ShortCommit }}
-X github.com/steipete/gogcli/internal/cmd.date={{ .Date }}
```

But `go install github.com/steipete/gogcli/cmd/gog@v0.14.0` does not use those GoReleaser linker flags. It falls back to the hardcoded source value in `internal/cmd/version.go`, which currently looks like a stale release-specific dev version.

## Impact

For environments that install with `go install module@version`, the binary appears to be a stale development build even when it was built from the requested tagged module version. This can confuse diagnostics, packaging verification, and support reports.

## Related PR

I opened a fix here:

- https://github.com/steipete/gogcli/pull/545

That PR keeps GoReleaser-injected release metadata authoritative, but uses Go build metadata (`runtime/debug.ReadBuildInfo`) as the fallback for source/module builds. That means future installs like `go install github.com/steipete/gogcli/cmd/gog@<tag>` can report the installed module version instead of a stale hardcoded dev version.

steipete:
Fixed on current `main` in `6af52a4` (`fix(version): infer module version for source installs`).

`internal/cmd/version.go` now keeps GoReleaser-injected metadata authoritative, but falls back to `debug.ReadBuildInfo()` when the linked version is `dev` / `*-dev`, so `go install ...@tag` can report the module tag.

[...]

PR #545, opened 2026-05-01
fix: infer module version for source builds

joshavant:
## Summary

This makes version reporting use Go build metadata as a fallback when release metadata has not been injected by linker flags.

Release builds still work the same way: GoReleaser injects `version`, `commit`, and `date` with `-ldflags`, and those injected values remain authoritative. For source/module builds without those linker flags, `gog` now falls back to `debug.ReadBuildInfo()` and reports the main module version when Go provides one.

## Why

`go install github.com/steipete/gogcli/cmd/gog@v0.14.0` does not run through GoReleaser, so it does not receive the release workflow's `-ldflags`. Before this change, those builds fell back to the hardcoded source value (`0.13.0-dev`), which made a binary built from a newer tag look like an older development build.

That is misleading for diagnostics, support reports, and packaging verification.

## Why This Change

This keeps the release workflow and GoReleaser config unchanged while improving source-build behavior:

- GoReleaser artifacts keep their rich injected output, e.g. `v0.14.0 (commit date)`.
- `go install module@version` can report the module version from Go's build info.
- Plain local development builds still fall back to `dev` when no useful module version is available.
- Stale hardcoded release-like fallback values are removed.

## Verification

- `go test ./internal/cmd`
- `go test ./cmd/gog`
- `go build -o /tmp/gog-dev ./cmd/gog && /tmp/gog-dev --version` prints the local module/VCS-derived build version
- `go build -ldflags \"-X github.com/steipete/gogcli/internal/cmd.version=v9.9.9 -X github.com/steipete/gogcli/internal/cmd.commit=abc123 -X github.com/steipete/gogcli/internal/cmd.date=2026-05-01T00:00:00Z\" -o /tmp/gog-release ./cmd/gog && /tmp/gog-release --version` prints `v9.9.9 (abc123 2026-05-01T00:00:00Z)`

Related issue: https://github.com/steipete/gogcli/issues/544

### DIFF
--- a/internal/cmd/version.go
+++ b/internal/cmd/version.go
@@ -4,22 +4,39 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"runtime/debug"
 	"strings"
 
 	"github.com/steipete/gogcli/internal/outfmt"
 )
 
 var (
-	version = "0.13.0-dev"
-	commit  = ""
-	date    = ""
+	version       = "dev"
+	commit        = ""
+	date          = ""
+	readBuildInfo = debug.ReadBuildInfo
 )
 
-func VersionString() string {
+func resolvedVersion() string {
 	v := strings.TrimSpace(version)
-	if v == "" {
-		v = "dev"
+	if v != "" && v != "dev" && !strings.HasSuffix(v, "-dev") {
+		return v
 	}
+	info, ok := readBuildInfo()
+	if ok {
+		moduleVersion := strings.TrimSpace(info.Main.Version)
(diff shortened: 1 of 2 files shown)


steipete:
Thanks for the PR. This fix is already on `main` as `6af52a4` (`fix(version): infer module version for source installs`), including the module build-info fallback and regression coverage.

Verified current `origin/main` has the behavior with `go test ./internal/cmd -run TestVersionStringUsesModuleVersionFallback`; I also closed #544 with the same proof. Closing this PR as superseded by main, with attribution preserved in the changelog.

PR #560, opened 2026-05-05
fix(docs): unblock markdown --append on Thai/CJK/emoji content

### LINKED ISSUE #559: docs write --markdown --append hangs forever on Thai content (infinite loop in nextRune)
## Summary

`gog docs write <docId> --markdown --append --file <md>` hangs indefinitely (CPU-bound, no progress, no error) when the markdown content ends with a non-ASCII rune. Thai content reliably triggers it because Thai paragraphs/headings naturally end with a multi-byte rune. The same content works with `--markdown --replace` and with `--append` alone (no `--markdown`).

**Version:** v0.14.0 (build `469f4b4`, also reproducible on current `main` HEAD `56755e9`).

## Reproduction

```bash
DOC_ID=$(gog docs create "repro" -j --results-only | jq -r .id)

# A single Thai heading is enough to trigger the hang:
printf '## ส่วนคำถาม\n' > /tmp/test-thai.md

# Works (~3s) — replace path goes through Drive's server-side markdown converter
gog docs write "$DOC_ID" --markdown --replace --file /tmp/test-thai.md

# Hangs forever — append path goes through gogcli's local markdown parser
gog docs write "$DOC_ID" --markdown --append --file /tmp/test-thai.md

# Works (<2s) — plain text path bypasses the markdown parser entirely
gog docs write "$DOC_ID" --append --file /tmp/test-thai.md
```

Empirically the hang burns ~29s of user CPU per 30s wall clock — it is a **CPU-bound infinite loop**, not a stuck network call.

## Root cause

`internal/cmd/docs_markdown.go:459` — `nextRune` returns `(\"\", 0)` when its input is a string consisting of a single multi-byte rune:

```go
func nextRune(s string) (string, int) {
    for i, r := range s {
        if i > 0 {
            return s[:i], i
        }
        if len(s) == 1 {  // len() is bytes, not runes
            return s, 1
        }
        _ = r
    }
    return "", 0
}
```

For input `\"ก\"` (one Thai rune, 3 bytes):
1. Iteration 1: `i=0`, `r='ก'`. Neither branch fires (`i>0` false, `len(s)==1` false because len is 3 bytes).
2. The range loop ends.
3. Function returns `(\"\", 0)`.

The caller in `ParseInlineFormatting` then runs `currentByte += size` where `size == 0`, so `currentByte` never advances and the outer `for currentByte < len(text)` loop spins forever.

The condition is reached for any text ending in a multi-byte rune — Thai, Chinese, Japanese, Korean, emoji, etc. The reason `--markdown --replace` is unaffected is that it uses Drive's native markdown upload (`Media(strings.NewReader(cleaned), ContentType(mimeTextMarkdown))` in `docs_edit.go:218`), bypassing gogcli's local parser entirely. The `--append` path calls `insertDocsMarkdownAt` → `MarkdownToDocsRequests` → `ParseInlineFormatting` → `nextRune`.

## Affected paths

`ParseInlineFormatting` is also called from `docs_formatter.go` for headings, list items, blockquotes, and paragraphs, so any markdown-handling path that runs through the local parser is affected — not just `docs write --append --markdown` but also `docs find-replace --format markdown` and `docs sed`-family commands when the replacement content ends in a multi-byte rune.

## Suggested fix

Replace the hand-rolled rune iteration with `unicode/utf8.DecodeRuneInString`:

```go
func nextRune(s string) (string, int) {
    if s == \"\" {
        return \"\", 0
    }
    _, size := utf8.DecodeRuneInString(s)
    return s[:size], size
}
```

I'm opening a PR with this fix plus regression tests (unit tests for `nextRune`, timeout-guarded tests for `ParseInlineFormatting` and `MarkdownToDocsRequests` on Thai/emoji content).

## Environment

- gog v0.14.0 (build 469f4b4 2026-04-28)
- Linux 6.17, Go 1.26.2
- Auth backend: keyring + age secret store (irrelevant — bug is purely local CPU)

ninyawee:
## Summary

Fixes #559 — `gog docs write --markdown --append` hung indefinitely (CPU-bound infinite loop) on any markdown content ending in a non-ASCII rune. Thai content reliably triggered it because Thai paragraphs/headings naturally end with multi-byte runes; the same bug applies to CJK and emoji content.

The hang was in `nextRune` (`internal/cmd/docs_markdown.go`). For a string containing exactly one multi-byte rune, the previous range-based implementation:

```go
for i, r := range s {
    if i > 0 { return s[:i], i }   // never fires (only one iteration)
    if len(s) == 1 { return s, 1 } // false: len() is bytes, not runes
    _ = r
}
return "", 0   // ← falls through here
```

returned size 0, and the caller in `ParseInlineFormatting` advanced `currentByte += 0` forever inside its `for currentByte < len(text)` loop.

`--markdown --replace` is unaffected because that path uploads to Drive's server-side markdown converter (`Media(strings.NewReader(cleaned), ContentType(mimeTextMarkdown))` in `docs_edit.go`). Only `--append` and the find-replace/sed paths invoke the local parser, so this issue affects:

- `docs write --markdown --append`
- `docs find-replace --format markdown` (when replacement text ends in a multi-byte rune)
- `docs sed` family commands using markdown replacement content

## Fix

Replace the hand-rolled rune iteration with `utf8.DecodeRuneInString`. It handles every case correctly — empty input, ASCII, multi-byte runes, and invalid UTF-8 (`RuneError` with its byte size).

## Test plan

- [x] Added unit tests for `nextRune` covering the previously-broken paths (single Thai rune, single emoji, empty input, mixed strings).
- [x] Added timeout-guarded tests for `ParseInlineFormatting` on Thai headings, FAQ-style mixed text, bold/italic/code mixed with Thai, and trailing emoji. Each call is wrapped in a 2s timeout so any future infinite-loop regression fails fast in CI rather than stalling it.
- [x] Added an end-to-end test `TestMarkdownToDocsRequests_ThaiAppend` exercising the exact path used by `docs write --markdown --append` — `ParseMarkdown` → `MarkdownToDocsRequests` at a non-zero base index — on a multi-element Thai sample (heading + paragraph + bullet list + blockquote).
- [x] Verified the new tests **fail on `main` HEAD** (each times out and the goroutine stack confirms it's stuck inside `ParseInlineFormatting` at line 434) and **pass with the fix**.
- [x] `make lint` clean. Full `go test ./internal/cmd/...` passes apart from a pre-existing failure in `TestRequireAccount_Missing` that is unrelated to this change (also fails on stock `main`).
- [x] Verified end-to-end against the live Google Docs API:
  - The repro case (`--markdown --append` on a single-line Thai heading) used to time out at 30s; now completes in ~1.4s.
  - A 14 KB Thai markdown doc with 136 generated batch-update requests now appends in ~1.7s.

## Notes

- Issue #559 was filed by me from the same investigation; closing it on merge.
- I am running gogcli via `mise` (`github:steipete/gogcli@0.14.0`); the bug reproduces identically on current `main` (`56755e9`) which is what this branch targets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 0.16.0 - Unreleased
 
+### Fixed
+
+- Docs: avoid infinite loops when local Markdown parsing ends on Thai, CJK, emoji, or other multi-byte runes. (#560 / #559) — thanks @ninyawee.
+
 ## 0.15.0 - 2026-05-05
 
 ### Added
--- a/internal/cmd/docs_markdown.go
+++ b/internal/cmd/docs_markdown.go
@@ -5,6 +5,7 @@ import (
 	"regexp"
 	"strings"
 	"unicode/utf16"
+	"unicode/utf8"
 )
 
 const (
@@ -455,18 +456,16 @@ func ParseInlineFormatting(text string) ([]TextStyle, string) {
 	return styles, strippedText
 }
 
-// nextRune returns the first rune and its byte size from a string
+// nextRune returns the first rune and its byte size from a string.
+// For a string consisting of a single multi-byte rune (e.g. Thai or other
+// non-ASCII text), the previous range-based implementation returned size 0,
+// which caused callers like ParseInlineFormatting to spin in an infinite loop.
 func nextRune(s string) (string, int) {
-	for i, r := range s {
-		if i > 0 {
-			return s[:i], i
-		}
-		if len(s) == 1 {
-			return s, 1
-		}
-		_ = r
+	if s == "" {
+		return "", 0
 	}
-	return "", 0
+	_, size := utf8.DecodeRuneInString(s)
+	return s[:size], size
 }
 
 func parseHeading(line string) (int, string) {
--- a/internal/cmd/docs_markdown_thai_test.go
+++ b/internal/cmd/docs_markdown_thai_test.go
@@ -1,0 +1,110 @@
+package cmd
+
+import (
+	"testing"
+	"time"
+)
+
+// withTimeout runs fn in a goroutine and fails the test if it does not return
+// within d. Used to catch infinite loops without blocking the whole test run.
+func withTimeout(t *testing.T, d time.Duration, name string, fn func()) {
+	t.Helper()
+	done := make(chan struct{})
+	go func() {
+		defer close(done)
+		fn()
+	}()
+	select {
+	case <-done:
+	case <-time.After(d):
+		t.Fatalf("%s: timed out after %s (suspected infinite loop)", name, d)
+	}
+}
+
+// TestNextRune_SingleMultiByteRune is the direct unit-level regression test for
+// the bug that caused `gog docs write --markdown --append` to hang on any
+// content ending in a non-ASCII rune. The previous range-based nextRune
+// returned size=0 for a string that contained exactly one multi-byte rune
+// (e.g. a single Thai character), and ParseInlineFormatting then advanced
+// currentByte by 0, looping forever.
+func TestNextRune_SingleMultiByteRune(t *testing.T) {
+	cases := []struct {
+		name     string
+		in       string
+		wantStr  string
+		wantSize int
+	}{
+		{name: "empty", in: "", wantStr: "", wantSize: 0},
+		{name: "single ascii", in: "a", wantStr: "a", wantSize: 1},
+		{name: "two ascii", in: "ab", wantStr: "a", wantSize: 1},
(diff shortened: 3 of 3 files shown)


steipete:
Landed on main as d255d01.\n\nI added the missing changelog credit, then verified with:\n- `go test -vet=off ./internal/cmd -run 'TestNextRune_SingleMultiByteRune|TestParseInlineFormatting_Thai|TestMarkdownToDocsRequests_ThaiAppend'`\n- `make ci`\n- `git diff --check`\n\nThanks @ninyawee for the excellent repro and targeted fix.

PR #568, opened 2026-05-09
build(macos): auto self-sign local builds to stabilize keychain trust

xinbenlv:
## Problem

`make build` produces a `gog` binary that is only ad-hoc / linker-signed (Go's default).
When `gog` reads or writes a token in the user's macOS Keychain, `securityd` cannot
anchor a stable trust decision against an ad-hoc binary, so:

- The "Allow / Always Allow / Deny" prompt **re-appears on every invocation**, even
  after the user clicks "Always Allow".
- The dialog shows only the keychain item description (`gogcli`), with no path or
  signing identity, making it impossible for the user to understand that each
  build looks like "a different app" to macOS.
- For long-running gog processes that poll keychain (~every 10s on this user's
  setup), this becomes a continuous prompt loop.

The issue is reproducible on macOS 26.4 against tokens stored by the gogcli keyring
backend. `securityd` log confirms: `code requirement check failed (-67050), client
is not Apple-signed` plus `AMFI: '<gog path>' has no CMS blob`.

## Fix

Give every local build a **stable designated requirement** by signing it post-`go build`.

Two modes:

1. **Real signing identity (preferred):** if `GOG_CODESIGN_IDENTITY` (or
   `CODESIGN_IDENTITY`) is set — same env var the existing
   `scripts/codesign-macos.sh` (goreleaser hook) uses — the binary is signed with
   that identity. A developer with an Apple Developer ID gets unified signing for
   both local and release builds.

2. **Auto self-signed fallback:** if no identity is provided, a one-time bootstrap
   creates a self-signed identity in a dedicated keychain at
   `~/Library/Keychains/gog-codesign.keychain-db`. The cert is local-only (never
   used to distribute software) — its sole purpose is to give `gog` a stable
   `designated requirement` so the user's macOS Keychain ACL can persist
   "Always Allow" trust across rebuilds.

## Changes

- `scripts/codesign-macos-local.sh` — post-build signing hook (idempotent).
- `scripts/codesign-macos-bootstrap.sh` — one-time self-signed identity creation.
- `Makefile` — `build` target invokes `codesign-macos-local.sh` after `go build`.
  New explicit `codesign-local` and `codesign-bootstrap` targets.

## Behavior matrix

| Platform / Setting | Effect |
|---|---|
| Linux / Windows | No-op (script exits 0 on non-Darwin) |
| macOS, no env var | Bootstraps self-signed identity on first build, signs every build with it |
| macOS, `GOG_CODESIGN_IDENTITY` set | Signs with that identity (same as goreleaser hook) |
| macOS, `GOG_SKIP_LOCAL_CODESIGN=1` | No-op (opt-out for users who don't want it) |

## Untouched

- `.goreleaser.yaml` — release flow unchanged.
- `scripts/codesign-macos.sh` — release goreleaser hook unchanged.
- All existing CI workflows — unaffected (no Linux runner reaches this code path).
- Tests — none added; the script is tooling-only and shells out to `codesign` /
  `security` which can't be unit-tested in a useful way without elevated CI privs.

## Verification

Tested on macOS 26.4 against a real gogcli installation that was hitting the
prompt loop. After signing with the auto-bootstrapped self-signed identity,
`codesign --verify --strict` passes and the binary's designated requirement is:

```
identifier "com.openclaw.gogcli.gog" and certificate leaf = H"<sha1>"
```

This is stable across rebuilds (depends only on bundle id + cert leaf hash, not
CDHash), so the keychain ACL anchors "Always Allow" persistently.

## Carried patch

This change is also tracked as a carried patch in a downstream CPQ-managed fork.

Carried patch: patch-feat-pr568

---

Co-Authored-By: Oz <oz-agent@warp.dev>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ SHELL := /bin/bash
 .DEFAULT_GOAL := build
 
 .PHONY: build build-safe gog gogcli gog-help gogcli-help help fmt fmt-check lint test ci tools docs-commands docs-site docs-check
-.PHONY: worker-ci
+.PHONY: worker-ci codesign-local codesign-bootstrap
 
 BIN_DIR := $(CURDIR)/bin
 BIN := $(BIN_DIR)/gog
@@ -37,6 +37,13 @@ endif
 build:
 	@mkdir -p $(BIN_DIR)
 	@go build -ldflags "$(LDFLAGS)" -o $(BIN) $(CMD)
+	@./scripts/codesign-macos-local.sh $(BIN)
+
+codesign-local: build
+	@./scripts/codesign-macos-local.sh $(BIN)
+
+codesign-bootstrap:
+	@./scripts/codesign-macos-bootstrap.sh
 
 build-safe:
 	@./build-safe.sh $${PROFILE:-safety-profiles/agent-safe.yaml} -o $${OUTPUT:-$(BIN_DIR)/gog-safe}
--- a/scripts/codesign-macos-bootstrap.sh
+++ b/scripts/codesign-macos-bootstrap.sh
@@ -1,0 +1,114 @@
+#!/usr/bin/env bash
+# scripts/codesign-macos-bootstrap.sh
+#
+# One-time bootstrap of a self-signed code signing identity for local
+# `gog` builds. Idempotent: safe to re-run.
+#
+# Output state:
+#   ~/Library/Keychains/gog-codesign.keychain-db   — dedicated keychain
+#   ~/.gog-codesign-keychain-pass                  — keychain unlock password (mode 600)
+#   identity CN: gog-codesign-local-<hostname>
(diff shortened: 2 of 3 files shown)


steipete:
Thanks for the detailed diagnosis. I landed the release-side fix in `632c8bc`: the release workflow now fails closed if macOS signing secrets/identity are missing, `scripts/codesign-macos.sh` no longer silently skips Darwin signing, and `scripts/verify-release.sh` rejects ad-hoc Darwin artifacts or missing TeamIdentifier values. That addresses the Homebrew/shipped-binary Keychain trust problem from #569.

I am not going to land this PR as-is because wiring self-signed identity creation into plain `make build` would mutate a developer's Keychain/search list and write a local password file as a surprising side effect of a normal build. If we want local self-signing later, it should be an explicit opt-in command with very clear docs, not automatic on every build. Closing this PR in favor of the fail-closed release signing path.

PR #580, opened 2026-05-12
feat(calendar): list appointment schedules

### LINKED ISSUE #329: Feature request: Appointment Schedule support
Hey! Love this tool - it's become part of my daily workflow.

I've been using Google Calendar's Appointment Schedules (the bookable scheduling pages) and would love to manage them from `gog`. My main use case is updating my "Adjusted availability" each week - right now I have to go into the web UI every time.

I know this isn't fully supported in the official Calendar API yet - appointment schedules are stored as events with `eventType: "appointmentSchedule"` and you can list them with the `eventTypes` filter, but the schedule properties themselves aren't properly documented. So this might be a tricky one.

Even partial support would be great though:

1. `gog calendar appointment-schedules list` - see what schedules exist
2. `gog calendar appointment-schedules set-availability --date 2026-02-24 --from 10:30 --to 15:00` - override availability for a specific date
3. `gog calendar appointment-schedules clear-availability --date 2026-02-24` - remove an override

Happy to help test if you want to explore this. No worries if it's too undocumented to be worth it right now - just wanted to flag the interest!

alexuser:
## Summary

- Adds `gog calendar appointment-schedules list` as a read-only slice of appointment schedule support.
- Reuses the existing calendar events listing flow and filters Calendar API `Events.list` calls with `eventTypes=appointmentSchedule`.
- Covers the new appointment schedules list behavior with focused JSON/query-param and command-shape tests.

Fixes #329.

## Validation

- `go test ./internal/cmd -run 'Appointment|ListCalendarEvents'`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/internal/cmd/calendar.go
+++ b/internal/cmd/calendar.go
@@ -7,7 +7,7 @@ type CalendarCmd struct {
 	ACL             CalendarAclCmd             `cmd:"" name:"acl" aliases:"permissions,perms" help:"List calendar ACL"`
 	Alias           CalendarAliasCmd           `cmd:"" name:"alias" help:"Manage calendar aliases"`
 	Events          CalendarEventsCmd          `cmd:"" name:"events" aliases:"list,ls" help:"List events from a calendar or all calendars"`
-	Appointments    CalendarAppointmentsCmd    `cmd:"" name:"appointments" aliases:"appointment-schedules,appt" help:"Report Calendar appointment schedule API limitation"`
+	Appointments    CalendarAppointmentsCmd    `cmd:"" name:"appointments" aliases:"appointment-schedules,appt" help:"Manage appointment schedules"`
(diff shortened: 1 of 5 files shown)


steipete:
Thanks for the focused PR. I am going to close this one because the current Google Calendar API still does not expose appointment schedules through Events.list.

We verified the underlying request in #329 with the clawdbot test account: passing the appointment schedule event type returns a 400 invalidParameter response, and the allowed eventTypes are birthday, default, focusTime, fromGmail, outOfOffice, and workingLocation. The CLI therefore keeps `gog calendar appointments` as an explicit diagnostic instead of adding a list command that would fail against the live API.

If Google starts accepting appointment schedules in Events.list, this would be worth revisiting from a fresh branch. For now this is an upstream API limitation rather than a local implementation gap.

PR #634, opened 2026-05-22
feat(config): add XDG and GOG path overrides

### LINKED ISSUE #621: Feature: Honor XDG Base Directory Spec for credentials, vault, state, cache, and download paths
### Summary

gogcli's path resolution today follows `$XDG_CONFIG_HOME` only — every persisted file lands under `${XDG_CONFIG_HOME}/gogcli/` regardless of whether it's config, durable user data (OAuth tokens, keyring, service-account keys), mutable runtime state, regenerable cache, or bulk downloads. This conflates five distinct XDG kinds into one directory. **Proposal:** honor the [freedesktop.org XDG Base Directory Spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) per kind. When the relevant XDG env var is set, use it (e.g. `$XDG_DATA_HOME/gogcli/` for credentials); when unset, fall back to the **spec default** for that kind (`~/.local/share/gogcli/` for data, `~/.local/state/gogcli/` for state, `~/.cache/gogcli/` for cache, `~/.config/gogcli/` for config — never the current config-dir-for-everything layout). Today's `${XDG_CONFIG_HOME}/gogcli/` and `~/.gog/` paths are read-compat fallbacks for **existing files only**, per the read-legacy-write-new pattern in "Backwards compatibility" below — new installs go straight to the kind-correct defaults.

Mirrors the resolution sibling project [`openclaw/mcporter#155`](https://github.com/openclaw/mcporter/issues/155) shipped in mcporter [`v0.10.0`](https://github.com/openclaw/mcporter/tree/v0.10.0).

### What's actually written under `${XDG_CONFIG_HOME}/gogcli/` today

Source audit at gogcli tag [`v0.17.0`](https://github.com/openclaw/gogcli/tree/v0.17.0). Every path below derives from `config.Dir()` in [`internal/config/paths.go`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/config/paths.go), which itself resolves only via `XDG_CONFIG_HOME` (or `os.UserConfigDir()` as fallback). Categorical XDG kinds (per [the spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)) determine the right env var for each:

| File / subdir | Source site | XDG kind | Proposed XDG path |
|---|---|---|---|
| `credentials.json`, `credentials-<client>.json` (OAuth client metadata + refresh-token references) | [`ClientCredentialsPathFor()` in `paths.go`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/config/paths.go) | **data** | `$XDG_DATA_HOME/gogcli/credentials*.json` |
| `keyring/` (encrypted OAuth refresh tokens — file backend) | [`KeyringDir()` in `paths.go`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/config/paths.go) | **data** | `$XDG_DATA_HOME/gogcli/keyring/` |
| `keep-sa-<hash>.json`, `sa-<hash>.json` (Google service-account JSON keys) | [`KeepServiceAccountPath()`, `ServiceAccountPath()` in `paths.go`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/config/paths.go) | **data** | `$XDG_DATA_HOME/gogcli/{keep-sa,sa}-*.json` |
| `tracking.json` (mutable per-account tracking config — rotating keys, version counters) | [`ConfigPath()` in `internal/tracking/config.go`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/tracking/config.go) | **state** | `$XDG_STATE_HOME/gogcli/tracking.json` |
| `state/gmail-watch/` (per-account Gmail watch cursors) | [`GmailWatchDir()` in `paths.go`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/config/paths.go) | **state** | `$XDG_STATE_HOME/gogcli/gmail-watch/` |
| `drive-downloads/` (bulk Google Drive file downloads) | [`DriveDownloadsDir()` in `paths.go`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/config/paths.go) | **download** (user-data) | `$XDG_DOWNLOAD_DIR/gogcli/drive/` (with fallback) |
| `gmail-attachments/` (bulk Gmail attachment downloads) | [`GmailAttachmentsDir()` in `paths.go`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/config/paths.go) | **download** (user-data) | `$XDG_DOWNLOAD_DIR/gogcli/gmail-attachments/` (with fallback) |
| `<cache>/gog/gmail-*/...` (Gmail backup intermediate cache — 4 sites in `backup_gmail.go`) | [`backup_gmail.go`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/cmd/backup_gmail.go) (direct `os.UserCacheDir()` calls) | **cache** | `$XDG_CACHE_HOME/gogcli/gmail-backup/` |
| `~/.gog/backup.json` (backup config — separate root, bypasses XDG entirely) | [`defaultConfigPath()` in `internal/backup/config.go`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/backup/config.go) | **config** | `$XDG_CONFIG_HOME/gogcli/backup.json` |

**Note on `os.UserCacheDir()` in `backup_gmail.go`**: Go's stdlib already resolves `os.UserCacheDir()` via `$XDG_CACHE_HOME` on Linux, so the four cache sites accidentally honor XDG today — but they bypass any future tool-specific override mechanism (e.g. [`GOG_HOME`](https://github.com/openclaw/gogcli/issues) — see related feature request), and the path is `$XDG_CACHE_HOME/gog/` (note: `gog`, not `gogcli`!) because that's whatever the bare `os.UserCacheDir()` returns joined with whatever `cmd` code appends. Inconsistent with the `AppName = "gogcli"` constant used everywhere else.

**Note on `backup.json` at `~/.gog/`** ([`defaultConfigPath()` in `internal/backup/config.go`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/backup/config.go)): a second top-level dotfile root that bypasses XDG entirely and uses `gog` (not `gogcli`) as the directory name. Pre-existing inconsistency worth aligning in the same change.

### This codebase already has the right patterns — just not applied consistently

Two existing patterns make the proposed change consistent with current code, not foreign to it:

1. **gogcli already honors `XDG_CONFIG_HOME`** for the main config dir ([`config.Dir()`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/config/paths.go)) and the tracking config ([`tracking.ConfigPath()`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/tracking/config.go)). The proposal is to extend the *kind-aware* pattern to the other XDG vars, not to bolt on XDG from scratch.

2. **`AppName = "gogcli"` is already centralized** ([`paths.go`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/config/paths.go)) and reused across every path-resolution function in `internal/config/`. Adding `DataDir()`, `StateDir()`, `CacheDir()`, `DownloadsDir()` siblings to the existing `Dir()` reuses the same naming and the same `AppName` constant.

### Why XDG over a single bespoke `GOG_HOME`

A separate (related) feature request is planned for `GOG_HOME` and per-kind `GOG_*_DIR` env vars as tool-specific path overrides — a `MCPORTER_CONFIG`-style escape hatch above the XDG layer. That request and this one are **complementary, not alternatives**:

- **This proposal (XDG kinds)** fixes the *semantic* problem: gogcli today stores OAuth credentials, mutable state, and bulk downloads in the *config* directory, which is wrong by spec regardless of any override mechanism.
- **`GOG_HOME` (the other proposal)** adds a tool-specific override *above* the XDG layer for environments where XDG resolution is unavailable or undesirable (sandboxes that strip XDG vars, CI runners, etc.).

Both are valuable because they fix different problems. This proposal should land first because it's the semantic correctness fix; `GOG_HOME` is the operator-ergonomics addition on top.

The spec rationale itself:

1. **Standardized**. Every Linux operator already knows what the XDG vars mean — no new tool-specific env var to discover and document.

2. **Categorical separation matters in deployments.** Per the [spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html):

   | Spec var | Default | What goes there | gogcli use |
   |---|---|---|---|
   | `XDG_CONFIG_HOME` | `~/.config` | User config files | OAuth client metadata, tracking config (keyring backend selection if it ever becomes file-overridable) |
   | `XDG_DATA_HOME` | `~/.local/share` | Durable user data | `credentials*.json`, `keyring/`, service-account JSON keys |
   | `XDG_STATE_HOME` | `~/.local/state` | Logs, history, transient state | `tracking.json`, `gmail-watch/` cursors |
   | `XDG_CACHE_HOME` | `~/.cache` | Regenerable | Gmail backup intermediate cache |

   Operators routinely place cache on tmpfs/local-SSD (regenerable, fast loss OK), data on backed-up volumes (irreplaceable), state on a per-instance volume (instance-scoped), and config under version control (declarative). XDG keeps each on its appropriate medium; collapsing everything under `XDG_CONFIG_HOME` defeats this entirely.

   Concrete pain point from the calling deployment: gogcli's encrypted OAuth refresh-token keyring is **data** (irreplaceable without a fresh OAuth flow) but lives under what is conventionally treated as **config** (often version-controlled, never backed up). When the operator configures their container persistence layer to back up `$XDG_DATA_HOME` but not `$XDG_CONFIG_HOME` (a reasonable choice — config is in source control, data is not), gogcli's keyring is silently lost on instance replacement.

3. **Composes with multi-tool environments.** gogcli isn't the only Linux CLI in many container images. If the operator already sets XDG vars at the container/image/systemd-unit level for other tools, gogcli joining the convention means *zero* additional config — the same env vars cover everything.

4. **Better defaults on Linux.** Today's `${XDG_CONFIG_HOME}/gogcli/` (often `~/.config/gogcli/`) holds credentials and downloads alongside config — non-standard for Linux. Moving credentials to `~/.local/share/gogcli/` and downloads to `~/Downloads/gogcli/` aligns with where Linux users expect to find each kind.

5. **Spec-conformant tools compose with system-level integrations.** Backup tools, dotfile managers, restic policies, and rsync exclude lists all key off XDG paths. A tool that lumps everything under one XDG kind defeats all of them.

### Backwards compatibility (must-have)

The standard pattern (adopted by [`gh`](https://cli.github.com/manual/gh_help_environment), [`pnpm`](https://pnpm.io/npmrc#env-variables), [`yt-dlp`](https://github.com/yt-dlp/yt-dlp), and many other CLIs — including the sibling [`mcporter` per `#155`](https://github.com/openclaw/mcporter/issues/155)):

1. **Honor XDG when explicitly set.** If `$XDG_DATA_HOME` is non-empty, use `$XDG_DATA_HOME/gogcli/` for data files; fall back otherwise.
2. **Auto-detect existing legacy paths.** On first run after upgrade, if `credentials.json` (or keyring, or tracking.json, etc.) exists at the legacy `${XDG_CONFIG_HOME}/gogcli/` location but not at the new kind-specific path, *read* from the legacy path. On the next *write*, write to the new path and emit a one-line stderr notice telling the operator the file moved. No silent breakage on upgrade.
3. **Consolidate `~/.gog/backup.json`.** The non-XDG `~/.gog/` root in `internal/backup/config.go` should migrate to `$XDG_CONFIG_HOME/gogcli/backup.json` with the same read-legacy-write-new fallback.

Step (1) alone is enough to fix the spec violation; (2) and (3) close the migration cleanly without forcing operators to manually move files.

### Minimal implementation sketch

Extend `internal/config/paths.go` with kind-aware resolvers paralleling the existing `Dir()`:

```go
package config

import (
    "fmt"
    "os"
    "path/filepath"
    "strings"
)

const AppName = "gogcli"

type XdgKind int

const (
    KindConfig XdgKind = iota
    KindData
    KindState
    KindCache
)

// xdgEnvVar returns the spec env var for a given kind.
func xdgEnvVar(kind XdgKind) string {
    switch kind {
    case KindConfig: return "XDG_CONFIG_HOME"
    case KindData:   return "XDG_DATA_HOME"
    case KindState:  return "XDG_STATE_HOME"
    case KindCache:  return "XDG_CACHE_HOME"
    }
    return ""
}

// xdgDefault returns the spec default path for a given kind under $HOME.
func xdgDefault(kind XdgKind) string {
    switch kind {
    case KindConfig: return ".config"
    case KindData:   return filepath.Join(".local", "share")
    case KindState:  return filepath.Join(".local", "state")
    case KindCache:  return ".cache"
    }
    return ""
}

// kindDir returns the gogcli directory for a given XDG kind.
//
// Resolution order:
//   1. $XDG_<KIND>_HOME if set (per XDG Base Directory Spec)
//   2. $HOME/<spec-default>/gogcli (also per XDG spec)
//
// Backwards-compat fallback for callers reading existing files:
// pre-XDG-kind-aware files live under Dir() (config kind) regardless of
// their semantic kind. Callers should attempt the kind-specific path first,
// then fall back to Dir() for read-only legacy compatibility — see the
// migration helpers below.
func kindDir(kind XdgKind) (string, error) {
    if xdg := strings.TrimSpace(os.Getenv(xdgEnvVar(kind))); xdg != "" {
        return filepath.Join(xdg, AppName), nil
    }
    home, err := os.UserHomeDir()
    if err != nil {
        return "", fmt.Errorf("resolve user home dir: %w", err)
    }
    return filepath.Join(home, xdgDefault(kind), AppName), nil
}

// Dir is preserved for backwards-compat and points at the config kind.
func Dir() (string, error) { return kindDir(KindConfig) }

// DataDir returns the gogcli data root ($XDG_DATA_HOME/gogcli).
func DataDir() (string, error) { return kindDir(KindData) }

// StateDir returns the gogcli state root ($XDG_STATE_HOME/gogcli).
func StateDir() (string, error) { return kindDir(KindState) }

// CacheDir returns the gogcli cache root ($XDG_CACHE_HOME/gogcli).
func CacheDir() (string, error) { return kindDir(KindCache) }
```

Then update each call site to its correct kind:

```go
// internal/config/paths.go: KeyringDir, ClientCredentialsPathFor,
// {Keep,}ServiceAccountPath, ListServiceAccountEmails — switch from Dir() to DataDir()
func KeyringDir() (string, error)             { return joinUnder(DataDir, "keyring") }
func ClientCredentialsPathFor(...) (string)   { return joinUnder(DataDir, "credentials-*.json") }
func KeepServiceAccountPath(...) (string)     { return joinUnder(DataDir, "keep-sa-*.json") }
func ServiceAccountPath(...) (string)         { return joinUnder(DataDir, "sa-*.json") }
// (with backwards-compat read fallback through legacyDir() → Dir())

// internal/config/paths.go: GmailWatchDir — switch from Dir() to StateDir()
func GmailWatchDir() (string, error) {
    sd, err := StateDir()
    if err != nil { return "", err }
    return filepath.Join(sd, "gmail-watch"), nil
}

// internal/tracking/config.go: ConfigPath — switch from config.Dir() to config.StateDir()
func ConfigPath() (string, error) {
    sd, err := config.StateDir()
    if err != nil { return "", err }
    return filepath.Join(sd, "tracking.json"), nil
}
// legacyConfigPath() already exists for backwards-compat reads — keep + extend.

// internal/cmd/backup_gmail.go: 4 os.UserCacheDir() sites — switch to config.CacheDir()
// (no more bare os.UserCacheDir(); routes through the kind-aware resolver
// so any future override mechanism applies uniformly)
dir, err := config.CacheDir()  // was: os.UserCacheDir()

// internal/config/paths.go: DriveDownloadsDir, GmailAttachmentsDir
// — switch to a new downloadsDir() resolver that honors $XDG_DOWNLOAD_DIR
// (parsed from $XDG_CONFIG_HOME/user-dirs.dirs on Linux, fall back to $HOME/Downloads)
func DriveDownloadsDir() (string, error)    { return joinUnder(downloadsDir, "drive") }
func GmailAttachmentsDir() (string, error)  { return joinUnder(downloadsDir, "gmail-attachments") }

// internal/backup/config.go: defaultConfigPath — switch from ~/.gog/backup.json to
// config.Dir() + "/backup.json", with read-legacy-write-new fallback.
func defaultConfigPath() (string, error) {
    cd, err := config.Dir()
    if err != nil { return "", err }
    return filepath.Join(cd, "backup.json"), nil
}
```

Each updated reader should follow the read-legacy-write-new compat pattern:

```go
// Read primary, fall back to legacy, then migrate on write
data, err := os.ReadFile(newPath)
if errors.Is(err, fs.ErrNotExist) {
    data, err = os.ReadFile(legacyPath)
    if err == nil {
        // notify operator + return data; next write goes to newPath
        fmt.Fprintf(os.Stderr, "note: read %s from legacy path; will migrate on next write\n", base)
    }
}
```

### Use case

Concrete deployment context: gogcli running inside a containerized agent host (Cloudflare Sandbox in this case, but the shape generalizes to any container/sandbox runtime). Per-tenant state — including OAuth refresh tokens for Google services the agent has authenticated against — must persist across container replacement via an external sync mechanism (rclone to object storage in this case). The container entrypoint exports XDG vars pointing into the persisted root:

- `XDG_CONFIG_HOME=/persist`
- `XDG_DATA_HOME=/persist`
- `XDG_STATE_HOME=/persist`
- `XDG_CACHE_HOME=/persist/cache` (or tmpfs, depending on policy)

With today's gogcli (only `XDG_CONFIG_HOME` is read), the operator must collapse the four kinds into one. With this change, the operator can route each kind to its appropriate storage tier — backed-up volumes for data, tmpfs for cache, etc. — without the XDG-conflation workaround.

This is the same shape of use case behind mcporter#155, and the resolution there (mcporter `v0.10.0`) closed the gap on mcporter's side. Closing it on gogcli's side completes the spec-conformance pair across the two sibling tools.

### Related references

**Sibling-project precedent:**
- [`openclaw/mcporter#155`](https://github.com/openclaw/mcporter/issues/155) ("Honor XDG Base Directory Spec for config, vault, cache, and daemon paths") — closed as completed, shipped in mcporter [`v0.10.0`](https://github.com/openclaw/mcporter/tree/v0.10.0). Same spec, same framing, same conclusion: split state by XDG kind, with read-legacy-write-new compatibility for existing users. This issue applies the same pattern to gogcli.

**Related (complementary, not alternative):**
- [gogcli#622](https://github.com/openclaw/gogcli/issues/622) — operator-side `GOG_HOME` and per-kind `GOG_*_DIR` env-var overrides above the XDG layer. Fixes the operator-ergonomics gap (XDG-hostile sandboxes); this issue fixes the semantic-kind-mapping gap. Both should land; sequencing this one first preserves backwards-compat the cleanest because the per-kind resolvers exist before the override layer can route them.

**Upstream OpenClaw issue this complements:**
- [`openclaw/openclaw#84854`](https://github.com/openclaw/openclaw/issues/84854) — OpenClaw's host-env sanitizer strips `XDG_CONFIG_HOME` and `XDG_CONFIG_DIRS` from subprocess env, breaking inheritance for XDG-aware tools. Not blocking this proposal (the gogcli-side fix here is correct regardless), but the two together close the inheritance + producer side of the XDG-conformance story.

**XDG Base Directory Specification:**
- [freedesktop.org spec, latest revision](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) — defines all seven XDG Base Directory env vars, their inheritance semantics, and the per-kind separation this proposal implements.
- [`user-dirs.dirs` and `XDG_DOWNLOAD_DIR`](https://www.freedesktop.org/wiki/Software/xdg-user-dirs/) — the spec-conformant way to discover the user's Downloads dir, relevant for the bulk-download path resolution above.

**gogcli source code referenced (all at tag [`v0.17.0`](https://github.com/openclaw/gogcli/tree/v0.17.0)):**
- [`internal/config/paths.go`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/config/paths.go) — central path-resolution: `Dir()`, `KeyringDir()`, `ClientCredentialsPathFor()`, `DriveDownloadsDir()`, `GmailAttachmentsDir()`, `GmailWatchDir()`, `KeepServiceAccountPath()`, `ServiceAccountPath()`, `ListServiceAccountEmails()`.
- [`internal/tracking/config.go`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/tracking/config.go) — `ConfigPath()`, `legacyConfigPath()` (legacy-fallback pattern already in use here for read-only — proposal extends it across the rest of the file inventory).
- [`internal/cmd/backup_gmail.go`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/cmd/backup_gmail.go) — 4 direct `os.UserCacheDir()` call sites that bypass the central resolver.
- [`internal/backup/config.go`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/backup/config.go) — `defaultConfigPath()` uses `~/.gog/backup.json` (non-XDG `~/.gog/` root) — pre-existing inconsistency to consolidate in this change.

### LINKED ISSUE #622: Feature: Add GOG_HOME (and per-kind GOG_*_DIR) env vars and matching CLI flags to control state paths independently of XDG
### Summary

Even with full XDG Base Directory Spec support ([gogcli#621](https://github.com/openclaw/gogcli/issues/621)), gogcli's path resolution remains coupled to operator control over `$XDG_*_HOME` vars. In environments where the parent process can't or won't set XDG vars portably — agent sandboxes, container runtimes, CI runners — gogcli ends up with the wrong paths.

Add a gogcli-specific override layer above XDG: a primary `GOG_HOME` umbrella env var (matching `CARGO_HOME`, `RUSTUP_HOME`, `GOPATH`) and per-XDG-kind overrides (`GOG_CONFIG_DIR`, `GOG_DATA_DIR`, `GOG_STATE_DIR`, `GOG_CACHE_DIR`), plus a `--home <path>` global CLI flag.

### Why `GOG_*` and not `GOGCLI_*`

Empirical verification at tag [`v0.17.0`](https://github.com/openclaw/gogcli/tree/v0.17.0): every env var gogcli currently reads uses the `GOG_*` prefix — `GOG_KEYRING_BACKEND`, `GOG_KEYRING_PASSWORD`, `GOG_KEYRING_SERVICE_NAME`, `GOG_ACCOUNT`, `GOG_HELP`, `GOG_COLOR`, `GOG_TIMEZONE`, `GOG_AUTH_MODE`, `GOG_*_API_KEY`, `GOG_*_BASE_URL`. Zero `GOGCLI_*` env vars exist in source. The `GOG_*` namespace is gogcli's established convention; this proposal matches it.

(This differs from sibling [`mcporter`](https://github.com/openclaw/mcporter)'s `MCPORTER_*` convention, but matching gogcli's own internal convention is the right principle — pushing `GOGCLI_*` would force gogcli to break its own namespace.)

### Why a `*_HOME` umbrella, not just a `*_CONFIG_DIR`

`config.Dir()` in [`internal/config/paths.go`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/config/paths.go) is the single function controlling *every* persisted-state location in gogcli today — config, keyring, credentials, drive downloads, gmail attachments, service-account keys, gmail-watch state, tracking config. After XDG conformance lands, that single function splits into per-kind resolvers (`DataDir`, `StateDir`, `CacheDir`, etc.). The override layer needs to match that shape: one umbrella (`GOG_HOME` — analogous to `CARGO_HOME`, `RUSTUP_HOME`) plus per-kind overrides for operators who want to split lifetimes (durable data on a backed-up volume, cache on tmpfs).

A single `GOG_CONFIG_DIR` would only cover one of the four+ resolvers, defeating the purpose. Matches the more established CLI convention (cargo, rustup, gh).

### Current behavior (after XDG-conformance fix)

Once [gogcli#621](https://github.com/openclaw/gogcli/issues/621) lands, the path-resolution chain for each kind is:

1. `$XDG_<KIND>_HOME` env var, if set
2. `$HOME/<spec-default>/gogcli` (XDG fallback)

There's no gogcli-specific override above XDG. If the parent process strips or rewrites the XDG vars, the operator has no recourse.

### Why this matters

gogcli's deployability remains coupled to upstream/sandbox XDG handling. Concrete pain point we hit:

[OpenClaw](https://github.com/openclaw/openclaw) strips `XDG_CONFIG_HOME` from the inherited environment of subprocesses it spawns, due to a policy oversight where `XDG_CONFIG_HOME` is listed in `blockedOverrideOnlyKeys` without a matching entry in `allowedInheritedOverrideOnlyKeys` (see [`host-env-security-policy.json`](https://github.com/openclaw/openclaw/blob/v2026.5.19/src/infra/host-env-security-policy.json)). Filed upstream as [openclaw#84854](https://github.com/openclaw/openclaw/issues/84854).

Even once that bug is fixed, other sandboxes / container runtimes / CI environments may also restrict or rewrite XDG vars. A tool-specific override is the standard escape hatch — it's portable, it doesn't collide with broader sandbox policies that target XDG vars, and it makes operator intent explicit ("I want gogcli's state *here*") rather than implicit via a shared env var.

[mcporter#184](https://github.com/openclaw/mcporter/issues/184) illustrates the inverse failure mode: embedders sometimes set `XDG_CONFIG_HOME` for an *unrelated* downstream tool, accidentally redirecting other XDG-aware tools too. A tool-specific override insulates gogcli from this kind of cross-tool XDG collision.

Precedent: [`mcporter`](https://github.com/openclaw/mcporter) (sibling project in `openclaw/`) exposes `MCPORTER_CONFIG` for exactly this reason — to override the XDG-derived config path when XDG resolution is unavailable or undesirable (see [`path-discovery.ts`](https://github.com/openclaw/mcporter/blob/v0.11.1/src/config/path-discovery.ts) at `v0.11.1`).

### Proposed solution

**Scope**: this proposal covers the four XDG Base Directory kinds (config, data, state, cache). The fifth kind from [gogcli#621](https://github.com/openclaw/gogcli/issues/621) — *downloads* — falls under a separate freedesktop spec ([xdg-user-dirs](https://www.freedesktop.org/wiki/Software/xdg-user-dirs/)) and is intentionally **not** part of `GOG_HOME`'s umbrella: downloads are user-visible content with their own routing semantics (operators expect `~/Downloads/`-style paths, not state-dir paths). Operators wanting downloads under `GOG_HOME` should set `XDG_DOWNLOAD_DIR` explicitly. A symmetric `GOG_DOWNLOAD_DIR` per-kind override could be added if the maintainer prefers — flagging as an open question.

Per-XDG-kind precedence chain (first match wins, most-specific to least-specific):

1. **`GOG_<KIND>_DIR`** env var (where `<KIND>` is `CONFIG` / `DATA` / `STATE` / `CACHE`) — direct directory path, used as-is with no `gogcli` subdir append. Highest precedence; lets operators split lifetimes (e.g. data on a backed-up volume, cache on tmpfs).
2. **`--home <path>`** CLI flag — per-invocation umbrella. Resolves to `<path>/<kind>/` for each kind (flat XDG-kind sub-layout). Wins over `GOG_HOME` per the cobra-flag-overrides-env convention. Global, persistent across all subcommands.
3. **`GOG_HOME`** env var — durable umbrella with the same flat XDG-kind sub-layout (`$GOG_HOME/{config,data,state,cache}/`). The intended shape for container entrypoints / systemd units / shell profiles.
4. **`$XDG_<KIND>_HOME/gogcli`** — spec-conformant XDG path, from [gogcli#621](https://github.com/openclaw/gogcli/issues/621).
5. **`$HOME/<XDG-default>/gogcli`** — XDG default fallback (e.g. `~/.config/gogcli`, `~/.local/share/gogcli`, `~/.local/state/gogcli`, `~/.cache/gogcli`).

Rank 1 (`GOG_<KIND>_DIR`) wins over both umbrella forms because per-kind overrides represent the most specific operator intent.

**Why flat XDG-kind sub-layout under `GOG_HOME` (not `$HOME`-style dotted sub-layout)**: operators setting `GOG_HOME=/persist` expect to find gogcli's state directly under `/persist/`, not under `/persist/.local/share/gogcli/`. Cargo/Rustup precedent backs this — `CARGO_HOME` contains `registry/`, `git/`, `bin/` directly (not `.local/share/cargo/`). The `.local/share/`-style nesting is meaningful only in the XDG default case where the resolver synthesizes a path under `$HOME`; once the operator has explicitly named a directory via `GOG_HOME`, that directory *is* the gogcli root, with kinds as immediate children.

Sketch (extending the resolver from [gogcli#621](https://github.com/openclaw/gogcli/issues/621)):

```go
package config

// kindName is introduced by this proposal (not part of gogcli#621).
// gogcli#621 defines xdgDefault() for the XDG-default fallback case;
// kindName() is the flat-layout equivalent for use under $GOG_HOME.
//
// Different from xdgDefault() (gogcli#621): xdgDefault() returns the XDG-spec
// default path under $HOME ("config", ".local/share", ".local/state", ".cache"),
// suitable for the fallback case. kindName() returns the short kind name
// ("config", "data", "state", "cache") for use under operator-named $GOG_HOME,
// matching the CARGO_HOME / RUSTUP_HOME convention of flat sub-layouts.
func kindName(kind XdgKind) string {
    switch kind {
    case KindConfig: return "config"
    case KindData:   return "data"
    case KindState:  return "state"
    case KindCache:  return "cache"
    }
    return ""
}

// kindDir returns the gogcli directory for a given XDG kind.
//
// Resolution order (most specific to least):
//   1. GOG_<KIND>_DIR env var — operator-supplied absolute path; used as-is
//   2. $GOG_HOME/<kindName> — umbrella override, flat XDG-kind sub-layout
//   3. $XDG_<KIND>_HOME/gogcli — spec-conformant XDG path
//   4. $HOME/<XDG-default>/gogcli — XDG default fallback
//
// The --home CLI flag is resolved at the cobra layer by overwriting GOG_HOME
// in the process env before kindDir runs, so it doesn't need its own branch
// here — it appears as rank 2 to the operator but is implemented as a flag
// → env coercion above this resolver.
func kindDir(kind XdgKind) (string, error) {
    // Per-kind direct override
    if perKind := strings.TrimSpace(os.Getenv(gogKindEnvVar(kind))); perKind != "" {
        return perKind, nil  // operator-supplied absolute path; no AppName join
    }
    // Umbrella GOG_HOME — flat XDG-kind sub-layout
    if home := strings.TrimSpace(os.Getenv("GOG_HOME")); home != "" {
        return filepath.Join(home, kindName(kind)), nil
        // e.g. GOG_HOME=/persist → /persist/data for the data kind
        // (or operator overrides per-kind via GOG_DATA_DIR for full control)
    }
    // XDG layer (from gogcli#621)
    if xdg := strings.TrimSpace(os.Getenv(xdgEnvVar(kind))); xdg != "" {
        return filepath.Join(xdg, AppName), nil
    }
    // XDG default
    home, err := os.UserHomeDir()
    if err != nil {
        return "", fmt.Errorf("resolve user home dir: %w", err)
    }
    return filepath.Join(home, xdgDefault(kind), AppName), nil
}

func gogKindEnvVar(kind XdgKind) string {
    switch kind {
    case KindConfig: return "GOG_CONFIG_DIR"
    case KindData:   return "GOG_DATA_DIR"
    case KindState:  return "GOG_STATE_DIR"
    case KindCache:  return "GOG_CACHE_DIR"
    }
    return ""
}
```

CLI: `--home <path>` as a persistent global flag on the root cobra command, equivalent to setting `GOG_HOME` for the duration of the invocation.

**Interaction with gogcli#621's public API**: the public wrappers added by gogcli#621 (`Dir()`, `DataDir()`, `StateDir()`, `CacheDir()`) are preserved — they delegate to `kindDir(kind)` exactly as before, gaining the new override layer transparently. No call-site changes beyond what gogcli#621 already specifies.

### Open design questions for the maintainer

1. **Per-kind override semantics — strip the `gogcli` suffix?**: `GOG_DATA_DIR=/foo/bar` should resolve to `/foo/bar` (no `gogcli` join) — same as `MCPORTER_CONFIG` doesn't auto-join. Convention from other tools (`CARGO_HOME`, `RUSTUP_HOME`, `GOPATH`) is "no auto-join, operator supplies the literal directory." The sketch above implements this; flagging in case maintainer prefers an auto-join variant.
2. **CLI flag scope**: `--home` as a persistent global flag on the root command, matching `--config` / `--profile` conventions in most cobra CLIs. Per-subcommand override flags (`--config-dir`, etc.) are likely overkill — env vars cover the use case.
3. **Validation**: should the per-kind override reject non-absolute paths (defensive) or accept them (let CWD context apply)? mcporter rejects non-absolute via `path.isAbsolute` check. **Recommended**: reject non-absolute, fail-fast with a clear error.

(Note: the prior draft included an open question about `GOG_HOME` flat vs. XDG-default sub-layout. The sketch above commits to flat XDG-kind sub-layout per the rationale paragraph in "Proposed solution"; flagging here in case the maintainer prefers a different shape.)

### Alternatives considered

- **Symlink workaround** at the deployment layer (e.g. `ln -sfn /persist/gogcli ~/.config/gogcli`) — works as break-glass but couples deployment to an internal gogcli path convention that could change. Also doesn't compose with the XDG-kind split.
- **Wrapper script** exporting XDG vars before invoking `gogcli` — works but only when the wrapper has control over how gogcli is spawned. Doesn't help when gogcli is spawned by another tool (an agent, a skill runner) that already sanitized the env.
- **Fix only the parent sandbox** (OpenClaw, in our case) — necessary, and being pursued separately ([openclaw#84854](https://github.com/openclaw/openclaw/issues/84854)). But this leaves gogcli brittle in every *other* environment that does similar XDG sanitization.
- **Single `GOG_CONFIG_DIR` only** (no umbrella, no per-kind variants) — covers only one resolver; defeats the kind-aware separation that XDG conformance just introduced. This was the first-pass shape of this proposal before the kind-aware framing made the broader override necessary.

### Impact

- **Affected:** gogcli users in containerized / sandboxed / agent-spawned environments where the parent process can't or won't set XDG vars portably.
- **Severity:** medium — has workarounds (symlinks, wrapper scripts) but each is a small papercut accumulating across deployments. Without this knob, every deployer of gogcli has to learn gogcli's XDG dependency and engineer around it.
- **Frequency:** affects every deployment of the affected class; one-time learning curve per deployer.
- **Consequence:** lost or misplaced OAuth refresh tokens, broken keyring continuity, persistent-state drift between container restarts when the surrounding stack is XDG-hostile.

### Evidence / examples

- gogcli env-var inventory (grepped from source at tag [`v0.17.0`](https://github.com/openclaw/gogcli/tree/v0.17.0)): all use `GOG_*` prefix, zero `GOGCLI_*` exist. The proposed naming matches the established convention.
- mcporter precedent for tool-specific path override: [`src/config/path-discovery.ts`](https://github.com/openclaw/mcporter/blob/v0.11.1/src/config/path-discovery.ts) reads `MCPORTER_CONFIG` env var as the highest-precedence escape hatch (at [`v0.11.1`](https://github.com/openclaw/mcporter/tree/v0.11.1)).
- Conventional precedent for `*_HOME` umbrella vars: `CARGO_HOME`, `RUSTUP_HOME`, `GOPATH`, `npm_config_prefix`, `PIP_CONFIG_FILE`, `GH_CONFIG_DIR`, `KUBECONFIG`, `DOCKER_CONFIG` — every mature CLI exposes a tool-specific path override above the XDG default.

### Additional information

Backward compatibility: fully preserved. The new env vars and flag are *additional* precedence above the XDG layer; existing users see no behavior change unless they opt in.

Migration path once both this and the XDG-conformance proposal ship: set `GOG_HOME=/persist/gogcli` (or per-kind variants for split-lifetime layouts — durable data on a backed-up volume, cache on tmpfs) in the container entrypoint. Retires the dependency on `XDG_CONFIG_HOME` (and the other XDG vars) reaching gogcli through the parent sandbox's env sanitizer, *and* gets the right per-kind separation.

### Related references

**Companion proposal (must-land-first):**
- [gogcli#621](https://github.com/openclaw/gogcli/issues/621) — "Honor XDG Base Directory Spec for credentials, vault, state, cache, and download paths" (filed 2026-05-21, OPEN). Introduces the per-kind resolvers this proposal layers overrides on top of. Should land first; this proposal is incoherent without it.

**Concrete trigger:**
- [openclaw#84854](https://github.com/openclaw/openclaw/issues/84854) — filed 2026-05-21: OpenClaw's host-env sanitizer strips `XDG_CONFIG_HOME` and `XDG_CONFIG_DIRS` from inherited subprocess env, contrary to the XDG Base Directory Spec. Fixing that bug closes the immediate failure mode for the OpenClaw-spawned case; this feature request is the *defense-in-depth* so gogcli isn't tied to a single env-var-name's policy treatment in every container/sandbox/runner above it.

**Sibling-project precedent:**
- [`openclaw/mcporter#155`](https://github.com/openclaw/mcporter/issues/155) ("Honor XDG Base Directory Spec for config, vault, cache, and daemon paths") — closed as completed, shipped in mcporter [`v0.10.0`](https://github.com/openclaw/mcporter/tree/v0.10.0). mcporter additionally exposes `MCPORTER_CONFIG` in [`src/config/path-discovery.ts`](https://github.com/openclaw/mcporter/blob/v0.11.1/src/config/path-discovery.ts) as a tool-specific path override above the XDG default — exactly the pattern proposed here for `GOG_HOME` + per-kind overrides.
- [`openclaw/mcporter#184`](https://github.com/openclaw/mcporter/issues/184) — illustrates why a tool-specific override is valuable even when XDG support exists: embedders sometimes set `XDG_CONFIG_HOME` for an *unrelated* downstream tool, accidentally redirecting other XDG-aware tools too. A `GOG_*` override insulates gogcli from this kind of cross-tool XDG collision.

**Conventional precedent in other CLIs:**
- `CARGO_HOME`, `RUSTUP_HOME`, `GOPATH`, `npm_config_prefix`, `PIP_CONFIG_FILE`, `GH_CONFIG_DIR`, `KUBECONFIG`, `DOCKER_CONFIG` — every mature CLI exposes a tool-specific path override above the XDG default.

**XDG Base Directory Specification:**
- [freedesktop.org spec, latest revision](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) — current XDG behavior in gogcli already follows the spec for one kind (config); the companion XDG-conformance proposal extends that to all four kinds, and this proposal adds the operator-override layer above that.

**gogcli source code referenced:**
- [`internal/config/paths.go`](https://github.com/openclaw/gogcli/blob/v0.17.0/internal/config/paths.go) at tag [`v0.17.0`](https://github.com/openclaw/gogcli/tree/v0.17.0) — the central path-resolution that this proposal extends. Today's `AppName = "gogcli"` constant and `Dir()` shape are reused; the new `kindDir(kind)` resolver from the XDG-conformance proposal gains the `GOG_<KIND>_DIR` / `GOG_HOME` precedence layer described above.

steipete:
## Summary

This PR implements the #621/#622 path model as one change:

- Add kind-aware config/data/state/cache path resolution with `GOG_HOME`, per-kind `GOG_CONFIG_DIR`, `GOG_DATA_DIR`, `GOG_STATE_DIR`, `GOG_CACHE_DIR`, and global `--home`.
- Move new auth metadata, file-keyring data, service-account keys, tracking state, Gmail watch state, and backup/cache state to the correct kind-specific roots.
- Preserve legacy reads for normal migration, but keep explicit `GOG_HOME`/per-kind overrides isolated so CI, containers, and agents do not silently import host secrets/state.
- Keep Drive/Gmail download defaults unchanged unless the command already has its own output flag.
- Document path precedence, XDG behavior, compatibility rules, and container/systemd usage.

## Compatibility notes

- Relative `GOG_*_DIR` and `--home` values are rejected; relative XDG values are ignored per XDG expectations.
- Legacy OAuth credentials, service-account keys, file-keyring directories, tracking config, backup config, and Gmail watch files remain readable when no explicit gog-specific override is active.
- Existing legacy file-keyring secrets remain authoritative when both old and new data roots exist, so migration does not hide stored client secrets.
- Explicit gog-specific roots intentionally skip legacy secret/state fallback to preserve sandbox isolation.
- `--home` applies before parser/help rendering even when placed after a subcommand.

## Proof

- `make ci`
- `go test ./internal/config ./internal/secrets ./internal/cmd ./internal/googleapi ./internal/backup ./internal/tracking`
- `make lint`
- `GOOS=windows GOARCH=amd64 go test -c ./internal/tracking -o /tmp/gogcli-tracking-windows.test.exe`
- `GOOS=windows GOARCH=amd64 .tools/golangci-lint run ./internal/tracking`
- GitHub Actions PR CI run 26291047143: `test`, `worker`, `windows`, and `darwin-cgo-build` all green
- Live smoke: isolated `GOG_HOME` writes credentials metadata to `GOG_HOME/data`, config path to `GOG_HOME/config`, keeps `client_secret` out of metadata, and lists the keyring-backed client
- Live smoke: legacy credentials in the old config dir still list and remove when no explicit data override is set
- Autoreview: `/Users/steipete/Projects/agent-scripts/skills/autoreview/scripts/autoreview --mode branch --base origin/main` clean, no accepted/actionable findings

Closes #621.
Closes #622.

Thanks @alexminza for the original proposal and direction.


alexminza:
Thank you @steipete for your relentless dedication and effort to improve the projects and the world around you! Danke schön!

PR #653, opened 2026-05-29
[codex] keep Gmail label IDs case-sensitive

steipete:
## Summary

- keep Gmail label ID handling case-sensitive during label resolution
- prefer exact case-sensitive label IDs before case-insensitive label-name lookup
- keep duplicate-name preflight from case-folding label IDs while still blocking exact ID/name collisions

## Root cause

Gmail label resolution built a single lowercase lookup map for both label names and label IDs. That made custom label IDs behave case-insensitively even though Gmail IDs are opaque case-sensitive tokens, and it also made exact IDs vulnerable to case-folded label-name shadowing unless exact IDs are checked first.

## Validation

- go test ./internal/cmd -run 'TestEnsureLabelNameAvailable|TestResolveLabelIDs|TestGmailLabelsGetCmd_ExactIDBeatsCaseFoldedName|TestFetchLabelNameToID'
- make test
- live test with clawdbot@gmail.com: created temp Gmail label `Label_18`, exact `gmail labels get Label_18` succeeded, lowercased `gmail labels get label_18` failed as expected, temp label deleted, cleanup check found `tmp_label_leftovers:0`
- /Users/steipete/Projects/agent-scripts/skills/autoreview/scripts/autoreview --mode local

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Gmail: keep label IDs case-sensitive during label resolution and duplicate-name checks while still matching label names case-insensitively.
 - Docs: update the bundled `gog` agent skill to preserve broad user OAuth scopes during reauth and rely on command guards for scoped execution.
 
 ## 0.19.0 - 2026-05-22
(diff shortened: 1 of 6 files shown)


steipete:
@clawsweeper re-review

Updated after local autoreview findings:
- exact case-sensitive label IDs are checked before case-folded label names
- exact ID/name collisions stay blocked
- live clawdbot Gmail label proof added to the PR body

PR #681, opened 2026-06-03
feat(gmail): preserve draft attachments on update; add --clear-attachments

### LINKED ISSUE #680: gmail drafts update: preserve existing attachments when --attach is omitted
### Summary
`gmail drafts update` rebuilds the draft message from scratch via `buildDraftMessage` → `attachmentsFromPaths(input.Attach)`, using **only** the `--attach` paths passed on that invocation. It fetches the existing draft for thread/To/message-id fallback, but does **not** carry forward the draft's existing attachments. So updating a draft that already has attachments — e.g. to tweak the body — and omitting `--attach` **silently drops the attachments**.

### Why it matters
This is the classic Gmail `drafts.update` replaces-the-whole-draft gotcha. A caller iterating on a draft (body/subject edits) reasonably expects attachments to persist the way To/Subject/body do. Losing them is silent (no error) and only visible on a later `drafts get` / in the Gmail UI.

### Proposal (preferred: preserve-by-default)
When `--attach` is omitted on `drafts update`, re-read the existing draft's attachment parts and carry them forward; only replace attachments when the caller explicitly passes `--attach`. This matches how the command already preserves To when `--to` is omitted. (Alternative: keep replace semantics but document loudly — less surprising to preserve.)

### Repro
1. `gmail drafts create --subject S --body B --attach ./file.png` → draft has file.png.
2. `gmail drafts update <id> --body B2` (no `--attach`).
3. `gmail drafts get <id>` → file.png is gone.

(Reported via the gogcli-mcp wrapper. The wrapper documents the current replace semantics as a workaround, but preserve-by-default would be the least-surprising fix.)

chrischall:
Closes #680.

## Problem
`gmail drafts update` rebuilds the whole message from the flags on that invocation, so omitting `--attach` **silently drops the draft's existing attachments** — a data-loss footgun when editing only the body/subject of a draft that has files attached.

**Live repro on shipped `gog v0.20.0`:**
```
$ gog gmail drafts create --subject bugrepro --body v1 --attach ev.png
$ gog gmail drafts get <id>   → attachments: ['ev.png']
$ gog gmail drafts update <id> --subject bugrepro --body v2     # no --attach
$ gog gmail drafts get <id>   → attachments: NONE               # dropped
```

## Fix (addresses the product question raised on the issue)
The issue flagged that preserve-by-default needs an explicit clear/replace path. This PR does both:

- **Preserve by default:** when `--attach` is omitted (and `--clear-attachments` is not set), re-read the existing draft's attachment bytes and carry them into the rebuilt message — reusing the exact `collectAttachments` + `fetchAttachmentBytes` path `gmail forward` already uses to re-attach.
- **`--attach` replaces** attachments (explicit replacement; unchanged).
- **New `--clear-attachments`** removes all attachments; mutually exclusive with `--attach`.
- `clear_attachments` + `preserve_attachments` added to the dry-run/JSON payload.

So: edit a draft body → attachments stay; `--attach` → replace; `--clear-attachments` → intentional removal. No silent loss, and the intentional-clear workflow still has a home.

## Tests
- `TestGmailDraftsUpdateCmd_PreservesAttachmentsWhenOmitted` — wire-level: stubs the draft GET (with an attachment), the attachment-bytes GET, and the update PUT; asserts the rebuilt draft's raw MIME contains the preserved `filename="report.pdf"` and its bytes, and that the attachment-bytes endpoint was hit.
- `TestGmailDraftsUpdateCmd_ClearAttachmentsRemovesThem` — `--clear-attachments` drops the attachment and skips the byte fetch.
- `TestGmailDraftsUpdateCmd_AttachAndClearMutuallyExclusive` — guard.
- `go test ./internal/cmd` green; `make lint` 0 issues; `make docs-commands` regenerated.

## Verification notes
- The **bug** is reproduced live above on shipped `gog v0.20.0`.
- The **fix** is verified by the wire-level tests, which simulate the exact Gmail API interaction (draft GET → attachment bytes GET → PUT with the attachment re-embedded). A live run of the *patched* binary against a real account was blocked locally — the unsigned dev build can't get a non-interactive macOS Keychain grant — but the reattach mechanism is identical to the already-shipped `gmail forward` path.

(Reported via the gogcli-mcp wrapper, which currently documents the replace-semantics caveat; once this ships the wrapper drops that caveat.)

## ⚠️ Migration note (behavior change)

This intentionally changes the default semantics of `gmail drafts update`:

- **Before:** omitting `--attach` rebuilt the draft *without* attachments — i.e. an omitted `--attach` silently **cleared** them.
- **After:** omitting `--attach` **preserves** the existing attachments. To intentionally clear them, pass the new **`--clear-attachments`** flag.

Callers/scripts that relied on omitted-`--attach` to drop attachments must switch to `--clear-attachments`. `--attach <paths>` still replaces.

## Patched live proof (real Gmail account)

Ran a **patched** binary (`v0.21.1-…-8a8bc925`, this branch) against a real account — all three semantics confirmed:

```
1. create --attach ev.png            -> attachments: [(ev.png, 67)]
2. update --body "v2" (no --attach)   -> attachments: [(ev.png, 67)]   # PRESERVED (was NONE on shipped gog)
3. update --attach notes.txt          -> attachments: [(notes.txt, 11)] # REPLACED
4. update --clear-attachments         -> attachments: NONE              # CLEARED
```

Preserve-by-default, explicit `--attach` replace, and explicit `--clear-attachments` removal all verified live (scratch draft created, exercised, deleted). Combined with the shipped-gog bug repro above, this covers the before/after end to end.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Fixed
 
+- Gmail: preserve existing `gmail drafts update` attachments when `--attach` is omitted, add `--clear-attachments` for intentional removal, and keep `--attach` as explicit replacement. (#680, #681) — thanks @chrischall.
+
 ## 0.21.0 - 2026-06-01
 
 ### Added
--- a/docs/commands/gog-gmail-drafts-update.md
+++ b/docs/commands/gog-gmail-drafts-update.md
@@ -20,12 +20,13 @@ gog gmail (mail,email) drafts (draft) update (edit,set) <draftId> [flags]
 | --- | --- | --- | --- |
 | `--access-token` | `string` |  | Use provided access token directly (bypasses stored refresh tokens; token expires in ~1h) |
 | `-a`<br>`--account`<br>`--acct` | `string` |  | Account email for API commands (gmail/calendar/chat/classroom/drive/drivelabels/docs/slides/contacts/tasks/people/sheets/forms/sites/appscript/analytics/searchconsole/youtube/photos) |
-| `--attach` | `[]string` |  | Attachment file path (repeatable) |
+| `--attach` | `[]string` |  | Attachment file path (repeatable). Replaces existing attachments; omit to preserve them, or use --clear-attachments to remove all. |
 | `--bcc` | `string` |  | BCC recipients (comma-separated) |
 | `--body` | `string` |  | Body (plain text; required unless --body-html is set) |
 | `--body-file` | `string` |  | Body file path (plain text; '-' for stdin) |
 | `--body-html` | `string` |  | Body (HTML; optional) |
 | `--cc` | `string` |  | CC recipients (comma-separated) |
+| `--clear-attachments` | `bool` |  | Remove all attachments from the draft. By default, omitting --attach preserves the draft's existing attachments. |
 | `--client` | `string` |  | OAuth client name (selects stored credentials + token bucket) |
 | `--color` | `string` | auto | Color output: auto\|always\|never |
 | `--disable-commands` | `string` |  | Comma-separated list of disabled commands; dot paths allowed |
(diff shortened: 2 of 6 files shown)


chrischall:
@clawsweeper re-review — addressed both P1 items: added the migration note to the PR body, and added **patched-binary live Gmail proof** showing preserve (body-only update keeps the attachment), replace (`--attach`), and clear (`--clear-attachments`). The behavior-change/compatibility risk is now documented with the `--clear-attachments` migration path.

PR #721, opened 2026-06-10
fix(auth): repair duplicate token alias writes

### LINKED ISSUE #718: macOS Keychain: refreshed token persists but logs misleading persist warning with -25299
## Summary

On macOS Keychain, a token refresh can successfully persist the refreshed access token/expiry, but `gog` still logs:

```text
time=2026-06-09T17:00:43.727+02:00 level=WARN msg="persist refreshed token metadata failed" email=toruviel@gmail.com client=default err="store token: set token: Failed to update item in keychain: The specified item already exists in the keychain. (-25299)"
```

This warning appears misleading: immediately after the warning, exporting the token shows the refreshed expiry was stored, and subsequent auth-using commands do not warn again until the next refresh window.

## Environment

- OS: macOS / Darwin arm64
- `gog`: `v0.21.0 (2c84d89 2026-06-01T21:54:29Z)`
- Keyring backend: macOS Keychain/default backend
- Account/client: default OAuth client, Gmail/Calendar scopes among others

I also checked the relevant code path in `v0.21.0`, `v0.22.0`, `v0.23.0`, and `main`; the warning path appears unchanged.

## Repro / observed behavior

Before expiry, commands do not warn. Token export showed:

```text
access_exp=2026-06-09T15:00:14Z
```

At/after expiry, the first auth-using command refreshes and logs the warning:

```bash
gog gmail list 'is:inbox is:unread'
```

Output starts with:

```text
time=2026-06-09T17:00:43.727+02:00 level=WARN msg="persist refreshed token metadata failed" email=toruviel@gmail.com client=default err="store token: set token: Failed to update item in keychain: The specified item already exists in the keychain. (-25299)"
```

But immediately afterwards, repeated auth-using commands succeed without the warning:

```bash
gog gmail list 'is:inbox is:unread'
gog gmail list 'is:inbox is:unread'
```

and exporting the token shows that the new access token expiry *was* persisted:

```text
access_exp=2026-06-09T16:00:42Z
```

So the runtime behavior looks like:

1. stored access token expires;
2. first command refreshes successfully;
3. `SetToken` returns `-25299` from macOS Keychain and `client_auth.go` logs `persist refreshed token metadata failed`;
4. nevertheless the refreshed expiry is visible in the stored token;
5. subsequent commands do not refresh/warn until the next expiry.

## Why this seems wrong/misleading

The warning sounds like the refreshed token metadata was not persisted at all, which would imply every new `gog` process after expiry should refresh and warn again. That is not what happens: the refreshed expiry is visible in `gog auth tokens export`, and subsequent commands use it.

Looking at `internal/secrets/store.go`, `SetToken` writes multiple entries/aliases: primary token, legacy token for the default client, and subject-token alias when a subject is present. A plausible failure mode is that the primary token write succeeds, but a later alias write returns `errSecDuplicateItem` (`-25299`). If so, the current warning is too broad: it reports the whole refresh persist as failed, even though the primary token appears to have persisted.

## Expected behavior

One of:

- handle macOS `errSecDuplicateItem` during refresh alias writes by updating/replacing/retrying the specific item; or
- make the warning more precise, e.g. distinguish primary token persistence from legacy/subject alias persistence; or
- if `-25299` is benign after read-back verification confirms the primary token was updated, downgrade/suppress the warning.

## Relevant code path

`internal/googleapi/client_auth.go`:

```go
if err := p.store.SetToken(p.client, persistEmail, updated); err != nil {
    slog.Warn("persist refreshed token metadata failed", "email", persistEmail, "client", p.client, "err", err)
    return t, nil
}
```

`internal/secrets/store.go` writes primary/legacy/subject token entries via `verifiedSet(...)`. The observed error does not identify which of those writes failed, only:

```text
store token: set token: Failed to update item in keychain: The specified item already exists in the keychain. (-25299)
```

TurboTheTurtle:
## Summary

- keep primary token persistence strict while repairing duplicate failures on legacy/subject token aliases
- remove and retry alias writes when macOS Keychain reports duplicate item (`-25299`)
- verify stale alias replacement, refreshed access metadata, and strict primary behavior

Fixes #718.

## Validation

- `go test ./internal/googleapi ./internal/secrets -count=1`
- `DEVELOPER_DIR=/Library/Developer/CommandLineTools make ci`
- autoreview clean: no accepted/actionable findings
- controlled live macOS Keychain smoke with an isolated service
  - stored a stale alias in the real Keychain backend
  - injected the reported `-25299` error at the alias write boundary
  - exercised real Keychain removal, retry, and verified readback
  - confirmed refreshed alias data and cleaned up the disposable item

## Notes

The reporter supplied the natural expiry-window repro. The local live smoke used controlled duplicate injection because a fresh isolated Keychain service did not naturally reproduce the stale duplicate state.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Fixed
 
 - Docs: avoid duplicate empty paragraphs adjacent to Markdown headings while preserving body paragraph spacing. (#717, #720) — thanks @TurboTheTurtle.
+- Auth: repair duplicate macOS Keychain writes for legacy and subject token aliases without weakening primary token persistence. (#718, #721) — thanks @TurboTheTurtle.
 
 ## 0.23.0 - 2026-06-09
 
--- a/internal/secrets/store.go
+++ b/internal/secrets/store.go
@@ -505,13 +505,13 @@ func (s *KeyringStore) setTokenNoLock(client string, email string, tok Token) er
 	}
 
 	if normalizedClient == config.DefaultClientName {
-		if err := verifiedSet(s.ring, legacyTokenKey(email), payload, "legacy token"); err != nil {
+		if err := verifiedSetAlias(s.ring, legacyTokenKey(email), payload, "legacy token"); err != nil {
 			return wrapKeychainError(fmt.Errorf("store legacy token: %w", err))
 		}
 	}
 
 	if tok.Subject != "" {
-		if err := verifiedSet(s.ring, subjectTokenKey(normalizedClient, tok.Subject), payload, "subject token"); err != nil {
+		if err := verifiedSetAlias(s.ring, subjectTokenKey(normalizedClient, tok.Subject), payload, "subject token"); err != nil {
 			return wrapKeychainError(fmt.Errorf("store subject token: %w", err))
 		}
 	}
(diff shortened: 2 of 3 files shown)


steipete:
Landed as [`ebc8218b`](https://github.com/openclaw/gogcli/commit/ebc8218b609a49d94eb158b4d8a9bc6437606659).

Validation:

- focused auth/secrets tests passed
- full local `make ci` passed
- autoreview clean: no accepted/actionable findings
- controlled isolated macOS Keychain smoke exercised stale alias removal, retry, and verified readback through the real backend
- exact-head Linux, Windows, Darwin CGO, worker, Docker, and Socket checks passed

Thanks @TurboTheTurtle.

PR #740, opened 2026-06-11
fix(docs): preserve inline replacement paragraphs

### LINKED ISSUE #736: bug(docs find-replace --format markdown): replacement appends a trailing empty paragraph
### What happens

Replacing inline text inside an existing paragraph with `docs find-replace --format markdown` inserts an extra empty paragraph after the edited paragraph. Repeated markdown replacements on the same area accumulate double blank lines.

### Repro

Start with a paragraph followed by a single blank line, then a next paragraph:

```
First paragraph ending here.

Second paragraph.
```

```
gog docs find-replace <id> "ending here." "ending here. Plus more." --format markdown
```

Result: the first paragraph is now followed by TWO blank paragraphs instead of one.

### Expected

Replacing inline text within a paragraph should not introduce paragraph breaks — the replacement should stay within the target paragraph, leaving surrounding blank lines unchanged.

### Note

The default `--format plain` path does not appear to add the empty paragraph; this is specific to `--format markdown`.

### Environment

`gog` v0.23.0 (and current `main`).

steipete:
## Summary

- keep single-paragraph Markdown find-replace fragments inside the existing paragraph
- preserve explicit terminal newlines and block Markdown behavior
- add regression coverage for inline and explicit-newline replacements

## Verification

- focused Docs find-replace tests
- autoreview clean
- `make ci`
- live disposable Google Doc: two successive Markdown replacements produced exact paragraph spacing on raw readback; document trashed afterward

Closes #736

Thanks @sebsnyk.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 
 ### Fixed
 
+- Docs: keep inline Markdown find-replace fragments inside their existing paragraph unless the replacement explicitly ends with a newline. (#736) — thanks @sebsnyk.
 - Docs: render HTML `<br>` variants as line breaks inside Markdown table cells while preserving protected literals. (#730) — thanks @sebsnyk.
 - Docs: avoid duplicate empty paragraphs adjacent to Markdown headings while preserving body paragraph spacing. (#717, #720) — thanks @TurboTheTurtle.
 - Auth: repair duplicate macOS Keychain writes for legacy and subject token aliases without weakening primary token persistence. (#718, #721) — thanks @TurboTheTurtle.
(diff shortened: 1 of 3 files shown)


steipete:
Landed in d6e7a8106e85a3abf7705723cb96e3c89878c913.

Proof:
- focused find-replace regression tests passed
- autoreview clean
- full local `make ci` passed
- exact-head GitHub CI green on Linux, macOS, Windows, worker, and Docker
- live disposable Google Doc verified two successive Markdown replacements kept exact inline paragraph spacing; test document trashed

Thanks @sebsnyk.

PR #741, opened 2026-06-11
fix(docs): accept cell list content

### LINKED ISSUE #733: bug(docs cell-update): markdown list content is silently ignored (no update, no error)
### What happens

`gog docs cell-update` silently ignores Markdown **list** content. Passing a bullet (or numbered) list as `--content` leaves the cell unchanged and returns no error — the command reports success but the cell keeps its previous content.

Plain multi-line content with real newlines updates the cell correctly (it creates paragraph boundaries), so the cell model is fine; only list-markdown input is dropped.

### Repro

```
# silently does nothing — cell keeps old content, no error
gog docs cell-update <id> --table-index 1 --row 2 --col 2 --content $'- one\n- two'

# works — produces two paragraphs in the cell
gog docs cell-update <id> --table-index 1 --row 2 --col 2 --content $'one\ntwo'
```

### Expected

Either render a real list in the cell, or return a clear error that list block content is not supported via `cell-update`. A silent no-op is the trap — callers cannot tell the write was dropped.

### Environment

`gog` v0.23.0 (also reproduces on current `main`). `--content` defaults to markdown format.

steipete:
## Summary

- accept Markdown bullet-list values passed as `--content $'- one\n- two'`
- preserve existing argument behavior outside `docs cell-update`
- reject nonempty Markdown that produces no editable cell text
- document native bullet and numbered list support

## Verification

- focused cell-update and full CLI parser tests
- autoreview clean
- `make ci`
- live disposable Google Doc verified bullet and numbered replace and append behavior from raw Docs metadata
- live ineffective-Markdown check returned nonzero and preserved the cell; document trashed afterward

Closes #733

Thanks @sebsnyk.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 
 ### Fixed
 
+- Docs: accept leading-dash Markdown list values in `docs cell-update --content` and reject nonempty Markdown that produces no editable cell text. (#733) — thanks @sebsnyk.
 - Docs: keep inline Markdown find-replace fragments inside their existing paragraph unless the replacement explicitly ends with a newline. (#736) — thanks @sebsnyk.
 - Docs: render HTML `<br>` variants as line breaks inside Markdown table cells while preserving protected literals. (#730) — thanks @sebsnyk.
 - Docs: avoid duplicate empty paragraphs adjacent to Markdown headings while preserving body paragraph spacing. (#717, #720) — thanks @TurboTheTurtle.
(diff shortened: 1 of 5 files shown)


steipete:
@clawsweeper automerge

steipete:
Landed in b252845077f57e2daca6838bc47289a89ed88174.

Proof:
- focused cell-update and full CLI parser tests passed
- autoreview clean
- full local `make ci` passed
- exact-head GitHub CI green on Linux, macOS, Windows, worker, and Docker
- live disposable Google Doc verified native bullet and numbered replace and append behavior using raw Docs metadata
- live ineffective-Markdown check failed clearly and preserved the cell; test document trashed

Thanks @sebsnyk.

PR #742, opened 2026-06-11
fix(docs): reset markdown replacement styles

### LINKED ISSUE #735: bug(docs find-replace --format markdown): leading **bold** span in replacement is mis-paired
### What happens

When the replacement string passed to `docs find-replace --format markdown` **begins** with a bold span, the leading `**` is dropped/mishandled. Asterisk pairing shifts by one, so the wrong ranges become bold.

### Repro

```
gog docs find-replace <id> "PLACEHOLDER" "**Alpha** keeps text. The **Bravo** keeps text." --format markdown
```

Expected bold: `Alpha` and `Bravo`.
Actual: the spans *between* and *after* the markers are bolded instead — `keeps text. The` and `keeps text.` come out bold, while `Alpha` and `Bravo` are plain.

### Note

A replacement that does NOT start with a bold span (i.e. there is leading text before the first `**`) pairs correctly. The trigger is a `**bold**` span at position 0 of the replacement string.

### Environment

`gog` v0.23.0 (and current `main`).

steipete:
## Summary

- reset inherited text styles across the inserted Markdown replacement range
- apply generated Markdown styles after the reset
- keep synthetic paragraph prefixes and surrounding document text outside the reset
- share the existing deterministic Docs text-style reset helper

## Verification

- focused find-replace, Markdown, inline, and brace reset tests
- autoreview clean
- `make ci`
- live exact reporter case and leading-plain control
- live styled-boundary matrix confirmed only `Alpha` and `Bravo` bold while surrounding text retained its prior style
- disposable documents trashed

Closes #735

Thanks @sebsnyk.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 
 ### Fixed
 
+- Docs: reset inherited text styles before applying Markdown find-replace formatting so leading bold spans and later inline styles stay paired correctly. (#735) — thanks @sebsnyk.
 - Docs: accept leading-dash Markdown list values in `docs cell-update --content` and reject nonempty Markdown that produces no editable cell text. (#733) — thanks @sebsnyk.
 - Docs: keep inline Markdown find-replace fragments inside their existing paragraph unless the replacement explicitly ends with a newline. (#736) — thanks @sebsnyk.
 - Docs: render HTML `<br>` variants as line breaks inside Markdown table cells while preserving protected literals. (#730) — thanks @sebsnyk.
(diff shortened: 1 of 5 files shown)


steipete:
Landed in 9900654cf3a2f1b1c2eb4a8e0035983695262d3b.

Proof:
- focused find-replace, Markdown, inline, and shared reset tests passed
- autoreview clean
- full local `make ci` passed
- exact PR-event GitHub CI green on Linux, macOS, Windows, worker, Docker, and Socket checks
- live exact reporter case and leading-plain control passed
- live styled-boundary matrix confirmed only `Alpha` and `Bravo` bold while surrounding styles remained unchanged; test documents trashed

Thanks @sebsnyk.

PR #743, opened 2026-06-11
feat(calendar): update event attachments

### LINKED ISSUE #738: feat(calendar update): support --attachment flag (currently only on create)
## Gap

`gog calendar create` accepts `--attachment=URL` (repeatable), but `gog calendar update` does not expose the flag. The corresponding Calendar API operation supports attachments on PATCH via `supportsAttachments=true`, so the underlying capability is there.

## Why this matters

Attachments often need to be added **after** an event exists:
- Recurring meetings where each occurrence gets its own agenda doc
- Attaching meeting notes / pre-reads to events organized by someone else
- Updating attachments without recreating the event (which isn't possible when you're an attendee, not the organizer)

## Current workaround

Minting an access token from gog's keychain refresh token and calling the API directly:

```bash
REFRESH=\$(security find-generic-password -s gogcli -a \"token:default:<email>\" -w | \\
  python3 -c \"import sys,json; print(json.load(sys.stdin)['refresh_token'])\")
# ... mint token via /oauth2/token endpoint ...
curl -X PATCH \"https://www.googleapis.com/calendar/v3/calendars/primary/events/<eventId>?supportsAttachments=true\" \\
  -H \"Authorization: Bearer \$TOKEN\" -H \"Content-Type: application/json\" \\
  -d '{\"attachments\":[{\"fileUrl\":\"...\",\"title\":\"...\",\"mimeType\":\"...\",\"fileId\":\"...\"}]}'
```

This works but requires bypassing the CLI for what feels like a first-class operation.

## Suggested shape

Mirror the `create` command's flag on `update`:

```
--attachment=ATTACHMENT,...    File attachment URL (can be repeated)
```

Open question for the maintainer: replace-all vs append semantics. Probably:
- `--attachment` → replace
- `--add-attachment` → append (matching the `--attendees` / `--add-attendee` pattern already in update)

Also surface the `supportsAttachments=true` API param implicitly when the flag is set.

## Repro

```
$ gog calendar update <calendarId> <eventId> --attachment=https://drive.google.com/...
unknown flag --attachment
$ gog calendar update --help | grep attach
(no output)
$ gog calendar create --help | grep attach
      --attachment=ATTACHMENT,...    File attachment URL (can be repeated)
```

Version: gog v0.23.0 (71a8504), macOS arm64.

steipete:
## Summary

- add repeatable `--attachment` to `calendar update`
- replace the complete attachment array when URLs are supplied
- clear attachments with `--attachment=`
- send `supportsAttachments=true` for replacement and clear
- expose attachment support accurately in dry-run output

## Verification

- patch construction, empty-array serialization, query parameter, and dry-run tests
- generated command docs
- autoreview clean
- `make ci`
- live owned Drive Doc attached to a disposable Calendar event, read back, cleared, and read back again
- event deleted and Drive file trashed

Closes #738

Thanks @TreyLawrence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Calendar: add repeatable `--attachment` to `calendar update` for replacing or clearing event attachments. (#738) — thanks @TreyLawrence.
 - Sheets: add `sheets validation` get/set/clear commands for dropdown, checkbox, number, date, range, and custom-formula rules, and preserve table-managed dropdowns during validation-only copy/paste. (#710) — thanks @chrischall.
 - Sheets: add table-aware `sheets delete-dimension` for deleting row or column spans while preserving intersecting table objects and remaining data. (#711) — thanks @chrischall.
 - Docs: add direct `docs table-row`, `docs table-column`, `docs table-merge`, and `docs table-unmerge` commands with index, header-text, all-table, and tab-aware selection. (#686) — thanks @sebsnyk.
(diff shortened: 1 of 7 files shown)


steipete:
Landed as `091581b275f2a550980cfe7c2b613c0330095c19`.

Verification:
- Added update-path unit coverage for attachment query flags, replacement and explicit clearing, plus dry-run mutation reporting.
- Codex autoreview: no actionable findings.
- Local: `make ci` passed.
- GitHub: exact-head pull-request CI run `27319902084` and Docker run `27319902107` passed for `38f33342a8c2c328bc65c46f606f5f6fb05c238b`.
- Live Google proof with `clawdbot@gmail.com`: created a disposable Drive Doc and Calendar event, attached the Doc through `calendar update --attachment`, read back one attachment, cleared with `--attachment=`, read back zero attachments, then deleted the event and trashed the Doc.

Thanks @TreyLawrence for the report.

PR #750, opened 2026-06-11
feat(agent): add safe capability discovery

### LINKED ISSUE #677: Support auth.md-based capability discovery for agent and MCP clients
## Problem

gogcli is explicitly built for coding agents, CI pipelines, and MCP clients. It already has strong foundations for agent-safe execution: named OAuth clients, keyring backends, command allowlists/denylists (`--enable-commands-exact`), baked safety-profile binaries, and a typed MCP server (`gog mcp`) that deliberately avoids a generic command bridge.

However, there is currently no structured way for an agent or MCP client to discover:

- which scopes/capabilities are available before invoking tools
- what authentication methods gogcli supports in a given deployment
- how to negotiate a scoped delegated session rather than inheriting full account privileges
- whether the running instance has write access or is in a read-only safety profile

Right now, agents must either be pre-configured out-of-band or probe blindly. This creates friction in autonomous harnesses and multi-agent orchestration where the agent runtime doesn't know what the tool is allowed to do.

---

## Proposed Feature

Add optional support for `auth.md`-based capability discovery, specifically tailored to gogcli's existing agent-safety model.

Building on what already exists:

- `gog schema --json` exposes the command/flag schema
- `--enable-commands-exact` defines allowed command surfaces
- safety profiles bake in runtime guards
- `gog mcp` exposes a typed, read-only-by-default MCP stdio server

The proposal is to extend this with a machine-readable auth and capability document that agent clients can discover at startup.

### What gogcli could expose

- available authentication methods (OAuth, ADC, service account, direct access token)
- active account and its authorized scopes (e.g. `gmail.modify`, `drive`, `calendar`)
- current command surface: which commands are allowed/denied for this process
- whether write operations are permitted or blocked (safety profile status)
- token expiry hints so agents can proactively re-authenticate before a long task
- MCP tool availability filtered by current auth scope

### Possible interfaces

```bash
# Machine-readable auth + capability snapshot
gog agent auth-info --json

# Or exposed as an MCP tool for agent clients
# mcp tool: gog_auth_info -> returns scopes, allowed commands, write status
```

For HTTP/SSE deployments of the MCP server, this could also be exposed as a well-known endpoint:

```
GET /.well-known/auth.md
GET /.well-known/capabilities.json
```

---

## Why This Matters for gogcli Specifically

gogcli is increasingly used inside agent runtimes (Claude Code, OpenClaw, Codex, Gemini CLI) via the MCP server. In these environments:

- the agent doesn't know upfront if the gogcli instance has `drive` write access or is read-only
- the agent can't tell if `--gmail-no-send` is active without attempting a send
- safety profiles are baked into the binary — there's no way for an agent to introspect them without trying a blocked command and getting an error

Structured capability discovery would let the agent adapt its behavior before calling tools, rather than learning constraints through failures.

---

## Suggested Implementation Ideas

Possible architecture (lightweight, non-breaking):

1. New `gog agent auth-info` subcommand under the existing `gog agent` group
2. Returns JSON with: active account, authorized scopes, allowed command surface, write status, token expiry
3. MCP tool `gog_capabilities` registered alongside existing typed tools in `gog mcp`
4. Optional: `/.well-known/capabilities.json` for HTTP transport mode

Potential flow for an agent runtime:

1. Agent starts gogcli MCP server
2. Calls `gog_capabilities` tool immediately
3. Learns: read-only, gmail+drive scopes, no send, token valid for 45 min
4. Adapts tool calls accordingly — no blind 403s, no wasted retries

---

## Security Considerations

The implementation should avoid:

- exposing raw refresh tokens or client secrets in the capability document
- returning keyring passwords or ADC key material
- leaking account email in shared/multi-tenant MCP deployments

Prefer:

- scopes and capabilities only — no credential material
- redacted account identity (configurable)
- the same access controls that already gate `gog agent` commands
- read-only by default, consistent with `gog mcp` philosophy

---

This would make gogcli a first-class citizen in agentic workflows — not just a tool that agents call, but one that agents can reason about before calling.

---

## Reference

Official `auth.md` specification and format: https://github.com/workos/auth.md

steipete:
## Summary

- add `gog agent capabilities` (`auth-info`) with a versioned runtime capability snapshot
- keep default discovery offline, account-redacted, and keyring-free
- add read-only `gog_capabilities` MCP discovery for the server's actual filtered tool surface
- make credential metadata and account identity separate explicit opt-ins
- document that this is gog-specific capability discovery, not auth.md protocol conformance or an HTTP endpoint

## Safety decisions

- no token, client-secret, service-account-key, or credential-path fields
- `--include-auth` is the only path that may open the configured keyring
- `--include-account` controls identity disclosure independently
- report configured command rules instead of claiming a misleading global read/write state

## Proof

- `make ci`
- `go test ./internal/cmd -run 'TestAgentCapabilities|TestMCPCapabilities|TestMCPEnabledTools|TestOpen' -count=1`
- offline live smoke: default JSON, explicit account disclosure, and filtered `mcp --list-tools`
- structured autoreview: clean, no accepted/actionable findings

Closes #677

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ It is built for terminals, shell scripts, CI, and coding agents:
 - runtime command allowlists/denylists and baked safety-profile binaries
 - typed [MCP server](docs/mcp.md) for agent clients, read-only by default and
   without a generic command runner
+- offline-by-default [agent capability discovery](docs/agent-capabilities.md)
+  for runtime safety and MCP tool exposure
 - read-only audit/reporting commands for risky surfaces like Drive and Contacts
 - [generated docs](docs/commands/README.md) for every command
 
@@ -25,6 +27,7 @@ Start here:
 - [Install](docs/install.md)
 - [Quickstart](docs/quickstart.md)
 - [Auth clients and service accounts](docs/auth-clients.md)
+- [Agent capabilities](docs/agent-capabilities.md)
 - [MCP server](docs/mcp.md)
 - [Command index](docs/commands/README.md)
 - [Gmail watch / Pub/Sub push](docs/watch.md) (<https://gogcli.sh/watch.html>)
--- a/docs/agent-capabilities.md
+++ b/docs/agent-capabilities.md
@@ -1,0 +1,69 @@
+---
+title: Agent capabilities
+description: "Inspect gog authentication disclosure, runtime safety rules, and MCP tool exposure before calling Google APIs."
+---
+
+# Agent capabilities
+
+`gog agent capabilities` emits a local runtime snapshot for agents, CI jobs,
+and MCP clients. The default command is offline: it does not open the keyring,
+inspect stored credentials, or print an account identity.
+
+```bash
+gog agent capabilities --json
+```
+
+The response includes:
+
+- supported authentication methods
+- active dry-run and Gmail no-send state
+- runtime command allow and deny rules
+- baked safety-profile name and status
+- the command for retrieving the full visible CLI schema
+
+The snapshot reports configured command rules rather than claiming a single
+global read/write state. Different commands can have different safety and OAuth
+requirements.
+
+## Credential metadata
+
+Credential inspection is explicit because it can access the configured keyring:
+
+```bash
+gog --account you@example.com agent capabilities --include-auth --json
+```
+
+`--include-auth` may add the active authentication method, OAuth client,
+granted services and scopes, and cached access-token expiry. It never returns
+access tokens, refresh tokens, client secrets, service-account keys, or
+credential paths.
+
+Account identity remains omitted unless separately requested:
+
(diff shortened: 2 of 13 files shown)


steipete:
Closing in favor of a CLI-wide design.

`gog agent` is the wrong namespace: machine-readable output, exit codes, schema discovery, safety controls, and capability discovery are contracts of the whole CLI. The replacement will remove the redundant `agent` command group, keep `gog exit-codes` top-level, add top-level `gog capabilities`, and expose only redacted, server-controlled capability data through MCP.

This also addresses the review finding that the proposed MCP tool let callers request account and credential metadata without server-operator opt-in. Credential inspection will remain an explicit local CLI action, not an MCP tool argument.

PR #762, opened 2026-06-12
fix(gmail): archive complete threads explicitly

### LINKED ISSUE #752: bug(gmail archive): given a thread ID, archives only one message but reports success (no thread-level archive)
### What happens

`gog gmail archive` only accepts **message IDs**, but `gog gmail search` / list output presents a **thread ID** in its first column. When a user archives the IDs they just listed (the natural workflow), any **multi-message thread** has only its first message archived — the rest stay in the inbox. The command still prints `Archived N messages`, so the partial failure is silent.

### Repro

```
gog gmail search "in:inbox" --max 20 -p     # first column = thread IDs
gog gmail archive <thread_id_of_a_multi_message_thread>
# -> "Archived 1 messages", but the thread is still in the inbox
gog gmail search "in:inbox" --max 20 -p     # thread still present
```

### Why it's a bug

The Gmail API supports thread-level archiving directly: `users.threads.modify` with `removeLabelIds: ["INBOX"]` removes the label from **every** message in the thread. So archiving by thread is a first-class operation, not something the caller should have to decompose into per-message calls.

`gog gmail threads modify <threadId> --remove INBOX` already does the right thing — but `archive` (the obvious command, fed the IDs that search emits) does not, and gives no signal that it only touched one message.

### Suggested fix

Any of:
- Accept thread IDs on `gmail archive` and route them through `threads.modify` (detect, or add a `--thread` flag).
- Make `archive` thread-aware by default (archive all messages of the thread an ID belongs to).
- At minimum, when an ID resolves to a thread with more messages than were archived, warn instead of reporting a clean success.

### Environment

`gog` v0.24.0-1.

steipete:
## Summary

- add explicit `gmail archive --thread <threadId>...` semantics using Gmail `users.threads.modify`
- preserve existing positional message-ID and `--query` behavior unless `--thread` is set
- reject ambiguous `--thread --query` combinations before auth
- continue across per-thread API failures, emit retry-safe per-thread JSON/plain results, then return nonzero for partial failure
- update generated command docs and changelog

Closes #752.

## Verification

- focused archive dry-run, API request, normalization, query-conflict, success-output, and partial-failure tests
- full `make ci` (629 generated command pages, 23 feature pages)
- structured autoreview: clean after adding retry-safe partial-failure reporting
- live Gmail E2E with `clawdbot@gmail.com` and sending disabled:
  - selected an existing two-message test thread with no `INBOX` labels
  - added `INBOX` to the whole thread and verified both messages entered the inbox
  - ran `gmail archive --thread`; verified both messages lost `INBOX`
  - verified JSON count/failure/result fields
  - verified final message labels exactly matched the original state

## CLI contract

`--thread` applies only to positional IDs. Gmail message and thread IDs are opaque, so the command does not guess their type. `--query` retains message-level semantics.

## Risk

Low/medium. The behavior is additive and explicit. Multiple thread API calls are not atomic; partial outcomes are fully reported and the command exits nonzero when any thread fails.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - CLI: add Git-style `gog help <command>`, make explicit output flags override environment defaults, validate color and JSON-only transforms before command execution, report early usage errors on stderr, and reject contradictory schema plain output.
 - Docs: prevent multi-paragraph Markdown range replacements from inheriting the matched paragraph's heading or list style. (#756) — thanks @sebsnyk.
 - Docs: preserve nested Markdown list levels as native bullets inside imported and updated table cells. (#749) — thanks @sebsnyk.
+- Gmail: add explicit `gmail archive --thread` semantics so IDs from thread search can archive every message in each thread. (#752) — thanks @sebsnyk.
 
 ## 0.24.0 - 2026-06-11
 
--- a/docs/commands.generated.md
+++ b/docs/commands.generated.md
@@ -364,7 +364,7 @@ Generated from `gog schema --json`.
       - [`gog forms (form) watch (watches) list (ls) <formId>`](commands/gog-forms-watch-list.md) - List active watches
       - [`gog forms (form) watch (watches) renew (refresh) <formId> <watchId>`](commands/gog-forms-watch-renew.md) - Renew a watch (extends 7 days)
   - [`gog gmail (mail,email) <command> [flags]`](commands/gog-gmail.md) - Gmail
-    - [`gog gmail (mail,email) archive [<messageId> ...] [flags]`](commands/gog-gmail-archive.md) - Archive messages (remove from inbox)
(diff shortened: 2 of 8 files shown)


steipete:
Landed as 3f653c7246561e89dc1f8a141bdec936b4cb6c23.

Proof:
- focused dry-run, normalization, API request, query-conflict, success, and partial-failure tests pass
- full `make ci` pass; 629 generated command pages and 23 feature pages
- autoreview found partial-failure retry ambiguity; fixed with per-thread continuation/results and nonzero summary; final autoreview clean
- exact-head CI green on duplicate Linux, Windows, Darwin/cgo, worker, and Docker checks; one hung Windows runner was cancelled and rerun successfully
- live Gmail E2E with sending disabled: added `INBOX` to both messages in an existing two-message test thread, `gmail archive --thread` removed it from both, JSON reported success, and final labels exactly matched the original state

PR #772, opened 2026-06-13
fix(auth): Avoid keyring prompt storms

lox:
Keychain-backed reads can currently fan out into extra prompts in read-oriented commands. `auth tokens export` uses the normal token getter, which can migrate legacy default-client tokens by writing a primary key and reading it back, and `auth list` falls back to per-token probing after a keyring timeout.

This changes export to use a read-only token path when the store supports it, keeps legacy migration for normal token reads, and stops `auth list` fallback probing once the keyring has already timed out. It also preserves the actual legacy-read error when the primary key is missing but the legacy key read fails.

The user-facing effect is that export/list paths avoid turning one blocked Keychain operation into multiple password prompts while retaining existing legacy-token compatibility.


steipete:
Maintainer live proof on macOS Keychain, exact head `90ca2ca99302fa94ad967c0609243ab0135a982e`:

```text
$ GOG_KEYRING_BACKEND=keychain /usr/bin/time -p bin/gog --json --no-input auth list
read token for [redacted]: read token: keyring connection timed out after 10s while reading keyring item
real 10.04
```

The command stopped at the first 10-second timeout instead of entering the old `Keys` plus per-token fallback traversal.

```text
$ GOG_KEYRING_BACKEND=keychain /usr/bin/time -p bin/gog --json --no-input auth tokens export [redacted] --out /dev/null --overwrite
read token: keyring connection timed out after 10s while reading keyring item
real 10.04
```

Export performed one read attempt, wrote no token file, and did not enter the legacy migration write path. Unit regression coverage and `GOTOOLCHAIN=go1.26.2 make ci` also pass locally.

@clawsweeper re-review

steipete:
Landed with current-main integration and contributor credit:

- Squash commit: `c58b381c97ade39a6418acb37efa0ad65f72f9b4`.
- Token export reads legacy entries without migrating or writing them.
- Auth listing stops after a keyring timeout from either bulk listing or a per-token fallback read instead of traversing the remaining Keychain entries.
- Legacy read failures remain visible rather than being collapsed into not-found errors.
- Added focused regression coverage and a changelog entry thanking @lox.

Verification:

- `GOTOOLCHAIN=go1.26.2 make ci`
- `/Users/steipete/Projects/agent-skills/skills/autoreview/scripts/autoreview --mode branch --base origin/main --parallel-tests "GOTOOLCHAIN=go1.26.2 make ci" --stream-engine-output` returned no actionable findings.
- Exact-head GitHub checks passed for `a32c0fd4456b1ac35df9f210a68610fc475332b8`.
- Live macOS Keychain proof with `GOG_KEYRING_BACKEND=keychain`: `auth list` stopped after one 10-second timeout, and token export against the same legacy entry performed one 10-second read attempt with no migration write or token file output.

Thanks @lox.

PR #794, opened 2026-06-13
fix(docs): keep batch reads side-effect-free

steipete:
## Summary

- keep `batch list`, `batch show`, and batch target validation on lock-free atomic reads
- validate batch IDs before local state access
- make `batch end --dry-run` read and render the persisted payload without creating directories or lock files
- retain exclusive locking for every real batch mutation

## Verification

- focused batch tests
- structured Codex autoreview: no actionable findings
- `make ci`
- isolated binary smoke: absent/invalid list/show/end paths created zero files
- isolated existing-state smoke: list/show/end dry-run preserved state bytes and created no lock
- complete 556-path visible/hidden replay: unchanged classifications, zero timeouts, file rows reduced from eight to five intentional keyring locks

No agent transcript attached pending maintainer opt-in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Fixed
 
+- Docs: keep batch list/show, batch target validation, and batch end dry-runs read-only without creating state directories or lock files.
 - Gmail: make `watch status` read atomic watch state without creating state directories or lock files.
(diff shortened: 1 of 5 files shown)


steipete:
@clawsweeper re-review

Please review exact head `68ea4297`. Local structured autoreview, full `make ci`, focused tests, and isolated binary state/lock proof are green.

PR #832, opened 2026-06-17
feat(slides): support public image URLs

### LINKED ISSUE #825: slides: insert-image/replace-slide fail with publishOutNotPermitted on restricted tenants — need a by-URL path
### Problem
`gog slides insert-image` (and `replace-slide`) upload the image to Drive and then make it publicly shareable so Slides can fetch it. On tenants where Drive public sharing is disabled by policy, this fails:

```
gog slides insert-image --width=120 <deck> <slideId> /tmp/x.png
# Google API error (400 publishOutNotPermitted)
```

This makes image insertion impossible on locked-down org tenants. The Docs side has a more flexible path (`insert-image --at` can take a URL). Slides has no by-URL option, so the only workaround is to self-host the image (e.g. a temporary tunnel) and call the Slides API `createImage` with that URL directly — which Google fetches and caches without anything being shared from Drive.

### Repro
On a tenant with Drive public sharing disabled: the command above returns `400 publishOutNotPermitted`.

### Suggested
- Add a by-URL insert path for `gog slides insert-image` (accept an `http(s)` URL, pass it straight to `createImage`), mirroring Docs.
- Or document the publish requirement + the URL workaround.

gog 0.27.1.

steipete:
## Summary

- add mutually exclusive `--url` sources to `slides insert-image` and `slides replace-slide`
- require explicit URL dimensions for insertion instead of fetching remote content locally
- bypass Drive upload/public-sharing services entirely for URL sources
- preserve the existing positional local-file flow and improve failed temp-file cleanup reporting
- update generated command docs and changelog

Closes #825.

## Testing

- `go test ./internal/cmd -run 'Test(ResolveSlidesImageSource|SlidesInsertImage|SlidesReplaceSlide)' -count=1`
- `make ci`
- `/Users/steipete/Projects/agent-skills/skills/autoreview/scripts/autoreview --mode local --stream-engine-output` (clean after fixing its accepted path-preservation finding)
- live Google Slides E2E with the designated Clawdbot test OAuth account:
  - created a temporary presentation
  - inserted the README banner through a public HTTPS URL
  - read the slide back and verified the returned image object
  - replaced it through a second public HTTPS URL
  - polled read-back until the same object ID exposed changed cached content
  - moved the temporary presentation to Trash and verified `trashed=true`

## Caveat

The live account permits Drive sharing, so it cannot reproduce the tenant policy error itself. Unit tests prove both URL commands never initialize Drive, while the live test proves Google Slides accepts and caches the direct URL requests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 0.28.1 - Unreleased
 
+### Added
+
+- Slides: allow `insert-image` and `replace-slide` to use public HTTPS image URLs without temporary Drive sharing. (#825) — thanks @sebsnyk.
+
 ### Fixed
 
 - Docs: update the Docker authentication example to persist file-keyring tokens with `GOG_HOME`. (#828, #830) — thanks @WadydX.
--- a/docs/commands.generated.md
+++ b/docs/commands.generated.md
@@ -593,12 +593,12 @@ Generated from `gog schema --json`.
     - [`gog slides (slide) delete-slide <presentationId> <slideId>`](commands/gog-slides-delete-slide.md) - Delete a slide by object ID
     - [`gog slides (slide) export (download,dl) <presentationId> [flags]`](commands/gog-slides-export.md) - Export a Google Slides deck (pdf|pptx)
     - [`gog slides (slide) info (get,show) <presentationId>`](commands/gog-slides-info.md) - Get Google Slides presentation metadata
-    - [`gog slides (slide) insert-image --width=FLOAT-64 <presentationId> <slideId> <image> [flags]`](commands/gog-slides-insert-image.md) - Insert an image at a position and size on an existing slide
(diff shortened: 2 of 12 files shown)


steipete:
Landed as [`32a6dddf`](https://github.com/openclaw/gogcli/commit/32a6dddf8d86e0422b16351007fdbb21fd2dc2d7).

Proof:
- focused URL-source, insert, replace, validation, dry-run, and Drive-bypass tests passed
- full local `make ci` passed
- final Codex autoreview passed with no accepted/actionable findings after fixing its local-path preservation finding
- live Google Slides E2E passed with the designated Clawdbot test OAuth account: create temporary deck, insert public URL, read back image, replace from second URL, verify the same image object exposed changed cached content
- cleanup verified: the temporary deck was moved to Trash and returned `trashed=true`
- GitHub Actions green: both CI runs, Docker image, Linux, Windows, Darwin cgo, and worker checks

Caveat: the live account does not enforce the reporter's restricted-tenant sharing policy. The regression tests prove the new URL paths never initialize Drive, while the live test proves the direct Slides API fetch/cache path works end to end.

PR #844, opened 2026-06-18
feat(calendar): add --timezone flag to event create

malob:
## Summary

`calendar create` requires `--start-timezone` and `--end-timezone` to be set
separately to label an event's start/end with a named IANA zone. This adds a
single `--timezone`/`--tz` flag that applies one zone to both endpoints.

- `--timezone <IANA>` / `--tz <IANA>` sets the timezone on both `--from` and `--to`.
- Mutually exclusive with `--start-timezone`/`--end-timezone` — passing it
  alongside either is a usage error (the granular flags remain for the
  asymmetric case).
- Mirrors the existing `--timezone`/`--tz` flag on `calendar create-calendar`.
- Invalid-zone / all-day errors are attributed to the flag the user actually set.

This flag is specific to `calendar create`. `calendar update` could gain the same
convenience later, but its timezone handling is more involved — it depends on
which of `--from`/`--to` is being changed — so it's intentionally left out of
scope here.

## Why

For most events the start and end are in the same timezone — wanting different
zones for each is the unusual case. Setting that single zone today means
repeating it across both `--start-timezone` and `--end-timezone`, which is fiddly
and easy to get half-right (one set, the other forgotten). `--timezone` collapses
the common case to one flag and removes that easy mistake, while the
mutual-exclusivity check keeps the rarer split-zone case explicit.

## Behavior proof

Live round-trip against a real account (account/event id redacted; throwaway
event deleted afterward). Both endpoints persist the named zone, and the event
is gone after delete:

```console
$ ./bin/gog calendar create primary --summary "tz flag test" \
    --from 2026-08-13T13:40:00-07:00 --to 2026-08-13T14:40:00-07:00 \
    --timezone America/Los_Angeles --send-updates none --account <redacted> --json
# created — event id <event-id>

# read it back (start/end excerpt of the --json event) — both carry the named zone:
$ ./bin/gog calendar event primary <event-id> --account <redacted> --json
  "start": { "dateTime": "2026-08-13T13:40:00-07:00", "timeZone": "America/Los_Angeles" },
  "end":   { "dateTime": "2026-08-13T14:40:00-07:00", "timeZone": "America/Los_Angeles" },

$ ./bin/gog calendar delete primary <event-id> --force --send-updates none --account <redacted> --json
{"calendarId":"primary","deleted":true,"eventId":"<event-id>"}

# confirm gone (soft-deleted):
$ ./bin/gog calendar event primary <event-id> --account <redacted> --json
  "status": "cancelled"
```

## Testing

- `make fmt-check`, `make lint`, `make docs-check` clean; regenerated the command
  reference with `make docs-commands`.
- `go test ./internal/cmd -run 'UnifiedTimezone|ResolveUnifiedTimezone'` — covers
  set-both, the conflict error (exit 2), and invalid-zone/all-day error attribution.

## Compatibility

Additive and non-breaking — `--timezone` is new and existing
`--start-timezone`/`--end-timezone` usage is unchanged. Kong scopes the `--tz`
alias per command, so there's no collision with `create-calendar`.

No `CHANGELOG.md` entry — left to the release/landing process per repo convention.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Calendar: add `create --timezone`/`--tz` to apply one IANA timezone to both event endpoints while retaining granular start/end timezone flags. (#844) — thanks @malob.
 - Docs: add non-destructive `insert-image --before` and `--after` anchors while clarifying that `--at` replaces its placeholder. (#839) — thanks @sebsnyk.
 - Slides: allow `insert-image` and `replace-slide` to use public HTTPS image URLs without temporary Drive sharing. (#825) — thanks @sebsnyk.
 - Slides: add structured element geometry, styled text runs, table-cell content, image source URLs, native presentation metadata, and read-only text location with exact UTF-16 ranges. (#822) — thanks @sebsnyk.
--- a/docs/commands/gog-calendar-create.md
+++ b/docs/commands/gog-calendar-create.md
@@ -66,6 +66,7 @@ gog calendar (cal) create (add,new) <calendarId> [flags]
 | `--source-url` | `string` |  | URL where event was created/imported from |
 | `--start-timezone`<br>`--from-timezone` | `string` |  | IANA timezone metadata for --from (e.g., Europe/Rome) |
 | `--summary` | `string` |  | Event summary/title |
+| `--timezone`<br>`--tz` | `string` |  | IANA timezone metadata applied to both --from and --to (e.g., America/Los_Angeles); mutually exclusive with --start-timezone/--end-timezone |
 | `--to` | `string` |  | End time (RFC3339) |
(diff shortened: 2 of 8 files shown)


malob:
@clawsweeper re-review

steipete:
Landed after maintainer review and exact-head validation.

- Local: focused timezone tests and full `make ci` passed.
- Review: autoreview clean; no accepted/actionable findings.
- Live: reviewed the recorded create/read/delete proof showing both event endpoints persisted `Europe/London`.
- CI: exact-head CI and Docker runs passed on Linux, Windows, macOS/cgo, and the worker job.
- Caveat: maintainer live rerun could not access this machine's encrypted file keyring noninteractively, so the recorded live proof remains the API-level evidence.

Issue #221, opened 2026-02-11
gog hangs indefinitely after OAuth client creation on macOS 26.2 (Tahoe)


xiaoheiclaw:
## Description

All `gog` API commands (gmail search, calendar list, etc.) hang indefinitely after successfully creating the OAuth client. The process never returns and must be killed.

## Environment

- **gog version**: v0.9.0 (99d9575 2026-01-22)
- **OS**: macOS 26.2 (Apple Silicon, arm64)
- **Go version**: (bundled in binary)
- **Account**: Gmail/Google Workspace via OAuth

## Steps to Reproduce

1. `gog auth add user@gmail.com` — completes successfully
2. `gog auth list` — shows the account with valid token
3. `gog gmail search "is:unread" --max 1 --verbose` — **hangs forever**

## Verbose Output

```
time=2026-02-11T10:10:02.144+08:00 level=DEBUG msg="creating client options with custom scopes" serviceLabel=gmail email=user@gmail.com
time=2026-02-11T10:10:02.146+08:00 level=DEBUG msg="client options with custom scopes created successfully" serviceLabel=gmail email=user@gmail.com
```

After this, the process hangs indefinitely. No further output, no error.

## What I've Tried

- `keyring_backend: file` with `GOG_KEYRING_PASSWORD` env var — same hang
- `keyring_backend: keychain` — same hang
- With and without proxy (`http://127.0.0.1:7890`) — same hang
- `GODEBUG=http2client=0` to disable HTTP/2 — same hang
- `--no-input` flag — same hang
- Direct `curl` to Google APIs via same proxy works fine (gets 401 as expected)
- Tested with `gmail search`, `calendar list`, `calendar events` — all hang

## Expected Behavior

The command should return results or an error within a reasonable timeout.

## Notes

- `gog auth list` and `gog auth credentials` work fine (no API call needed)
- The hang happens at the actual Google API HTTP request stage
- This is on macOS 26.2 (Tahoe) — could be a TLS/HTTP compatibility issue with the Go version bundled in the binary

salmonumbrella:
Correction: included in PR https://github.com/steipete/gogcli/pull/228.\n\nThe change uses a clone of default Go HTTP transport settings so proxy environment variables and standard transport defaults are preserved, which addresses the post-auth hanging behavior reported here.

steipete:
Fixed on current `main`.

[...]

PR #845, opened 2026-06-19
feat(secrets): make keyring open timeout configurable, default 30s

malob:
## Summary

The keyring open/operation timeout — added so an unresponsive backend can't hang
the CLI forever (#513, #221) — defaults to 10s. This raises that default to 30s
and makes it configurable via `GOG_KEYRING_OPEN_TIMEOUT` (a Go duration),
mirroring the existing `GOG_KEYRING_LOCK_TIMEOUT`.

- `GOG_KEYRING_OPEN_TIMEOUT=45s` (any Go duration) overrides the timeout;
  empty/invalid/non-positive values fall back to the default.
- Default raised 10s → 30s.
- The timeout error now points at `GOG_KEYRING_OPEN_TIMEOUT`.

## Why

On macOS this timeout bounds every Keychain read, including the wait for an
interactive permission prompt. 10s is usually enough to approve one — but it
leaves little margin for the normal frictions: not noticing the prompt for a few
seconds, fumbling and re-entering the login password, or macOS showing two
prompts back to back (client secret + token). When that margin runs out, a
legitimate prompt times out before it's approved and auth fails on a perfectly
healthy machine. 30s gives comfortable headroom for those cases while still
bounding the failure the timeout was created for — an *indefinite* hang
(commands hung until SIGKILL, #513), not a matter of seconds.
`GOG_KEYRING_OPEN_TIMEOUT` then lets anyone tune it further, just like
`GOG_KEYRING_LOCK_TIMEOUT` already does for the lock wait.

If raising the default is contentious, the env var alone is still a worthwhile
addition (parity with the lock-timeout knob) — but I'd argue 30s should be the
default: 10s is too tight for the common interactive-Keychain path, and 30s still
firmly bounds the hang case.

## Behavior proof

With the Keychain permission prompt left unanswered (so the read blocks until the
timeout fires), wall-clock time matches the configured timeout:

```console
# default — read gives up after 30s
$ time ./bin/gog calendar calendars --account <redacted> --json
... keyring connection timed out after 30s while reading keyring item (...);
    set GOG_KEYRING_OPEN_TIMEOUT to allow more time, or set GOG_KEYRING_BACKEND=file ...
# ≈ 30s elapsed

# override — the same read gives up after the configured 5s
$ time GOG_KEYRING_OPEN_TIMEOUT=5s ./bin/gog calendar calendars --account <redacted> --json
... keyring connection timed out after 5s while reading keyring item (...);
    set GOG_KEYRING_OPEN_TIMEOUT to allow more time, or set GOG_KEYRING_BACKEND=file ...
# ≈ 5s elapsed
```

## Testing

- `make fmt-check`, `make lint`, `make docs-check` clean.
- `go test ./internal/secrets -run 'OpenTimeout|ParseKeyringOpenTimeout|TimeoutKeyring'` —
  covers env parsing (valid/empty/invalid/non-positive), the `OpenOptionsFromLookup`
  wiring, and that the timeout error names `GOG_KEYRING_OPEN_TIMEOUT`.
- Documented the env var in `docs/spec.md`.

No `CHANGELOG.md` entry — left to the release/landing process per repo convention.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Auth: make keyring open/operation timeouts configurable with `GOG_KEYRING_OPEN_TIMEOUT`, using a `30s` macOS default for permission prompts while retaining `10s` elsewhere. (#845) — thanks @malob.
 - Calendar: add `create --timezone`/`--tz` to apply one IANA timezone to both event endpoints while retaining granular start/end timezone flags. (#844) — thanks @malob.
 - Docs: add non-destructive `insert-image --before` and `--after` anchors while clarifying that `--at` replaces its placeholder. (#839) — thanks @sebsnyk.
 - Slides: allow `insert-image` and `replace-slide` to use public HTTPS image URLs without temporary Drive sharing. (#825) — thanks @sebsnyk.
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -112,7 +112,7 @@ Implementation: `internal/config/*`.
 - Legacy key format: `token:<email>` (migrated on first read)
 - Stored payload is JSON (refresh token + metadata like OIDC subject, current email, selected services/scopes).
 - Email is treated as display/contact state; Google's OIDC `sub` is used to detect the same account after an email rename and migrate aliases/defaults/client mappings on reauthorization.
-- macOS Keychain operations are bounded by a timeout (default `30s`, configurable via `GOG_KEYRING_OPEN_TIMEOUT`) so non-surfacing permission prompts return actionable guidance instead of hanging indefinitely.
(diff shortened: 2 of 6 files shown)


steipete:
Landed with the maintainer policy adjustment: `30s` by default on macOS, `10s` elsewhere, and `GOG_KEYRING_OPEN_TIMEOUT` authoritative on every platform.

- Local: focused `internal/secrets` tests and full `make ci` passed.
- Review: autoreview clean; no accepted/actionable findings.
- E2E: reviewed the recorded macOS Keychain prompt proof for both the default and explicit override; this machine's encrypted file-keyring access was unavailable for a second live prompt run.
- CI: exact-head CI and Docker runs passed.

PR #849, opened 2026-06-19
feat(sheets): verify formula errors structurally

alexknowshtml:
## Summary

- use the existing `--values-json @file` / `@-` path for shell-safe formula input
- add `--fail-on-formula-error` to read back the exact updated range as typed grid data
- fail on real Sheets formula errors while preserving RAW literal strings such as `#REF!`
- return structured `formulaErrors` entries with cell, type, and message
- document the workflow and extend disposable live coverage

This replaces the proposed `--safe-formula` input path with the command's canonical file/stdin input and replaces displayed-text matching with `effectiveValue.errorValue` inspection.

## Verification

- `make ci`
- autoreview: clean against `main`
- focused tests for file input, typed errors, RAW literals, and readback failures
- disposable live Google Sheets run covering a valid file formula, RAW literal error text, and a failing divide-by-zero formula

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Sheets: add structured formula-error verification to `sheets update --fail-on-formula-error`, using exact updated-range grid data and canonical `--values-json @file` input. (#849) — thanks @alexknowshtml.
 - Docs: add first-class footnote, section-break, horizontal-rule, and section-column commands with shared index, anchor, tab, dry-run, and batch behavior where supported. (#856) — thanks @sebsnyk.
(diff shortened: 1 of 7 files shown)


steipete:
Landed in 4d3862ae.

I kept the core goal and rewrote the implementation around the existing `--values-json @file` / `@-` input path. Formula verification now reads typed `effectiveValue.errorValue` data for the exact updated range, returns cell/type/message details, fails on readback errors, and does not misclassify RAW literal error-looking strings. Docs, focused regressions, disposable live Sheets proof, local CI, autoreview, and all hosted jobs are green.

Thanks @alexknowshtml for the contribution and for identifying the shell-safe formula and verification gap.

PR #854, opened 2026-06-19
feat(docs insert): add --markdown for positional formatted-markdown insertion

### LINKED ISSUE #851: feat(docs insert): add --markdown for positional formatted-markdown insertion (parity with docs update --markdown)
## Summary

`gog docs insert` places content at an arbitrary position (`--index`, `--at <text>`, `--occurrence`), but it is **plain text only** — there is no `--markdown` flag, so a heading, fenced code block, or list inserted via `insert` lands as literal unformatted text.

The converted-markdown path that `insert` is missing already exists in the codebase and is already exposed on a sibling command. Adding `--markdown` to `insert` is a small, consistent extension rather than new machinery.

## Current behavior (verified against current `main`, build `0.29.0`)

- `docs insert` has flags `--index` / `--at` / `--occurrence` / `--match-case` / `--tab` but no `--markdown`. Its `Run` emits a single plain `InsertTextRequest`:
  - `internal/cmd/docs_edit.go:883` — `DocsInsertCmd` struct (no `Markdown` field).
  - `internal/cmd/docs_edit.go:966` — builds the request via `docsedit.BuildInsertRequest(content, insertIndex, c.Tab)`.
  - `internal/docsedit/planner.go:71` — `BuildInsertRequest` is a bare `InsertText` request, no styling.

- `docs write --markdown` converts markdown but only at whole-tab (`--replace`) or end-of-doc (`--append`) granularity; it rejects positional use:
  - `internal/docsedit/markdown.go:114` — returns `"--markdown requires --replace or --append"`.

## The capability already exists — just not on `insert`

`docs update --markdown` **already** converts markdown and places it at an arbitrary position:

- `internal/cmd/docs_edit.go:761` — the non-replacing markdown branch of `DocsUpdateCmd.Run` calls `insertPreparedDocsMarkdownAt(ctx, svc, id, insertIndex, markdown, c.Tab, true)` and then rewrites heading links over the inserted range.
- `internal/cmd/docs_mutation.go:266` — `insertPreparedDocsMarkdownAt` parses the markdown, calls the converter at the supplied index, submits the batchUpdate, and handles tables/images.
- `internal/docsmarkdown/requests.go:33` — `MarkdownToDocsRequests(elements, baseIndex, tabID)` is already parameterized on an arbitrary `baseIndex`; nothing pins it to index 1 or the document end. Fenced code blocks get their styling (`Roboto Mono` + grey shading) inside the same converter (`internal/docsmarkdown/requests.go:77`), so a `--markdown` insert would render code blocks identically to `write --markdown`.

So positional formatted-markdown insertion is feasible and already implemented internally; the only thing missing is the flag on the command users reach for first.

## Why this is worth fixing

`insert` is the obviously-named command for "put this content here". A user who wants to drop a formatted markdown block mid-document will try `docs insert --markdown` first, get an unknown-flag error, and then has to discover that the converted path lives on `update --markdown` instead. That is a discoverability cliff for a capability the tool already has. The two commands diverge for no functional reason: `update` can do plain-or-markdown at a position; `insert` can only do plain.

## Proposed surface

Add `--markdown` to `docs insert`:

```
gog docs insert <docId> --markdown --at "## Section" --file block.md
gog docs insert <docId> --markdown --index 1 --text $'## Heading\n\n```\ncode\n```\n'
```

Behavior:

- Without `--markdown`: unchanged (plain `InsertText`).
- With `--markdown`: route through the existing `insertPreparedDocsMarkdownAt` helper at the resolved index, then run the existing heading-link rewrite over the inserted range — i.e. mirror the non-replacing branch of `docs update --markdown` (`internal/cmd/docs_edit.go:761`).
- `--at` / `--index` / `--occurrence` / `--tab` semantics are unchanged; they just resolve the index that the converted block is inserted at.

This keeps `insert` = "place at a position, never delete", and leaves `update --markdown --replace-range` / `update --markdown --at` as the delete-and-replace path. The split between the two commands stays clean.

## API feasibility

Not an API limitation. `documents.batchUpdate` accepts `InsertText` plus paragraph/text-style requests at any index, which is exactly what `MarkdownToDocsRequests` already emits relative to `baseIndex`. The same converter already runs at index 1 (replace), at the end index (append), and at an arbitrary index (`update --markdown`).

## Acceptance criteria

- `docs insert --markdown` converts markdown (headings, fenced code blocks with the standard mono + grey styling, lists, inline styles, tables/images via the existing helper) and inserts it at the index resolved from `--index` / `--at` / `--occurrence`.
- Output without `--markdown` is byte-for-byte unchanged.
- `--tab` targeting works for the markdown path (the helper already threads `tabID`).
- A unit test covering `docs insert --markdown --index N` asserting the converter requests (heading paragraph-style + code-fence text-style) land at the resolved index, alongside the existing plain-text insert tests.

sebsnyk:
## What

Adds `--markdown` to `gog docs insert`, so a formatted markdown block (heading, fenced code block, list, table, image, inline styles) can be inserted at a position resolved by `--index` / `--at` / `--occurrence`, without deleting anything at the anchor.

Closes #851.

## Why

`docs insert` already places content at an arbitrary position, but only as plain text — a heading or fenced code block inserted this way landed as literal, unformatted characters. The converted-markdown path already exists in the codebase and is already exposed on `docs update --markdown` (its non-replacing branch). So `insert` was the only positional command missing a capability the tool already has, which is a discoverability cliff: users reach for `docs insert --markdown` first, hit an unknown-flag error, and then have to learn that the converted path lives on `update` instead.

## How

- New `--markdown` bool on `DocsInsertCmd`.
- When set, after the placement index is resolved, route through the existing `insertPreparedDocsMarkdownAt` helper and run the existing heading-link rewrite over the inserted range — mirroring the non-replacing branch of `docs update --markdown`. The shared converter (`docsmarkdown.MarkdownToDocsRequests`) is already parameterized on an arbitrary base index, so code-fence styling (Roboto Mono + grey shading) and everything else render identically to `write --markdown`.
- `--markdown` is rejected together with `--batch`, since the markdown path issues its own `documents.batchUpdate` calls and cannot be queued into a persisted batch.
- Plain (non-`--markdown`) behavior is unchanged.

The split between the two commands stays clean: `insert` = place at a position, never delete; `update --markdown --at` / `--replace-range` = delete-and-replace.

## Tests

- `TestDocsInsertCmd_MarkdownAtIndex` — asserts a converted heading inserts at the resolved index with the converter's HEADING paragraph-style request (and that an explicit `--index` with no heading links issues no extra GET).
- `TestDocsInsertCmd_MarkdownRejectsBatch` — asserts the `--markdown` + `--batch` guard.
- Existing plain-insert tests unchanged and passing.

`go build ./...`, `go vet ./internal/cmd/`, and `go test ./internal/cmd/` all pass. CHANGELOG and the generated `docs insert` command page + editing guide are updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 0.29.1 - Unreleased
 
+### Added
+
+- Docs: add `insert --markdown` to convert markdown to Google Docs formatting and place the converted block at a position resolved by `--index`/`--at`/`--occurrence`, giving `insert` parity with the existing `update --markdown` and `write --markdown` paths. (#851)
+
 ## 0.29.0 - 2026-06-19
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Added
 
-- Docs: add `insert --markdown` to convert markdown to Google Docs formatting and place the converted block at a position resolved by `--index`/`--at`/`--occurrence`, giving `insert` parity with the existing `update --markdown` and `write --markdown` paths. (#851)
+- Docs: add `insert --markdown` to convert markdown to Google Docs formatting and place the converted block at a position resolved by `--index`/`--at`/`--occurrence`, giving `insert` parity with the existing `update --markdown` and `write --markdown` paths. (#851, #854) — thanks @sebsnyk.
 
 ## 0.29.0 - 2026-06-19
 
--- a/docs/commands/gog-docs-insert.md
+++ b/docs/commands/gog-docs-insert.md
@@ -35,6 +35,7 @@ gog docs (doc) insert <docId> [<content>] [flags]
 | `--home` | `string` |  | Override gogcli config/data/state/cache root (equivalent to GOG_HOME) |
 | `--index` | `*int64` |  | Character index to insert at (1 = beginning). Defaults to end-of-doc when omitted. |
 | `-j`<br>`--json`<br>`--machine` | `bool` | false | Output JSON to stdout (best for scripting) |
+| `--markdown` | `bool` |  | Convert markdown to Google Docs formatting before inserting |
 | `--match-case` | `bool` |  | Use case-sensitive --at matching |
 | `--no-input`<br>`--non-interactive`<br>`--noninteractive` | `bool` |  | Never prompt; fail instead (useful for CI) |
 | `--occurrence` | `*int` |  | Use the Nth --at match (1-based; required when --at is ambiguous) |
--- a/docs/docs-editing.md
+++ b/docs/docs-editing.md
@@ -111,6 +111,15 @@ Use `--occurrence N` when an anchor is ambiguous and `--match-case` when case
 must be exact. `docs comments locate` applies the same matching rules to a
 comment's quoted text and reports its tab plus UTF-16 range.
 
+`insert` and `update` both accept `--markdown` to convert the content (headings,
+fenced code blocks, lists, tables, images) before placing it at the resolved
+position. `insert --markdown` adds the block without deleting anything; `update
+--markdown` with `--at`/`--replace-range` replaces the matched range.
+
+```bash
+gog docs insert <docId> --markdown --at "## Section" --file block.md
+```
+
 Command pages:
 
 - [`gog docs find-range`](commands/gog-docs-find-range.md)
--- a/internal/cmd/docs_edit.go
+++ b/internal/cmd/docs_edit.go
@@ -969,7 +969,7 @@ func (c *DocsInsertCmd) Run(ctx context.Context, kctx *kong.Context, flags *Root
 	c.Tab = resolvedPlacement.TabID
 
 	if c.Markdown {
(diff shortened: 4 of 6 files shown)


steipete:
Landed as ab9a331d4c0773d715c15222bd12d5192db17c71.

Verification:
- `go test ./internal/cmd ./internal/docsedit ./internal/docsmarkdown`
- `make ci`
- Codex autoreview on the final branch diff: clean, no actionable findings.
- Exact-head live Google Docs proof: created a disposable document, inserted a literal anchor, then ran `docs insert --markdown --at ANCHOR` with a heading, fenced code block, and list. `docs raw` readback showed `HEADING_2`, `Roboto Mono` with the expected code color, native bullet list IDs, and the original anchor still present after the inserted block. The disposable document was moved to trash afterward.
- Required CI at head f0a0acef796cf73924326f61f7ce1d0ce4d85ef0: test, image, worker, Windows, and Darwin CGO all passed.

Caveat: live proof covered anchor placement plus heading/code/list formatting; table, image, and tab variants continue through the pre-existing shared Markdown insertion helper and were covered by existing/unit paths rather than separate live mutations here.

PR #858, opened 2026-06-19
feat(docs): add layout-preserving image replacement

### LINKED ISSUE #853: feat(docs): replace-image command (replaceImage) — swap image bytes in place without losing position/size
The Docs API `replaceImage` request swaps the bytes behind an existing inline image while preserving its position, size, and surrounding paragraphs — only the pixels change. `gog docs` has no command for it. Today the only way to refresh a diagram is `delete the positioned object + insert-image` again, which loses the original placement and re-applies a default width, so every refresh needs manual re-positioning.

This is a common workflow: a doc embeds a generated diagram, the source regenerates, and you want to drop in new bytes without disturbing layout.

Proposed:

```
gog docs replace-image <docId> --file <local.png>          # upload, then ReplaceImageRequest
gog docs replace-image <docId> --url <https-url>
  --match-alt <substring>   # pick when the doc has multiple inline images (match on alt/description)
  --object-id <id>          # explicit positioned-object id (from `gog docs images list`)
  --tab <tab>
```

`images list` already enumerates the inline objects, so an `--object-id` selector falls out naturally; `--match-alt` is the ergonomic path. The upload half can reuse whatever `insert-image --file` already does for restricted-tenant embedding.

Ref: documents.batchUpdate `ReplaceImageRequest` (imageObjectId, uri, imageReplaceMethod).

steipete:
## Summary

- add `docs replace-image` with exact object-ID, case-insensitive alt substring, or implicit single-image selection
- support public HTTPS URLs and local PNG/JPEG/GIF uploads with temporary Drive sharing cleanup
- preserve object position and rendered bounds through `ReplaceImageRequest` with `CENTER_CROP`
- retain tab targeting and document revision control
- document the command and add changelog credit for @sebsnyk

Closes #853.

## Verification

- `go test ./internal/cmd -run 'Docs.*Image|DocsImages|DocsSed'`
- `go test ./internal/docssed ./internal/docsedit`
- `make ci`
- Codex autoreview: clean, no actionable findings
- live Google Docs proof on exact branch head: inserted a public Docs logo, replaced it by exact object ID with a Sheets logo, verified the same object ID and 72×72 PT bounds plus the changed source URI, then replaced it again through implicit single-image selection; disposable document moved to trash

The local-file path has full mocked Drive upload, public-share, Docs replacement, and permission-revoke coverage. No additional external image upload was performed during live proof.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Docs: add `replace-image` with exact object-ID, alt-text, single-image, tab, public-URL, and local-file targeting while preserving the original image position and bounds. (#853) — thanks @sebsnyk.
 - Docs: add `insert --markdown` to convert markdown to Google Docs formatting and place the converted block at a position resolved by `--index`/`--at`/`--occurrence`, giving `insert` parity with the existing `update --markdown` and `write --markdown` paths. (#851, #854) — thanks @sebsnyk.
 
 ## 0.29.0 - 2026-06-19
--- a/docs/commands.generated.md
+++ b/docs/commands.generated.md
@@ -270,6 +270,7 @@ Generated from `gog schema --json`.
       - [`gog docs (doc) paragraphs list (ls) <docId> [flags]`](commands/gog-docs-paragraphs-list.md) - List paragraphs
     - [`gog docs (doc) raw <docId> [flags]`](commands/gog-docs-raw.md) - Dump raw Google Docs API response as JSON (Documents.Get; lossless; for scripting and LLM consumption)
     - [`gog docs (doc) rename-tab <docId> [flags]`](commands/gog-docs-rename-tab.md) - Rename a tab in a Google Doc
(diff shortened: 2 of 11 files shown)


steipete:
Landed as 91c0bcb105809589529693d7f499711b9a93255b.

Verification:
- `go test ./internal/cmd -run 'Docs.*Image|DocsImages|DocsSed'`
- `go test ./internal/docssed ./internal/docsedit`
- `make ci`
- Codex autoreview on local and final branch diffs: clean, no actionable findings.
- Exact-head live Google Docs proof: inserted a public Docs logo, replaced it by exact object ID with a Sheets logo, and confirmed through `docs images list` plus `docs raw` that the object ID and 72×72 PT bounds stayed unchanged while the source URI changed. Replaced it again with no selector to prove implicit single-image selection. The disposable document was moved to trash.
- Exact-head GitHub checks: both push and pull-request test, Windows, worker, and Darwin CGO jobs passed; Docker image and Socket checks also passed.

Caveat: the local-file path was not externally uploaded during live proof. Its Drive upload, temporary public permission, Docs replacement, and permission revocation sequence is covered end-to-end with mocked Drive and Docs services.

PR #875, opened 2026-06-23
feat(calendar): add 'changed' command to list recently modified events

sorenisanerd:
Adds `gog calendar changed` (aliases: recent, modified) which surfaces the N events most recently created or modified, sorted by last-modification time descending. Uses Events.List with UpdatedMin and omits SingleEvents so that recurring series appear as a single master entry rather than exploding into per-instance results. Supports --since (duration/date/RFC3339, default 30d), --max, --all, --cal, --calendars, --show-deleted, --weekday, --location, and --fail-empty flags.

Screenshot-2026-06-23_16:17:31


sorenisanerd:
I think I prefer the current behavior. Cancelling a *single instance* of a recurring event will show up show up regardless of `--show-deleted`. A deleted recurring event will only show up if `--show-deleted` is passed.

steipete:
Thanks @sorenisanerd! Landed in https://github.com/openclaw/gogcli/commit/e022bb984ac5f805b5ca469dcd83f4e8612d59cb with the changed-event command narrowed to the documented surface and cancellations included. Live disposable proof covered create, update, and cancelled-event readback; cleanup succeeded.

PR #877, opened 2026-06-23
fix(secrets): prompt re-auth on a corrupt stored token instead of failing fatally

### LINKED ISSUE #872: [Bug]: corrupt stored token fails fatally on every command instead of prompting re-auth
### Bug type

Error handling / auth recovery — a **corrupt (but present) stored token** is treated as a fatal error on every command instead of degrading to a re-authentication prompt.

### Summary

`getTokenNoLockOptions` returns a generic `decode token: …` error when a stored token entry is present but not decodable, and the API-client auth path (`buildTokenSource`) only recognizes `keyring.ErrKeyNotFound` as the "needs re-auth" signal. As a result, an **absent** token correctly maps to `AuthRequiredError` (re-auth), but a **corrupt** token falls through to a hard error — so the CLI fails on *every* command for that account with no actionable recovery, even though re-authenticating would fix it.

### Where (code pointers, current `main`)

- `internal/secrets/token.go:188-191` — on `json.Unmarshal` failure the function returns a plain wrapped error:
  ```go
  var st storedToken
  if err := json.Unmarshal(item.Data, &st); err != nil {
      return Token{}, fmt.Errorf("decode token: %w", err)
  }
  ```
- `internal/googleapi/client_auth.go:238-246` — the caller degrades to re-auth **only** for `ErrKeyNotFound`; any other error (including the corrupt-decode error above) becomes fatal:
  ```go
  if t, err := store.GetToken(client, email); err != nil {
      if errors.Is(err, keyring.ErrKeyNotFound) {
          return nil, &AuthRequiredError{...}   // absent token -> re-auth
      }
      return nil, fmt.Errorf("get token for %s: %w", email, err) // corrupt token -> fatal
  }
  ```

### Reproduction (verified)

The file backend is a first-class, documented configuration (`GOG_KEYRING_BACKEND=file` / `gog auth keyring file`), so a partially-written / corrupted token entry is a real on-disk failure mode (interrupted write, disk issue, or a stored-shape change across versions). The asymmetry is reproducible at the unit level with no binary build, using the same in-memory keyring the existing `TestKeyringStore_TokenRoundTrip` uses:

```go
func TestGetToken_CorruptEntry_NotClassifiableAsReauth(t *testing.T) {
	ring := keyring.NewArrayKeyring(nil)
	store := &KeyringStore{ring: ring}
	client := config.DefaultClientName
	email := normalize("a@b.com")

	// Present-but-corrupt token blob (e.g. interrupted write).
	_ = ring.Set(keyring.Item{Key: tokenKey(client, email), Data: []byte("{not-json")})

	_, err := store.GetTokenNoMigrate(client, email)
	// err = "decode token: invalid character 'n' looking for beginning of object key string"
	if errors.Is(err, keyring.ErrKeyNotFound) { t.Fatal("would re-auth") }
	// errors.Is(err, ErrKeyNotFound) == false  => caller treats as FATAL, not re-auth (the bug)

	// Control: an ABSENT token *does* classify as ErrKeyNotFound -> re-auth.
	_, errAbsent := store.GetTokenNoMigrate(client, normalize("absent@b.com"))
	// errors.Is(errAbsent, keyring.ErrKeyNotFound) == true  (correct)
}
```

Observed output (current `main`):
```
corrupt-token error: decode token: invalid character 'n' looking for beginning of object key string
errors.Is(err, ErrKeyNotFound) = false  => caller treats as FATAL, not re-auth (the bug)
control: absent token -> ErrKeyNotFound -> re-auth (correct)
```

### Expected vs actual

- **Expected:** a present-but-undecodable token is unusable and equivalent to "no usable credential" → surface `AuthRequiredError` so the user re-authenticates (same recovery as an absent token).
- **Actual:** every command for that account fails with `get token for <email>: decode token: …` and no re-auth path; the user must manually find and delete the keyring entry.

### Prior art in this org (same bug-class)

`openclaw/mcporter` had the identical class (corrupt credential cache crashed instead of degrading) — filed as #207, fixed in #208 by degrading a corrupt **token/client** cache to re-auth **while keeping security-gating material (OAuth `state`) fail-closed**. That distinction matters here too (see fix options).

### Proposed fix (options)

- **Option A (preferred, matches mcporter #208):** introduce a sentinel (e.g. `ErrCorruptToken`) returned specifically from the `json.Unmarshal` failure in `getTokenNoLockOptions`, and have `buildTokenSource` treat `errors.Is(err, ErrCorruptToken)` like `ErrKeyNotFound` → `AuthRequiredError`. This targets **only** the corrupt-decode case.
- **Option B:** same as A but also delete/quarantine the corrupt entry on read so the next auth flow starts clean.
- **Option C (not recommended):** broadly map *all* non-`ErrKeyNotFound` `GetToken` errors to re-auth. **Avoid** — this would mask genuinely transient/operational faults the code intentionally surfaces today (macOS Keychain locked in headless env, permission errors, wrong file-backend password — see the read-back guidance in `internal/secrets/store.go`). The fix must distinguish "credential is corrupt/unusable" from "the store is temporarily unavailable."

Happy to follow up with a PR (Option A) including the test above plus a caller-level assertion if a maintainer confirms the direction.

---
*Filed with AI assistance (static analysis + a sandbox-run repro). The reproduction above was executed against current `main`; the production-frequency of corrupt entries is environment-dependent and not directly observed.*

KrasimirKralev:
## Problem

A present-but-undecodable token entry (interrupted write, stored-shape change across versions) made every command fail with a fatal, non-actionable `get token for <email>: decode token: …`. An **absent** token correctly mapped to `AuthRequiredError` → re-auth; a **corrupt** one didn't, even though the recovery is identical: re-authenticate. The user had no way out except manually deleting the keyring entry.

Closes #872.

## Fix

**Option A from the issue.** Introduce an exported sentinel `secrets.ErrCorruptToken`, wrap the `json.Unmarshal` failure with it (both decode sites in `getTokenNoLockOptions` — primary key path and subject-key path), and widen the classifier at the call site:

```go
if errors.Is(err, keyring.ErrKeyNotFound) || errors.Is(err, secrets.ErrCorruptToken) {
    return nil, &AuthRequiredError{Service: serviceLabel, Email: email, Client: client, Cause: err}
}
```

The narrower scope is deliberate. Mapping *all* non-`ErrKeyNotFound` errors to re-auth (Option C in the issue) would mask genuinely transient / operational store faults the code intentionally surfaces today: macOS Keychain locked in a headless env, file-backend wrong password, EACCES, etc. Only the **decode-corruption** case — which is semantically equivalent to "no usable credential" — is widened.

Same pattern as [`openclaw/mcporter#208`](https://github.com/openclaw/mcporter/pull/208) (corrupt cred cache → degrade to re-auth; keep security-gating material fail-closed).

## Verification

Go 1.26.2:

- New `internal/secrets/token_corrupt_test.go` seeds an in-memory keyring with `{not-json` under the primary token key and asserts:
  - `errors.Is(err, ErrCorruptToken) == true`
  - `errors.Is(err, ErrKeyNotFound) == false`
  - Control: an absent entry still classifies as `ErrKeyNotFound` (unchanged).
- **Negative control:** with the `ErrCorruptToken` wrap reverted on `token.go`, the new test fails with the exact pre-fix message — `decode token: invalid character 'n' looking for beginning of object key string`. With the fix in place, it passes.
- `go test ./internal/secrets/... ./internal/googleapi/...` — both suites pass, no regressions.
- `gofmt -l` clean, `go vet` clean.

### DIFF
--- a/internal/googleapi/client_auth.go
+++ b/internal/googleapi/client_auth.go
@@ -236,7 +236,10 @@ func tokenSourceForAccountScopesWithStoredScopeCheck(
 	var tok secrets.Token
 
 	if t, err := store.GetToken(client, email); err != nil {
-		if errors.Is(err, keyring.ErrKeyNotFound) {
+		// An absent or corrupt stored token is equivalent to "no usable credential":
+		// surface AuthRequiredError so the caller can prompt re-auth, instead of
+		// failing every command with a non-actionable decode error.
+		if errors.Is(err, keyring.ErrKeyNotFound) || errors.Is(err, secrets.ErrCorruptToken) {
 			return nil, &AuthRequiredError{Service: serviceLabel, Email: email, Client: client, Cause: err}
 		}
 
--- a/internal/secrets/token.go
+++ b/internal/secrets/token.go
@@ -27,6 +27,10 @@ type Token struct {
 var (
 	errMissingEmail        = errors.New("missing email")
 	errMissingRefreshToken = errors.New("missing refresh token")
+	// ErrCorruptToken signals a token entry is present in the keyring but cannot
+	// be decoded (e.g. interrupted write, stored-shape change). Callers should
+	// treat it equivalently to a missing token and prompt re-authentication.
+	ErrCorruptToken = errors.New("stored token is corrupt")
 )
 
 type storedToken struct {
@@ -187,7 +191,7 @@ func (s *KeyringStore) getTokenNoLockOptions(client string, email string, migrat
 
 	var st storedToken
 	if err := json.Unmarshal(item.Data, &st); err != nil {
-		return Token{}, fmt.Errorf("decode token: %w", err)
+		return Token{}, fmt.Errorf("decode token: %w: %w", ErrCorruptToken, err)
 	}
 
 	return Token{
@@ -432,7 +436,7 @@ func (s *KeyringStore) getTokenBySubjectNoLock(client string, subject string) (T
 
 	var st storedToken
 	if err := json.Unmarshal(item.Data, &st); err != nil {
-		return Token{}, fmt.Errorf("decode token: %w", err)
+		return Token{}, fmt.Errorf("decode token: %w: %w", ErrCorruptToken, err)
 	}
 
 	return Token{


steipete:
Thanks for the detailed report and follow-up. The narrow corrupt-token sentinel fix landed in https://github.com/openclaw/gogcli/commit/779554365e63afdfb144ff93d0cd380a524a695e via #874, including secrets-level and caller-level regression coverage. Closing this duplicate in favor of the tested landed version.

PR #878, opened 2026-06-23
feat(docs): honor --tab when setting page layout

atmasphere:
## What

Add `--tab` support to `docs page-layout` so document-layout changes can target a specific tab, and fix `docs write --pageless` silently no-opping on secondary tabs of multi-tab documents.

## Why

Document-style fields (`documentMode`/pageless, page size, margins) are **per-tab** in the Docs API. `page-layout` and `write --pageless` built an `UpdateDocumentStyleRequest` but never set `TabId`, so the change always applied to the document's default tab. On a multi-tab document, setting pageless on a non-default tab reported success while nothing actually changed — `documentFormat.documentMode` stayed `PAGES` on the targeted tab.

## Changes

- `docs page-layout`: add `--tab` (title or ID, with the existing deprecated `--tab-id` alias), resolved to a concrete tab ID via `resolveDocsTabID`.
- `docs_layout.go`: thread `TabID` through `docsDocumentStyleOptions` onto `UpdateDocumentStyleRequest.TabId`.
- `docs write`: resolve `--tab` in `applyDocumentStyle` so `--pageless`/layout flags land on the targeted tab across every write path, including the markdown/Drive-replace path that never resolved the tab.
- Behaviour is unchanged when `--tab` is omitted (empty `TabId` → default tab).

## Tests

- New unit tests for the request builder (`TabID` set → request carries it; omitted → empty, default-tab behaviour preserved).
- `make fmt-check lint deadcode test` all pass locally; regenerated `docs/commands/gog-docs-page-layout.md` and updated `CHANGELOG.md`.
- Verified end-to-end against the live Docs API: `page-layout --layout pageless --tab <id>` flips `documentFormat.documentMode` to `PAGELESS` on the targeted secondary tab (confirmed by reading that tab back).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.30.1 - Unreleased
 
+- Docs: honor `--tab` when setting document layout so `page-layout --tab` (and `write --pageless --tab`) target the specified tab instead of always the default tab. Page layout is per-tab; previously these silently no-opped on secondary tabs of multi-tab documents. (#878) — thanks @atmasphere.
 - Auth: recover from corrupt stored OAuth token payloads by routing only classified decode corruption through the normal re-authentication flow while preserving operational keyring errors. (#872, #874) — thanks @KrasimirKralev.
 - Calendar: add repeatable or comma-separated `events --event-types` filtering across single, selected, and all-calendar listings while preserving the existing unfiltered default. (#870) — thanks @malob.
 - Gmail: render quoted-reply and forwarded-message dates in Gmail's human-readable style using the configured account timezone. (#873) — thanks @malob.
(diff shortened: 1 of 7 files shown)


steipete:
Landed as [57e3283a](https://github.com/openclaw/gogcli/commit/57e3283a520a4aa88d8d7591fc34cf9e6189f21b).

Validation:
- `go test ./internal/cmd -run 'TestDocsPageLayout|TestBuildUpdateDocumentStyleRequest'`
- `make ci`
- Codex autoreview: clean; no accepted/actionable findings
- PR CI: Linux test/lint/docs, Windows, macOS/cgo, worker, and Docker all green on the exact landed head
- Added request-level coverage proving a tab title resolves to the secondary tab ID and `UpdateDocumentStyleRequest.tabId` carries that ID

Maintainer adjustments: normalized result output to the existing `tabId` convention, added command-level regression coverage, and completed the changelog attribution.

Live-proof caveat: the contributor reports successful secondary-tab Docs API readback. I could not independently repeat the live mutation because the available local test OAuth was revoked and the existing-Chrome bridge was unavailable; the local HTTP boundary test and official Docs API contract provide the independent request-shape proof.

PR #891, opened 2026-06-30
feat(auth): Add 1Password keyring backend

lox:
Local and agent workflows currently have to choose between OS keychains, which can prompt repeatedly or behave poorly in headless contexts, and the encrypted file backend, which requires distributing `GOG_KEYRING_PASSWORD`. This adds a 1Password-backed option so gog auth material can live in a configured 1Password vault while local desktop workflows can use the 1Password app instead of a service account token.

This change adds a `1password` keyring backend using the 1Password Go SDK. It supports desktop-app auth through `onepassword_account` / `GOG_1PASSWORD_ACCOUNT`, service-account auth through `OP_SERVICE_ACCOUNT_TOKEN`, and vault/item selection through config or environment. 1Password items are API Credential entries titled `gogcli-keyring` by default, with the keyring key in `username` and the encoded payload in the concealed `credential` field.

Example local setup:

```bash
gog auth keyring 1password
gog config set onepassword_auth desktop
gog config set onepassword_account xxx.1password.com
gog config set onepassword_vault <vault-id>
```

Repeated short-lived subprocesses can also opt in to a state-dir access-token cache:

```bash
gog config set access_token_cache true
```

The cache stores only short-lived Google access tokens, checks it before reopening the refresh-token keyring, and clears it when stored auth is removed. That keeps agent-style Gmail/Calendar loops from repeatedly asking for 1Password approval while preserving 1Password as the source of truth for refresh tokens.

A couple of guardrails are included for review:

- 1Password only treats tagged `gogcli-keyring` API Credential items as managed, except for an empty starter template that is tagged on first use.
- `OP_SERVICE_ACCOUNT_TOKEN` remains environment-only; non-secret selectors can live in `config.json`.
- `auth doctor` reports the selected 1Password auth mode and missing vault/account/token configuration.


lox:
Could break this up into a smaller one @steipete if that makes it easier, but without the caching it's annoying to have at approve every cli invocation.

lox:
Going to break this up into smaller pieces.

PR #897, opened 2026-07-04
feat(auth): Add 1Password keyring backend

lox:
The previous 1Password backend PR was closed because the official SDK added too much binary and dependency weight to the CLI. This keeps the same opt-in keyring backend behavior, but switches the implementation to `github.com/lox/onepassword-sdk-native-go` so gog does not embed the official SDK's WASM runtime stack.

Measured on this branch against the commit immediately before the SDK swap:

```text
official SDK binary: 70,982,306 bytes
native SDK binary:   56,363,970 bytes
reduction:           14,618,336 bytes

official SDK module: 9,564 KiB
native SDK module:     512 KiB

official largest file: 9,496,396 bytes internal/wasm/core.wasm
native largest file:      43,834 bytes types.go
```

`go.mod` now has one new direct dependency on the native SDK and no longer needs `github.com/1password/onepassword-sdk-go`, `github.com/tetratelabs/wazero`, `github.com/extism/go-sdk`, `github.com/tetratelabs/wabin`, or the related WASM/Extism support packages.

The feature remains opt-in: users configure `GOG_KEYRING_BACKEND=1password` or `gog auth keyring 1password`, then set the 1Password vault/account/auth-mode config for either desktop app integration or service-account auth. Static/no-CGO builds still fail closed for desktop-app mode, while service-account mode remains available.


lox:
@steipete totally happy to provide an option for just using `op` directly. The UX of this is slightly nicer IMO in terms of the clarity in what you are approving. 

The other angle I am working on is I've forked the dead https://github.com/99designs/keyring that this project depends on (I was the original author at my first startup) into https://github.com/lox/keyring. It would provide a unified interface to the various keyring providers (including 1password) that this cli uses.

steipete:
Thanks for exploring this and for offering to adapt the approach. We do want better 1Password ergonomics, but we do not want to embed the native SDK: the binary-size increase is too large for this CLI.

I am closing this implementation direction. A small integration that shells out to the existing op CLI, keeps normal Gog builds dependency-free, and includes live locked/unlocked/error-path proof would be welcome as a fresh proposal.

PR #898, opened 2026-07-04
feat(auth): Use lox keyring providers

lox:
gog still depends on `github.com/99designs/keyring`, which is effectively abandoned and has accumulated unpatched security/dependency issues. The maintained fork is `github.com/lox/keyring/v2`, so this moves gog onto that supported module path instead of carrying risk from the old upstream.

gog's auth secret store also carried the old keyring adapter shape locally: provider ordering, file fallback setup, operation timeout wrapping, and the 99designs-compatible method surface. That made it harder to use the shared lox keyring packages without copying app-specific glue around.

This moves gog onto `github.com/lox/keyring/v2`, `github.com/lox/keyring-defaults`, and the external `github.com/lox/keyring-1password` provider. gog keeps the pieces that are actually app policy: service name selection, the encrypted file fallback prompt/dir, legacy file-key compatibility, locking, and token/default-account semantics.

For users, the existing `auto`, `keychain`, and `file` behavior is preserved, while `gog auth keyring 1password` and `GOG_KEYRING_BACKEND=1password` are available through the shared provider. The macOS Keychain trust behavior remains explicit via `defaults.KeychainTrustApplication(false)`, and operation timeouts now use `keyring.Timeout` from core.


steipete:
Thanks for putting this migration together. We are not going to combine a foundational keyring-provider migration with 1Password support in this form. The branch now conflicts with main, the new provider modules do not yet have a mature tagged-release history, and the change would regress Gog's current Keychain trust-application and doctor behavior.

I am closing this implementation direction. If we revisit the keyring library, it should be a standalone migration that preserves every current backend and upgrade path with live platform proof. A separate 1Password integration should follow the narrower op CLI approach discussed on the related proposal.

ronny-rentner:
I am really surprised how gogcli uses an unmaintaind keyring library with no security fixes in 4 years. 

This is super unprofessional. Most likely all your keys are compromised.

lox:
That seems like an overreaction @ronny-rentner, there are a few minor CVEs that as best as I can tell aren't exploitable. Also I was proposing using my brand new fork, which is equally eyebrow raising. 

I did make the mistake of conflating this change with the 1password changes I was noodling through. If there is interest from @steipete in just PR'ing the new fork of the keyring library, I can do that trivially, but I was being respectful of time and tokens.

ronny-rentner:
@lox I disagree. This is an under-reaction in a highly sensitive and security relevant area. I can only conclude that openclaw does not take security seriously. This is really bad when yo build agents and tools accessing all your private data.

As keyring is unmaintained, you cannot expect security reports to come in anymore after 4 years so even if there are only a few minor CVEs it means nothing.

PR #905, opened 2026-07-06
fix(calendar): display event-local times in event timezone

Hilo-Hilo:
## Summary
- Prefer an event's own `start.timeZone` / `end.timeZone` when computing local display fields (`startLocal`, `endLocal`, day-of-week, plain table output)
- Fall back to the containing calendar timezone only when the event has no valid timezone
- Add a regression test for an Asia/Seoul event on an Asia/Hong_Kong calendar, where Google may return the DateTime with the calendar offset while preserving the event timezone

## Why
Google Calendar events can carry a different timezone from the parent calendar. For example, an ICML conference event stored as `Asia/Seoul` on an `Asia/Hong_Kong` calendar may be returned as:

```json
{
  "dateTime": "2026-07-07T07:30:00+08:00",
  "timeZone": "Asia/Seoul"
}
```

That instant is `2026-07-07T08:30:00+09:00` in the event's own timezone. The current display path prefers the calendar timezone, so the derived local fields/table display can look one hour early for travel/conference events.

## Test plan
- `go test ./internal/cmd`
- `go test ./internal/cmd ./internal/...`

### DIFF
--- a/internal/cmd/calendar_event_days.go
+++ b/internal/cmd/calendar_event_days.go
@@ -145,32 +145,33 @@ func loadEventLocation(tz string) (*time.Location, bool) {
 	return tryLoadTimezoneLocation(tz)
 }
 
-// resolveEventTimezone resolves the timezone and location for an event.
-// It tries the calendar timezone first, then falls back to the event's timezone.
-// Returns the resolved timezone name and location. If loc is already provided,
-// it will be used as-is (only calendarTimezone may be updated for display).
+// resolveEventTimezone resolves the timezone and location used for local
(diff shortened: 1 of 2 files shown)


steipete:
Landed on `main` as 6790c6b1fb5c46fa884b2565ae03c986305cea60. Thank you, @Hilo-Hilo!

Maintainer follow-up expanded the regression matrix for invalid/missing timezone fallback and corrected travel events whose start and end carry different timezones, so each endpoint's local time and weekday are resolved independently. Proof: focused timezone tests, full `make ci`, and clean autoreview.

PR #908, opened 2026-07-08
feat(calendar): add --timezone display override to events and event

bxxd:
## Summary

- Add `--timezone` / `--tz` to `calendar events` and `calendar event`: an explicit display-timezone override (IANA name, or `local` for the system timezone)
- Flag unset keeps today's behavior exactly: each calendar's events render in that calendar's own timezone (#493)
- With an override set, every calendar in an `--all`/`--cal`/`--calendars` listing renders uniformly, and the per-calendar timezone lookups are skipped (they'd be unused)

## Why

Real-world failure that motivated this: a household agent builds a daily schedule from `calendar events --all -j`. Two of the account's nine calendars have their **calendar timezone set to UTC** — one is a SportsEngine ICS feed import (import calendars commonly come in as UTC and can't be re-zoned), one a Google calendar created with default settings. Google returns their event times in the calendar's timezone, and since #493 the table output and `startLocal`/`endLocal` JSON fields render calendar-local — which for these calendars *is* UTC:

```
2026-07-08T12:45:00Z | UTC | RP Swim Practice
2026-07-08T20:00:00Z | UTC | Sound 2031 Boys Practice
```

That 12:45Z is an 8:45am (America/New_York) swim practice. The agent published the family's schedule shifted +4h — swim at "12:45p", an evening practice that was actually 4pm. The stored instants are all correct; there is just no way today to ask the CLI to *display* them where the viewer lives. `--timezone` closes that hole while keeping #493's calendar-local default:

```
$ gog calendar events --all --timezone America/New_York ...
2026-07-08T08:45:00-04:00 | America/New_York | RP Swim Practice
2026-07-08T16:00:00-04:00 | America/New_York | Sound 2031 Boys Practice
```

Notes:

- **Display-only.** `--from`/`--to`/`--today` range parsing is unchanged (still the primary calendar's timezone).
- **Orthogonal to #905** (event-timezone vs calendar-timezone display preference). #905 changes which *stored* timezone wins by default; this flag is an explicit viewer override on top of whichever default wins. For a feed-import calendar whose events are natively UTC, no default-preference change can produce viewer-local display — only an override can.
- Naming follows the existing read-side precedent (`gmail messages --timezone`) and the calendar write-side `--timezone`/`--tz` aliases (#844).

## Test plan

- `make fmt`, `make lint` (0 issues), `go test ./internal/...` — all packages pass
- New regressions: UTC-calendar table + JSON output render in the override timezone; multi-calendar path renders uniformly and skips `/calendars/<id>` lookups (test handler deliberately 404s them); invalid value → usage error exit 2; empty value → no override; Kong-level flag wiring end-to-end
- Live-verified against the real account that hit this (output above)
- `make docs-commands` regenerated

---

🤖 Implemented with [Claude Code](https://claude.com/claude-code), human-reviewed before submission.

### DIFF
--- a/docs/commands.generated.md
+++ b/docs/commands.generated.md
@@ -93,7 +93,7 @@ Generated from `gog schema --json`.
     - [`gog calendar (cal) create-calendar (new-calendar) <summary> [flags]`](commands/gog-calendar-create-calendar.md) - Create a new secondary calendar
     - [`gog calendar (cal) delete (rm,del,remove) <calendarId> <eventId> [flags]`](commands/gog-calendar-delete.md) - Delete an event
     - [`gog calendar (cal) delete-calendar <calendarId>`](commands/gog-calendar-delete-calendar.md) - Delete an owned secondary calendar
-    - [`gog calendar (cal) event (get,info,show) <calendarId> <eventId>`](commands/gog-calendar-event.md) - Get event
+    - [`gog calendar (cal) event (get,info,show) <calendarId> <eventId> [flags]`](commands/gog-calendar-event.md) - Get event
     - [`gog calendar (cal) events (list,ls) [<calendarId> ...] [flags]`](commands/gog-calendar-events.md) - List events from a calendar or all calendars
     - [`gog calendar (cal) focus-time (focus) --from=STRING --to=STRING [<calendarId>] [flags]`](commands/gog-calendar-focus-time.md) - Create a Focus Time block
     - [`gog calendar (cal) freebusy [<calendarIds>] [flags]`](commands/gog-calendar-freebusy.md) - Get free/busy
--- a/docs/commands/gog-calendar-event.md
+++ b/docs/commands/gog-calendar-event.md
@@ -7,7 +7,7 @@ Get event
 ## Usage
 
 ```bash
-gog calendar (cal) event (get,info,show) <calendarId> <eventId>
+gog calendar (cal) event (get,info,show) <calendarId> <eventId> [flags]
 ```
 
 ## Parent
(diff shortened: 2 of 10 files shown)


bxxd:
Both review findings addressed in 02492b13:

- **Weekday/location mismatch**: `printCalendarEventWithTimezone` now derives `start-day-of-week`/`end-day-of-week` via `eventDaysOfWeekInLocation` with the same resolved display location used for `start-local`/`end-local`, matching what the JSON path (`wrapEventWithDaysWithTimezone`) already did. This also fixes the pre-existing mismatch when a calendar/event timezone crosses midnight relative to the display location, independent of the new flag.
- **Focused regression**: `TestCalendarEventCmd_TimezoneOverrideCrossDateWeekday` — a `02:00Z` Wednesday event with `--timezone America/New_York` must print `start-local 2026-07-07T22:00:00-04:00` with `start-day-of-week`/`end-day-of-week` `Tuesday`.

`make fmt` / `make lint` (0 issues) / `go test ./internal/...` all pass.

steipete:
Landed on `main` as 39425622ec597e7dd253f3a4310b9294c65481ea. Thank you, @bxxd!

Maintainer follow-up integrated the event-timezone behavior from #905 and added explicit precedence coverage: `--timezone` wins over event timezone, event timezone wins over calendar timezone, and calendar timezone remains the fallback. The combined tests also cover cross-midnight weekdays, JSON/table output, multi-calendar lookup skipping, invalid/empty values, and corrected user-facing default documentation. Proof: focused calendar tests, full `make ci`, and clean autoreview.

PR #909, opened 2026-07-09
Support Docs smart chips in cat output

### LINKED ISSUE #907: docs cat drops person smart chips from all output modes (plain, --plain, --json); readable only via --raw
### Summary

`docs cat` (plain, `--json`, and `--plain`) silently drops **person smart chips** from its output. A tagged person in a doc is a `person` element — a sibling of `textRun` inside `paragraph.elements` — and the text extractor only concatenates `textRun.content`, so the chip contributes nothing to the rendered text. The same applies to `date` chips and `richLink` (Drive file / calendar event) chips.

The result is silently misleading output: a table row like `Owner: @Jane Doe` renders as `Owner:` with an empty cell, with no indication anything was skipped. `--json` doesn't help — it returns the same lossy string wrapped in `{"tab": {"id", "index", "title", "text"}}`. Today the only way to see who is tagged is `cat --raw` plus a hand-rolled jq walk.

Notably the **write** side already has first-class chip support (`docs insert-person`, #638, `--at` anchors in #683) — so `gog` can insert a person chip it then cannot read back.

### Repro (self-contained, uses gog itself)

```bash
DOC=$(gog docs create "chip read test" -j | jq -r .id)
gog docs write "$DOC" --replace "Reviewer: "
gog docs insert-person "$DOC" --email sample@example.com --at "Reviewer: "

gog docs cat "$DOC"
# Output: "Reviewer: "            <-- chip dropped, no marker

gog docs cat "$DOC" --json | jq -r .tab.text
# Same lossy text; no person data anywhere in the JSON

gog docs cat "$DOC" --raw | jq '[.. | objects | select(has("person")) | .person.personProperties]'
# [{"name": "Sample Person", "email": "sample@example.com"}]  <-- data exists
```

Observed on v0.33.0 (also v0.22.0).

### Expected

Two parts, either would help; both would be ideal:

1. **`cat` text rendering**: render chips inline instead of skipping them, e.g.
   - person → `@Jane Doe <jane@example.com>`
   - date → the chip's display text
   - richLink → `[Title](url)`

   (Possibly behind a `--chips` flag if changing the default output is a concern, though rendering *something* seems strictly better than an invisible drop.)

2. **`cat --json` structured output**: include the chips as data, e.g. a `persons` / `chips` array per tab alongside `text`, so scripting consumers don't need to fall back to `--raw` + a recursive jq descent (which on multi-tab docs also means downloading the entire doc, since `--raw` returns all tabs).

### Related

- #638 — write-side chip insertion (shipped); this issue is the read-side mirror
- #818 — chips corrupting `find-replace` offset math (chips as an editing hazard; here they're an extraction gap)
- #697 — `raw` tab traversal (the current workaround path)

TurboTheTurtle:
Closes #907.

Summary
- Keeps existing plain-text output unchanged by default.
- Adds `--chips` for inline person/date/rich-link rendering.
- Adds additive `renderedText` and `chips` fields to JSON output when smart chips are present.

Evidence
- `make ci` passes.

### DIFF
--- a/docs/commands/gog-docs-cat.md
+++ b/docs/commands/gog-docs-cat.md
@@ -21,6 +21,7 @@ gog docs (doc) cat (text,read) <docId> [flags]
 | `--access-token` | `string` |  | Use provided access token directly (bypasses stored refresh tokens; token expires in ~1h) |
 | `-a`<br>`--account`<br>`--acct` | `string` |  | Account email, alias, or auto for authenticated Google API commands |
 | `--all-tabs` | `bool` |  | Show all tabs with headers |
+| `--chips` | `bool` |  | Render Google Docs smart chips inline in text output |
 | `--client` | `string` |  | OAuth client name (selects stored credentials + token bucket) |
 | `--color` | `string` | auto | Color output: auto\|always\|never |
 | `--disable-commands` | `string` |  | Comma-separated list of disabled commands; dot paths allowed |
--- a/internal/cmd/docs_commands_test.go
+++ b/internal/cmd/docs_commands_test.go
@@ -353,6 +353,59 @@ func newTabsTestServer(t *testing.T) (*docs.Service, func()) {
 	return docSvc, srv.Close
 }
 
+func newSmartChipDocsTestServer(t *testing.T) (*docs.Service, func()) {
+	t.Helper()
+
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if !strings.HasPrefix(r.URL.Path, "/v1/documents/") || r.Method != http.MethodGet {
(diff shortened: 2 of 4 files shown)


steipete:
Landed on `main` as f2497395a2f8b7a6ca13efb2525468327515ea64. Thank you, @TurboTheTurtle!

Maintainer follow-up strengthened numbered/tabbed output coverage, atomic byte-limit behavior, fallback handling, Markdown delimiter escaping, and MCP untrusted-content wrapping. Proof: focused smart-chip tests, full `make ci`, clean autoreview, and a disposable live Google Docs fixture containing native person, date, and Drive-file chips; default output remained unchanged, opt-in text/JSON rendered all three, and all fixtures were trashed.

PR #912, opened 2026-07-10
feat(auth): pass login_hint in auth add OAuth flows

InfiniviewStudios:
## Problem

`gog auth add <email>` knows exactly which account it expects, but the auth URL it builds never tells Google. When the requested scopes were granted recently, Google **silently auto-approves as the browser's default account** — no account chooser ever appears — so the flow completes as the wrong user in about a second and `auth add` fails with:

```
authorized as alex@company-b.com, expected alex@company-a.com
```

For anyone juggling multiple Google accounts in one browser profile, this makes re-auth a game of roulette: the guard correctly rejects the mismatched grant, but there is no way to steer the consent to the right account short of incognito windows.

## Fix

Pass the expected email as `login_hint` in all three `auth add` flows (browser listener, `--manual`, `--remote`), so Google pre-selects the requested account. This mirrors what the interactive accounts manager already does (`accounts_manager.go` appends `login_hint` since the permission-upgrade UI landed) — this PR just extends the same pattern to `auth add`.

- `AuthorizeOptions` gains a `LoginHint` field
- new `authorizeURLParams` helper wraps `pkceAuthURLParams` and appends `login_hint` when set, used by all three call sites
- `auth add` wires `c.Email` into the options

## Testing

- new `TestAuthorizeURLParamsLoginHint` (mirrors the existing `accounts_server_test.go` login_hint assertions): param present when set, absent when empty
- `go test ./internal/googleauth/ ./internal/cmd/` pass, `go vet` clean, `gofmt` clean
- live-verified: `gog auth add user@example.com --remote --step 1` now emits `login_hint=user%40example.com` in the auth URL, and the browser flow lands on the hinted account instead of the profile default

### DIFF
--- a/internal/cmd/auth_add.go
+++ b/internal/cmd/auth_add.go
@@ -172,6 +172,7 @@ func (c *AuthAddCmd) Run(ctx context.Context, flags *RootFlags) error {
 				DisableIncludeGrantedScopes: disableIncludeGrantedScopes,
 				Client:                      client,
 				RedirectURI:                 redirectURI,
+				LoginHint:                   strings.TrimSpace(c.Email),
 			})
 			if manualErr != nil {
 				return manualErr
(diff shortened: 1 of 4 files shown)


InfiniviewStudios:
Closing — decided to carry this as a local patch instead. Feel free to take the diff if it's useful.

PR #919, opened 2026-07-13
feat(gmail): add bounded inline attachment output

chrischall:
## Summary

- Add `gog gmail attachment --inline` for base64 content in JSON and plain output.
- Keep writing the attachment to the requested path; attachments above 3 MiB remain path-only with an explanatory reason.
- Return best-effort filename and MIME type metadata without changing the default output shape.
- Narrow the original proposal to transport only: PDF/text parsing, its dependency, and its subprocess boundary are removed.

## Safety and compatibility

- Inline mode always fetches provider bytes instead of trusting the size-only local cache.
- Base64 is generated from the fetched bytes retained in memory, not by reopening the caller-controlled destination path.
- Default cache validation still requires an exact attachment ID.
- Sender-controlled metadata is quoted when it contains terminal control characters.
- No new dependency; default `{path, cached, bytes}` JSON remains unchanged.

## Proof

- `make ci`: lint, deadcode, all Go and Node tests, generated command docs, docs coverage, and agent-skills checks pass.
- Structured final-tree review: clean.
- Signed exact-head live Gmail validation on `6401d196bfb24c0225e4fdb1d6ae4eb3dacef201`: help, credential-free dry-run, small inline round-trip, unchanged default shape, invalid-ID cache rejection, same-size stale-file refresh, and oversized fallback all pass. Temporary drafts and files were deleted; post-check found zero drafts.


chrischall:
@clawsweeper re-review — pushed 5e4cd61c fixing the --text --plain TSV contract (Go-quoted single field + regression test) and bounding PDF extraction with a 10s timeout + adversarial corrupt-PDF fixture; the PR body now includes redacted live CLI and MCP transcripts for default/--inline/--text and the fixed --plain output.

chrischall:
@clawsweeper re-review — 1d59b5bf replaces the goroutine timeout with a killable subprocess boundary: PDF parsing re-execs gog as a hidden `gmail pdf-extract` child under exec.CommandContext, so timeout = SIGKILL (no lingering parser work in long-lived MCP servers). Child re-checks the 3 MiB input cap, parent caps extracted text at 4 MiB. Tests spawn the real subprocess including a hung-child kill test; PR body has details and live re-verification.

PR #923, opened 2026-07-15
fix(gmail): accept padded raw message encoding

### LINKED ISSUE #922: bug(gmail): get --format raw fails "illegal base64 data" on multipart messages
## Summary

`gog gmail get <id> --format raw` fails on many messages with:

```
illegal base64 data at input byte <N>
```

It prints the envelope lines (`id` / `thread_id` / `label_ids`) and then dies, so the raw
RFC822 output is never emitted. `--format full` and `--format metadata` work on the *same*
message — only `raw` is affected.

## Repro

```bash
gog gmail get <id> --format raw       # fails: illegal base64 data at input byte <N>
gog gmail get <id> --format full      # works
gog gmail get <id> --format metadata  # works
```

## Affected class

Not multipart- or attachment-specific, though larger messages hit it more often by chance.
The Gmail API's `raw` field is **padded** base64url. Whether it fails depends only on the
RFC822 byte length:

- length divisible by 3 → no `=` padding → decodes fine
- length not divisible by 3 → 1–2 trailing `=` padding chars → **fails**

So roughly two-thirds of messages fail. In every failing case observed, the reported byte
offset `<N>` is exactly the index of the first `=` padding character (at the very end of the
base64 string). Example: a message with a 91208-char `raw` field failed at byte 91207 — the
lone trailing `=`.

## Root cause (v0.34.0)

`internal/cmd/gmail_get.go:130` decodes the API `raw` field with the **unpadded** URL
encoding:

```go
decoded, err := base64.RawURLEncoding.DecodeString(msg.Raw)
```

The Gmail API returns `raw` as *padded* base64url, but `base64.RawURLEncoding` is the padding-
free variant and rejects the trailing `=` as illegal input — hence "illegal base64 data at
input byte <N>" pointing at the padding.

## Suggested fix

Use the padded URL encoding:

```go
decoded, err := base64.URLEncoding.DecodeString(msg.Raw)
```

For robustness against responses that omit padding, normalize first (e.g. strip trailing `=`
and use `RawURLEncoding`, or pad to a multiple of 4 and use `URLEncoding`). A regression test
should cover a `raw` value whose decoded length is not a multiple of 3.

## Environment

gog v0.34.0. Root cause confirmed by reading source at `internal/cmd/gmail_get.go:130`.

goutamadwant:
## Summary

- Accept padded and unpadded base64url payloads from Gmail when `gog gmail get --format raw` renders an RFC822 message.
- Reuse the existing Gmail raw decoder so interactive retrieval and backup export follow the same decoding rules.
- Add a command-level regression covering a padded multipart-style raw message response.

Closes #922

## Real behavior proof

- Behavior addressed: a valid padded Gmail `raw` response failed before rendering the message.
- Before fix: `TestGmailGetCmd_RawAcceptsPaddedBase64URL` failed with `illegal base64 data at input byte 74`.
- After fix: the same command-level test passes and its output contains the decoded RFC822 message.
- Exact focused command: `go test ./internal/cmd -run '^TestGmailGetCmd_Raw(AcceptsPaddedBase64URL|Empty)$' -count=1`
- Test environment: the command runs against a local HTTP server returning the same Gmail API message shape, including padded base64url data.
- Limitation: this was not tested against an authenticated Gmail account.

## Validation

- `go test ./internal/cmd -count=1`
- `go test ./... -count=1`
- `go vet ./...`
- `make fmt-check`
- `make build`
- `make deadcode`
- `make docker-version-check docs-check agent-skills-check`
- `node --test scripts/eval-gws.test.mjs scripts/eval-gws-agents.test.mjs`
- `.tools/golangci-lint run --enable-only=govet,staticcheck,errcheck,unused --new-from-rev=origin/main ./internal/cmd/...` (`0 issues`)

The complete 44-linter invocation did not terminate locally after package loading; the focused core linter set passed, and repository CI remains the authoritative full lint run.

Disclosure: AI was used to understand the codebase and review the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.34.1 - Unreleased
 
 - Gmail: enforce per-account no-send guards before dry-run exits for first-class and Discovery send paths while preserving no-guard keyring avoidance. (#915, #916) — thanks @veteranbv.
+- Gmail: accept padded Gmail API base64url payloads in `gmail get --format raw` while retaining unpadded compatibility. (#922, #923) — thanks @goutamadwant.
 - MCP: add optional global and per-account capability ceilings in `config.json`, with narrow persistent write authorization, runtime-only restriction, and fail-closed selector validation. (#913) — thanks @mcaldas.
 
 ## 0.34.0 - 2026-07-11
--- a/internal/cmd/gmail_get.go
+++ b/internal/cmd/gmail_get.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"context"
-	"encoding/base64"
 	"strings"
 
 	"github.com/steipete/gogcli/internal/gmailcontent"
@@ -127,7 +126,7 @@ func (c *GmailGetCmd) Run(ctx context.Context, flags *RootFlags) error {
 			u.Err().Println("Empty raw message")
 			return nil
 		}
-		decoded, err := base64.RawURLEncoding.DecodeString(msg.Raw)
+		decoded, err := decodeGmailRaw(msg.Raw)
 		if err != nil {
 			return err
 		}
--- a/internal/cmd/gmail_get_cmd_test.go
+++ b/internal/cmd/gmail_get_cmd_test.go
@@ -322,6 +322,41 @@ func TestGmailGetCmd_Text_Metadata_WithAttachments(t *testing.T) {
 	}
 }
 
+func TestGmailGetCmd_RawAcceptsPaddedBase64URL(t *testing.T) {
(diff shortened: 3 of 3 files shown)


steipete:
Landed in 7785477091f90f8b0116b8b3c0f855bd78f4f65a from exact reviewed head fe7c109608dfa551f497bc2d88791b29369be5eb.

Proof:
- Live authenticated Gmail payload: 16,392 raw characters with one padding character. Previous main exited 1 with an illegal-base64 error; this head exited 0 and emitted 12,380 decoded bytes.
- Local make ci passed.
- AutoReview passed with no findings.
- Hosted Linux, Windows, Darwin/cgo, worker, and Docker checks passed on the exact head.

Thanks @goutamadwant for the focused fix and regression test.

PR #926, opened 2026-07-16
fix(calendar): expose --remove-meet

lox:
`calendar update --with-zoom` currently tells users with an existing Meet conference to run `--remove-meet` first, but the update command does not expose that flag. Following the CLI guidance therefore fails during argument parsing and leaves no first-class path for replacing Meet with Zoom.

This adds `--remove-meet` as a conference-data-only patch. It sends `conferenceData: null` with Calendar conference-data support enabled, leaving the event description and other fields untouched. The flag follows the existing recurring-event scope behavior and rejects conflicting Meet or Zoom creation flags.

For example:

```console
gog calendar update primary EVENT_ID --remove-meet
gog calendar update primary EVENT_ID --with-zoom
```

The generated command reference, CLI specification, and changelog now include the flag.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.34.1 - Unreleased
 
+- Calendar: add `calendar update --remove-meet` to clear only an event's Google Meet conference data before attaching another conferencing provider.
 - Docs: preserve external and internal text-run link targets in `docs cat --chips`, JSON, tab, table, and numbered output while keeping default text unchanged. (#917, #921) — thanks @neo-wanderer.
 - Gmail: enforce per-account no-send guards before dry-run exits for first-class and Discovery send paths while preserving no-guard keyring avoidance. (#915, #916) — thanks @veteranbv.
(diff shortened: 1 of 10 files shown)


steipete:
Closing this implementation rather than landing it.

The need is real: `calendar update --with-zoom` points users at a missing `--remove-meet` path. But exact head `3d3c1c7cf414351eb766f88db6cc94aabaf142bb` implements the Meet-specific flag by unconditionally appending `ConferenceData` to `NullFields`. That deletes any existing conference provider data, including non-Meet conferencing, without first fetching or classifying the event. The branch has also fallen behind current `main` and no longer has a clean merge state.

A scoped resubmission is welcome. The safe contract is:

- fetch and classify the existing conference provider before mutation;
- clear conference data only when it is Google Meet, otherwise preserve it or return a precise usage error;
- add regression coverage for non-Meet conference data;
- omit contributor-authored release notes so the maintainer can add changelog credit at landing;
- include redacted live Calendar proof for Meet removal and non-Meet preservation.

Thanks @lox for identifying the broken command guidance and proposing the missing CLI surface. A corrected resubmission should preserve that contribution context.

PR #934, opened 2026-07-22
fix(docs): emit MDEmptyLine for blank lines in ParseMarkdown

alexknowshtml:
## Summary

Blank lines in markdown input were silently dropped by `ParseMarkdown` (hit `continue` with no element emitted), causing all paragraphs to render without spacing when writing markdown to a Google Doc via the `batchUpdate` path.

`MDEmptyLine` was already handled in `MarkdownToDocsRequests` — the parser just never produced it. Consecutive blank lines collapse to one to avoid double-spacing.

## Changes

- `internal/cmd/docs_markdown.go`: emit `MDEmptyLine` for blank lines instead of skipping them; deduplicate consecutive blanks

## Test plan

- [ ] `gog docs write <docId> --file content.md --markdown --replace` produces paragraphs with visual spacing between them
- [ ] Consecutive blank lines in source markdown produce only one empty paragraph in the output


steipete:
Thanks for tackling this, Alex. The same blank-line preservation is already on `main` via #662 (`e4941e77`), including regression coverage for paragraph breaks and collapsed consecutive blank lines.

I also checked this exact head: it fails `TestParseMarkdown/mixed_content` because the existing expected parser output was not updated. Since `main` has the behavior and its regression coverage, I’m closing this as superseded rather than carrying a stale, conflicting branch. Appreciate the contribution.

PR #935, opened 2026-07-22
feat(docs): support --markdown with --tab-id in write command

alexknowshtml:
## Summary

`gog docs write --markdown --replace` previously used the Drive file-upload API, which converts the whole document and cannot target a specific tab. This PR adds a `batchUpdate`-based path used when `--tab-id` is provided.

## Changes

- **`docs_edit.go`**: Add `writeMarkdownToTab()` that loads the target tab, computes its end index, and delegates to `replaceDocsMarkdownRange` with the tab ID. Route `writeMarkdown()` through this path when `--tab-id` is set (previously returned an error). Also thread `c.TabID` through `DocsFindReplaceCmd.runMarkdown()`, which was previously ignoring it.
- **`docs_mutation.go`**: Add `tabID string` param to `replaceDocsMarkdownRange`; pass it through `Range`/`Location` in delete and insert requests. Add `applyTabIDToRequests()` helper that stamps `TabId` on all formatting requests returned by `MarkdownToDocsRequests` (paragraph style, text style, etc.). Guard empty-range deletes (`endIdx <= startIdx`) to avoid API errors on near-empty tabs.

## Test plan

- [ ] `gog docs write <docId> --file content.md --markdown --replace --tab-id t.abc123` writes formatted markdown to the specified tab
- [ ] `--markdown --replace` without `--tab-id` continues to use the Drive upload path (unchanged)
- [ ] `gog docs find-replace <docId> "placeholder" --content-file content.md --markdown --tab-id t.abc123` applies markdown formatting within the correct tab


steipete:
Thanks, Alex — this is useful direction. The safer tab-aware Markdown write workflow already landed on `main` in `58b866e9`, using `--tab` with comprehensive tab tests.

I checked this exact head as well: list-formatting requests omit `TabId`, and deferred table/image work can fall back to the first tab. In a multi-tab document that can mutate content outside the requested tab. Since `main` already covers the behavior safely, I’m closing this as superseded rather than asking you to rebase a historical stack. Thank you for the contribution.

PR #941, opened 2026-07-27
fix(sheets): never write outside the requested range on positional updates

cathrynlavery:
## Problem

`gog sheets update <id> "G31" "text, with, commas"` parses the positional value as comma-separated rows, so a single-cell write fans out down the column and silently overwrites neighboring cells. We hit this in production against a live tracker sheet.

## Fix

- When the target range resolves to a single cell, the positional value is written literally — commas and pipes are never treated as separators.
- For finite multi-cell ranges, the parsed matrix is validated against **maximum bounds**: only payloads whose rows or columns EXCEED the range are rejected (loudly, before the Values API call). Matrices smaller than the range pass through unchanged, preserving the partial-write behavior current main and the Values API support (review round 2 — thanks, the exact-equality check in round 1 was indeed a compatibility regression).
- Named ranges resolve through the existing range catalog before shape checks. `--values-json` and append behavior unchanged.
- `links set --cells-json` now accepts `@file` and `@-` (stdin), matching its sibling flags; docs updated.

## Real behavior proof (redacted)

Built from this branch (`v0.34.2-0.20260727222104+dirty`), run against a disposable test spreadsheet created for this proof; spreadsheet ID redacted:

```
$ gog sheets update <SHEET_ID> "A1" "text, with, commas"   # single cell, comma-bearing value
Updated 1 cells in 'gog-pr941-proof-disposable'!A1

$ gog sheets get <SHEET_ID> "A1:A3"   # A1 literal, neighbors untouched
text, with, commas

$ gog sheets update <SHEET_ID> "A1:B2" "r1a|r1b,r2a|r2b,r3a|r3b"   # 3 rows into 2-row range -> must reject BEFORE API call
positional values have 3 rows, which exceeds the update range maximum of 2 rows
exit: 2

$ gog sheets update <SHEET_ID> "C1:D2" "only-one-value"   # partial write within larger finite range -> must succeed (P1 compat)
Updated 1 cells in 'gog-pr941-proof-disposable'!C1

$ gog sheets get <SHEET_ID> "C1:D2"
only-one-value
```

All three review asks demonstrated live: (1) comma-containing single-cell update lands literally in exactly one cell with neighbors untouched, (2) oversize payload rejected before the API call (exit 2), (3) partial positional write within a larger finite range succeeds — the P1 case.

## Tests

`go test ./internal/sheetsvalues`, `go test ./internal/cmd -run 'TestSheetsUpdatePositional|TestSheetsLinksSet'`, and `go test ./...` all green (42 packages). New coverage: `TestSheetsUpdatePositionalPartialMultiCellRangeWritesOneCell` (asserts the recorder receives exactly `[["one"]]`), `TestParseArgsForShapeAllowsSmallerMatrix`, `TestParseArgsForShapeRejectsExceedingRangeBounds`, plus the existing single-cell literal and zero-write-on-error assertions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.34.2 - 2026-07-27
 
+- Sheets: keep positional updates within the requested range, preserving comma-bearing single-cell values and accurate named-range dry runs. (#941) — thanks @cathrynlavery.
 - Gmail: add source-specific `internalDateIso` timestamps to message and thread listings while preserving the legacy sender-header `date`. (#945, #946) — thanks @chrischall.
 - Gmail: prevent standalone draft updates from acquiring self-referential reply headers, with explicit recovery for affected drafts. (#942, #944) — thanks @chrischall.
 - Security: update gRPC-Go, PostCSS, and Sharp/libvips to patched releases, clearing three high-severity dependency alerts. (#940)
--- a/docs/commands/gog-sheets-links-set.md
+++ b/docs/commands/gog-sheets-links-set.md
@@ -20,7 +20,7 @@ gog sheets (sheet) links (hyperlinks) set (write) <spreadsheetId> [<cell> [<url>
 | --- | --- | --- | --- |
 | `--access-token` | `string` |  | Use provided access token directly (bypasses stored refresh tokens; token expires in ~1h) |
 | `-a`<br>`--account`<br>`--acct` | `string` |  | Account email, alias, or auto for authenticated Google API commands |
-| `--cells-json` | `string` |  | Batch: JSON array of {cell,url,text} or {cell,runs:[{text,uri}]} objects, written in one request. |
+| `--cells-json` | `string` |  | Batch: JSON array, @file, or @- for stdin containing {cell,url,text} or {cell,runs:[{text,uri}]} objects, written in one request. |
 | `--client` | `string` |  | OAuth client name (selects stored credentials + token bucket) |
 | `--color` | `string` | auto | Color output: auto\|always\|never |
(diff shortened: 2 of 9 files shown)


cathrynlavery:
@clawsweeper re-review

Both blockers addressed in `0bee84d`: finite-range validation is now max-bounds (partial writes within a larger finite range pass; only genuine overflow rejects), and the PR body now carries a redacted live-Sheets transcript from a branch-built binary covering the protected, rejected, and partial-write cases. Full suite green.

steipete:
Maintainer follow-up: I fixed the remaining named-range dry-run mismatch on the contributor branch in 2c888627c72cc152c0cfcd95bd550223b17dd8a7.

Positional updates to named or otherwise unresolved ranges now refuse an offline `--dry-run` preview with exit code 2 and direct users to A1 notation or `--values-json`. This preserves the repository's no-network dry-run contract while preventing a preview from showing delimiter-split values that the live named-range path would later reinterpret. Shape-independent `--values-json` named-range previews remain supported.

I also cleared the branch's five lint failures in the touched code/tests.

Proof:

- `go test ./internal/cmd -run 'TestSheetsUpdatePositional|TestSheetsLinksSet'` — pass
- `go test ./internal/sheetsvalues` — pass
- Built CLI: named-range positional dry-run exits 2; A1 single-cell dry-run exits 0 and preserves the full comma-bearing value; named-range `--values-json` exits 0; finite-range overflow exits 2
- Source-blind behavior contract — all clauses and anti-cheat probes pass

This clears the concrete review blocker. My recommendation is LAND: the branch fixes a live-proven neighboring-cell overwrite, preserves partial-write compatibility, and now has accurate dry-run behavior.

PR #944, opened 2026-07-28
fix(gmail): stop drafts update referencing the draft's own revision

### LINKED ISSUE #942: gmail drafts update: non-reply draft gains In-Reply-To/References pointing at its own previous revision
`gmail drafts update` rebuilds a draft using the draft's own thread as a reply target, so a non-reply draft gains `In-Reply-To`/`References` pointing at its own previous revision — an unsent message.

## Reproduction

```bash
# 1. A plain draft. No reply target of any kind.
gog gmail drafts create --subject "Scheduling" --body "..." --to co-parent@example.com --json
# -> draft r-178..., message 19fa6c3d5a030951
#    Headers: From, Subject, Date, Message-Id, MIME-Version, Content-Type.
#    No In-Reply-To, no References. Correct.

# 2. Update it. --reply-to-message-id is NOT passed.
gog gmail drafts update r-178... --subject "Scheduling" --body "... revised ..." --json

# 3. Re-read the draft.
gog gmail drafts get r-178... --json
```

Step 3 now shows:

```
In-Reply-To: <CAPMg=3SP=WNTkmviUdt6xUyRGsQxPwcEvrOOVpkmPbkXKn+sCQ@mail.gmail.com>
References:  <CAPMg=3SP=WNTkmviUdt6xUyRGsQxPwcEvrOOVpkmPbkXKn+sCQ@mail.gmail.com>
```

That `Message-Id` is the draft's **own prior version** from step 1 — a message that was never sent and that does not exist for any recipient. A second update chains onto the revision the first one produced.

## Cause

`GmailDraftsUpdateCmd.Run` treats the draft's existing thread as a reply target:

```go
// internal/cmd/gmail_drafts.go
targetThreadID := existingThreadID          // the draft's OWN thread
if threadID != "" {
    targetThreadID = threadID
}
...
if strings.TrimSpace(replyToMessageID) == "" {
    replyToThreadID = targetThreadID        // <-- thread continuity used as reply lineage
}
```

`replyToThreadID` flows into `prepareComposeReply` → `fetchReplyInfo`, whose thread branch picks `selectLatestThreadMessage(thread.Messages)` and sets `info.InReplyTo` from that message's `Message-ID` (`internal/cmd/gmail_reply.go`). For a draft that is not a reply, the only message in its thread *is* the draft, so it references itself.

The repo already has the correct selector — `selectLatestThreadReplyTarget` skips `DRAFT`-labelled and self-sent messages — but it is wired only into the `--quote` path. The header path uses the unfiltered `selectLatestThreadMessage`.

Thread continuity rides the same wire: `msg.ThreadId` is set from `reply.ThreadID` in `buildGmailMessage`, so simply not passing the thread as a reply target would also drop `threadId`. The two need separating.

## Why it matters

The draft claims to reply to a message that was never sent. On delivery, recipients' clients try to thread it against a nonexistent parent — it can appear as an orphaned reply or thread incorrectly. It is also silent: nothing in the command output indicates reply headers were added, so it surfaces only by fetching the draft and reading raw headers. The only workaround today is to delete the draft and recreate it, losing the draft id.

## Proposed behavior

1. `In-Reply-To`/`References` come only from an explicit reply target (`--reply-to-message-id` / `--thread-id`), never from the draft's own thread.
2. Reply headers never reference a draft — filter `DRAFT`-labelled messages out of thread reply-target selection, and refuse an explicit `--reply-to-message-id` that names a draft.
3. Genuine reply context survives an update by being carried forward from the draft's own stored headers rather than re-derived, so it stays anchored to the real parent.
4. Updating N times produces identical reply headers; `References` does not accumulate.
5. `thread_id` is preserved independently, as thread continuity.
6. The result reports the effective `inReplyTo`/`references` so a caller can verify without a raw-header fetch.
7. `--clear-reply-context` strips reply headers in place, so recovering a mis-threaded draft does not require delete-and-recreate.

Happy to send a PR — I have the fix and regression tests ready.

chrischall:
Fixes #942.

`gmail drafts update` rebuilds the whole message, and it was feeding the draft's own `threadId` back in as a **reply target**:

```go
targetThreadID := existingThreadID          // the draft's OWN thread
...
if strings.TrimSpace(replyToMessageID) == "" {
    replyToThreadID = targetThreadID        // thread continuity used as reply lineage
}
```

That flows to `fetchReplyInfo`, which anchors `In-Reply-To`/`References` to the newest message in the thread. For a draft that is not a reply, the only message in its thread *is* the draft — so every update pointed the draft at its own previous revision, a message that was never sent. A second update chained onto the first.

The two concepts were riding the same wire: `msg.ThreadId` is set from the resolved reply info in `buildGmailMessage`, so simply not passing the thread as a reply target would also have dropped `threadId`. This separates them.

## Changes

- **Reply lineage comes only from caller intent.** Only an explicit `--reply-to-message-id` / `--thread-id` is a reply target. The draft's own thread never is.
- **Thread continuity is carried independently**, via a new `ThreadContinuityID` on `draftComposeInput` applied only when no reply target resolved a thread of its own.
- **Genuine reply context survives an update** by re-applying the draft's own stored `In-Reply-To`/`References` verbatim instead of re-deriving them. Repeated updates are therefore idempotent and `References` does not accumulate.
- **Reply headers never name an unsent message.** `fetchReplyInfo` now filters `DRAFT`-labelled messages out of thread reply-target selection, and an explicit `--reply-to-message-id` pointing at a draft is refused. The draft-safe selector (`selectLatestThreadReplyTarget`) already existed but was wired only into the `--quote` path; the header path was using the unfiltered `selectLatestThreadMessage`.
- **The result reports effective threading.** `drafts create`/`update` now emit `inReplyTo`, `references`, and `replyContextSource` (`caller` / `carried` / null) so a caller can verify without a raw-header fetch. When lineage was carried rather than caller-specified, a warning goes to stderr.
- **`--clear-reply-context`** strips reply headers in place, so repairing a mis-threaded draft no longer requires delete-and-recreate (which loses the draft id).

## Compatibility

`threadId` behaviour is unchanged — a plain update keeps the draft in its thread, and a caller-supplied `--thread-id` still overrides it. The JSON result gains three fields; existing fields are untouched. No existing test needed changing, which is worth noting: nothing in the suite asserted the old self-referencing behaviour.

The `replyContextSource` field reports `null` rather than a `"none"` string when a draft has no reply context, so all three threading fields are null together.

## Testing

`make ci` green. New regression coverage in `internal/cmd/gmail_drafts_reply_context_test.go`:

- non-reply draft update emits no `In-Reply-To`/`References`, keeps `threadId`, and does not fetch its own thread at all
- three successive updates stay clean with a stable thread
- a genuine reply keeps pointing at the real parent across an update that does not re-specify it
- `References` does not drift across repeated updates of a real reply
- `--clear-reply-context` strips the headers and keeps the thread; it conflicts with an explicit reply target
- `--reply-to-message-id` naming a draft is refused
- `--thread-id` whose newest message is a draft anchors to the newest real message; a drafts-only thread is refused
- the JSON result reports effective `inReplyTo`/`references`, null when absent

---

## Review round 2 (updated after ClawSweeper review)

**Dropped the CHANGELOG entry** (`c9836084`) — `AGENTS.md` puts it in the maintainer's landing checklist.

**Added live Gmail proof** for both behaviors flagged as unverified. Detail in [this comment](https://github.com/openclaw/gogcli/pull/944#issuecomment-5135171317); summary:

- **Bug reproduced on current `main`** (`1b261244`): a standalone draft with no reply target gained `In-Reply-To`/`References` after one update, pointing at the revision that update replaced.
- **Standalone draft on this branch**: no reply headers after update, and `threadId` unchanged — thread continuity survives while the self-referential lineage does not.
- **Genuine reply on this branch**: real parent preserved verbatim across four updates, `References` does not accumulate, `replyContextSource` reports `caller` then `carried`.
- **Recovery**: a draft corrupted by `main` keeps its bad headers on this branch (they're indistinguishable from genuine context by inspection); `--clear-reply-context` removes them while keeping `threadId`.

Test drafts were deleted afterward. `make ci` green on `c9836084`.

On the open scope question — whether `--clear-reply-context` and the threading JSON fields should land here or in a follow-up — I've argued the case on #942 and will take whichever split you prefer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - Gmail: report exact file-input byte counts in compose dry-run output, including forwarded-message notes with trailing newlines. (#936, #937) — thanks @chrischall.
 - Setup: replace obsolete Google Cloud API enablement links with direct API Library URLs while preserving project-scoped recovery hints. (#933) — thanks @LinXunFeng.
 - Docs: `docs write` now rejects an explicitly empty `--tab`/`--tab-id` value instead of silently targeting the whole document.
-- Gmail: stop `drafts update` anchoring `In-Reply-To`/`References` to the draft's own previous revision, keeping thread continuity separate from reply lineage, preserving genuine reply context across repeated updates without growing `References`, refusing drafts as reply targets, reporting the effective reply context, and adding `--clear-reply-context`. (#942)
+- Gmail: stop `drafts update` anchoring `In-Reply-To`/`References` to the draft's own previous revision, keeping thread continuity separate from reply lineage, preserving genuine reply context across repeated updates without growing `References`, refusing drafts as reply targets, reporting the effective reply context, and adding `--clear-reply-context`. (#942, #944)
 
 ## 0.34.1 - 2026-07-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Gmail: report exact file-input byte counts in compose dry-run output, including forwarded-message notes with trailing newlines. (#936, #937) — thanks @chrischall.
 - Setup: replace obsolete Google Cloud API enablement links with direct API Library URLs while preserving project-scoped recovery hints. (#933) — thanks @LinXunFeng.
 - Docs: `docs write` now rejects an explicitly empty `--tab`/`--tab-id` value instead of silently targeting the whole document.
+- Gmail: stop `drafts update` anchoring `In-Reply-To`/`References` to the draft's own previous revision, keeping thread continuity separate from reply lineage, preserving genuine reply context across repeated updates without growing `References`, refusing drafts as reply targets, reporting the effective reply context, and adding `--clear-reply-context`. (#942)
 
 ## 0.34.1 - 2026-07-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.34.2 - 2026-07-27
 
+- Gmail: prevent standalone draft updates from acquiring self-referential reply headers, with explicit recovery for affected drafts. (#942, #944) — thanks @chrischall.
 - Security: update gRPC-Go, PostCSS, and Sharp/libvips to patched releases, clearing three high-severity dependency alerts. (#940)
 - Drive: add opt-in `upload --replace --if-version` conflict protection using a fail-closed ETag precondition, without changing unconditional replacement. (#929) — thanks @karstenevers.
 - Gmail: acknowledge terminal watch OAuth failures only after durably recording reauthentication recovery state, preserving the last processed history cursor for catch-up after renewal. (#930) — thanks @litang9.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@
 - Gmail: report exact file-input byte counts in compose dry-run output, including forwarded-message notes with trailing newlines. (#936, #937) — thanks @chrischall.
 - Setup: replace obsolete Google Cloud API enablement links with direct API Library URLs while preserving project-scoped recovery hints. (#933) — thanks @LinXunFeng.
 - Docs: `docs write` now rejects an explicitly empty `--tab`/`--tab-id` value instead of silently targeting the whole document.
-- Gmail: stop `drafts update` anchoring `In-Reply-To`/`References` to the draft's own previous revision, keeping thread continuity separate from reply lineage, preserving genuine reply context across repeated updates without growing `References`, refusing drafts as reply targets, reporting the effective reply context, and adding `--clear-reply-context`. (#942, #944)
 
 ## 0.34.1 - 2026-07-16
 
--- a/docs/commands/gog-gmail-drafts-update.md
+++ b/docs/commands/gog-gmail-drafts-update.md
@@ -28,6 +28,7 @@ gog gmail (mail,email) drafts (draft) update (edit,set) <draftId> [flags]
 | `--body-html-file` | `string` |  | HTML body file path ('-' for stdin) |
 | `--cc` | `string` |  | CC recipients (comma-separated) |
 | `--clear-attachments` | `bool` |  | Remove all attachments from the draft. By default, omitting --attach preserves the draft's existing attachments. |
+| `--clear-reply-context` | `bool` |  | Strip In-Reply-To/References from the draft, making it a standalone message. By default an update preserves the draft's existing reply headers. |
 | `--client` | `string` |  | OAuth client name (selects stored credentials + token bucket) |
 | `--color` | `string` | auto | Color output: auto\|always\|never |
 | `--disable-commands` | `string` |  | Comma-separated list of disabled commands; dot paths allowed |
--- a/internal/cmd/gmail_drafts.go
+++ b/internal/cmd/gmail_drafts.go
@@ -767,8 +767,8 @@ func (c *GmailDraftsUpdateCmd) Run(ctx context.Context, flags *RootFlags) error 
 	if replyToMessageID != "" && threadID != "" {
 		return usage("use only one of --reply-to-message-id or --thread-id")
 	}
-	if c.ClearReplyContext && (replyToMessageID != "" || threadID != "") {
-		return usage("use only one of --clear-reply-context or --reply-to-message-id/--thread-id")
+	if c.ClearReplyContext && (replyToMessageID != "" || threadID != "" || c.Quote) {
+		return usage("use only one of --clear-reply-context or --reply-to-message-id/--thread-id/--quote")
 	}
 
 	attachPaths, err := expandComposeAttachmentPaths(c.Attach)
--- a/internal/cmd/gmail_drafts.go
+++ b/internal/cmd/gmail_drafts.go
@@ -278,10 +278,20 @@ type draftComposeInput struct {
 	BodyHTML         string
 	ReplyToMessageID string
 	ReplyToThreadID  string
-	ReplyAll         bool
-	ReplyTo          string
-	Quote            bool
-	Attach           []string
+	// ThreadContinuityID keeps a rebuilt message in the thread it already
+	// belongs to. It is deliberately NOT a reply target: where a message lives
+	// says nothing about what it replies to. Only used when no reply target
+	// resolved a thread of its own.
+	ThreadContinuityID string
+	// InReplyTo/References carry a reply lineage that has already been
+	// resolved elsewhere — on update, the lineage the draft itself already
+	// stores. Applied verbatim, so repeated updates are idempotent.
+	InReplyTo  string
+	References string
+	ReplyAll   bool
+	ReplyTo    string
+	Quote      bool
+	Attach     []string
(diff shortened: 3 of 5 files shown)


paulmiu (commented):
Verified against a real production failure with the same shape: updating a legitimate Gmail reply draft caused the old implementation to select the draft revision as the newest thread message, then the replacement draft carried an In-Reply-To value whose target no longer existed. This PR fixes the underlying model by separating thread continuity from reply lineage, preserves genuine headers idempotently, excludes drafts as reply targets, and provides an explicit recovery path for malformed legacy drafts. I also ran all focused TestGmailDraftsUpdateCmd_ regression cases locally with Go 1.26.5; they pass. The complete upstream CI matrix is green. This is the right fix.

paulmiu:
I added a follow-up contribution that turns reply-context handling into a permanent integrity guarantee: https://github.com/chrischall/gogcli/pull/9

This is broader than repairing drafts affected by an older gog release. `drafts update` rebuilds the complete MIME message, so stored `In-Reply-To` and `References` are persisted state that must be validated before every rewrite. They can become stale or malformed through any API client, imports, deleted messages, manual tooling, future edge cases, or the known older gog bug. Without validation, gog can report success while preserving reply context that no recipient can thread against.

The follow-up establishes these invariants for every reply-draft update:

- healthy context is validated against a real, non-draft message and preserved verbatim;
- stale context is repaired from the newest surviving `References` ancestor, falling back to the latest sent or received message;
- an update with no safe parent fails before Gmail is mutated;
- output distinguishes `explicit`, `preserved`, and `repaired`, warning only on an actual repair;
- standalone drafts remain on the zero-fetch path.

The known legacy corruption is an important acceptance case, not the scope of the change. Correct threading becomes the default for human and automated callers without requiring knowledge of a recovery flag. `--clear-reply-context` remains only the intentional standalone override.

Focused draft-update tests, build, formatting, lint, dead-code analysis, generated docs/coverage, and agent-skill generation all pass.

paulmiu (commented):
Clarification to my earlier review: the companion improvement in https://github.com/chrischall/gogcli/pull/9 is not primarily legacy cleanup. It makes stored reply headers a validated invariant for every `drafts update` rewrite.

`In-Reply-To` and `References` can become stale through any API client, imports, deleted messages, manual tooling, future edge cases, or the known older gog bug. The contribution validates healthy context, automatically repairs stale context using real messages in the Gmail thread, fails before mutation when repair is unsafe, and exposes `explicit` / `preserved` / `repaired` state to automation. Standalone drafts remain on the zero-fetch path.

The legacy corruption is one regression case proving the need for this ongoing guardrail; the long-term benefit is that gog no longer silently accepts invalid reply context from any source.

chrischall:
Both items addressed.

## ✅ Remove the release-owned changelog entry

Dropped (`c9836084`). `AGENTS.md` puts `CHANGELOG.md` in the maintainer's landing checklist — "update `CHANGELOG.md` (PR #/issue + thanks)" — so it isn't mine to write. Same correction applied to #946.

## ✅ Real behavior proof

Built from `main` (`1b261244`) and from this branch, both run against a live Gmail account. Full write-up on #942; the results:

**The bug, reproduced on current `main`.** A plain draft with no reply target, updated once without `--reply-to-message-id`:

```
Subject:     proof standalone
Message-Id:  <CAPMg=3QfdQCByhqrpXFw41EHB=nu_S41tMR-F6QdQyNFmvrt-g@mail.gmail.com>
In-Reply-To: <CAPMg=3Rt_TeN8oZ6hLwiR=k-rHfAQ6G86vNvNFonRa0=Qd5ffA@mail.gmail.com>
References:  <CAPMg=3Rt_TeN8oZ6hLwiR=k-rHfAQ6G86vNvNFonRa0=Qd5ffA@mail.gmail.com>
```

**Standalone draft on this branch** — the first of the two behaviors you asked to see:

```
$ gog gmail drafts create --subject "proof fixed" --body "v1" --json
{ "draftId": "r1982123979045837716", "inReplyTo": null, "replyContextSource": null,
  "message": { "id": "19fb46bf5bb1988f", "threadId": "19fb46bf5bb1988f" } }

$ gog gmail drafts update r1982123979045837716 --subject "proof fixed" --body "v2 revised" --json
{ "inReplyTo": null, "replyContextSource": null,
  "message": { "id": "19fb46bf6199d6c4", "threadId": "19fb46bf5bb1988f" } }

--- raw headers of the updated message ---
Subject:     proof fixed
Message-Id:  <CAPMg=3S94gW5bDCGxWdp2srCcZM=2Q-0cnzwuN149fC9zdUROA@mail.gmail.com>
In-Reply-To: (absent)
References:  (absent)
threadId:    19fb46bf5bb1988f
```

No reply headers, and Gmail kept the thread — which is the specific thing you flagged as unproven: thread continuity survives while the self-referential lineage does not.

**Genuine reply on this branch** — the second behavior. A draft replying to a real received message, then updated four times:

```
$ gog gmail drafts create --reply-to-message-id 19fb4605c40146e0 --body "reply v1" --json
inReplyTo:          <ISmm9tpAS3-Oqba-xL7TyA@geopod-ismtpd-110>
replyContextSource: caller

--- after update 3 ---
inReplyTo:          <ISmm9tpAS3-Oqba-xL7TyA@geopod-ismtpd-110>
references:         <ISmm9tpAS3-Oqba-xL7TyA@geopod-ismtpd-110>
replyContextSource: carried
threadId:           19fb4605c40146e0

--- after update 4 ---
inReplyTo:          <ISmm9tpAS3-Oqba-xL7TyA@geopod-ismtpd-110>
references:         <ISmm9tpAS3-Oqba-xL7TyA@geopod-ismtpd-110>
replyContextSource: carried
threadId:           19fb4605c40146e0
```

The real parent is preserved verbatim, `References` does not accumulate across repeated updates, and the source is reported as `caller` on the update that set it and `carried` thereafter. Each update also warns on stderr that context was preserved and names the escape hatch.

**Recovery, live.** Taking the draft `main` corrupted at the top of this comment and updating it on this branch keeps the bad headers (`replyContextSource: carried` — they are indistinguishable from genuine ones by inspection). `--clear-reply-context` removes them while keeping `threadId`. That's the case for keeping the flag in this PR rather than deferring it; argued in full on #942, since it's really the scope question rather than a finding here.

All test drafts were deleted afterward.

`make ci` green on `c9836084`.

chrischall:
@clawsweeper re-review

steipete:
Maintainer follow-up: I fixed the remaining `--clear-reply-context --quote` interaction on the contributor branch in 48cb088b6e3c64d50d68ae2272ce762a66240cee.

The command now rejects `--clear-reply-context` together with `--quote`, `--reply-to-message-id`, or `--thread-id` before authentication. Regression coverage includes all three conflicts.

Proof:

- `go test ./internal/cmd -run 'TestGmailDraftsUpdateCmd_ClearReplyContext'` — pass
- Built CLI invocation with `--clear-reply-context --quote` — exit 2 with the mutual-exclusion error
- Source-blind behavior contract — all clauses pass, including the clear-only dry-run control (exit 0)

This clears the concrete code-review blocker. My recommendation is LAND: the core bug and recovery path have direct live-Gmail proof in the existing discussion, and the branch is now internally consistent.

PR #946, opened 2026-07-28
feat(gmail): add offset-bearing internalDateIso to message and thread listings

### LINKED ISSUE #945: gmail listings emit a naive "2026-07-28 03:36" date with no offset, derived from the sender's Date header
`gog gmail messages`/`threads` JSON emits `date` in `listDateLayout` (`"2006-01-02 15:04"`), which carries **no offset**:

```json
{ "id": "19fa6c3d5a030951", "date": "2026-07-28 03:36", "subject": "…" }
```

A caller reading that has to already know which zone gog was configured for. Get it wrong and the value can land on the wrong calendar **day**, not merely the wrong hour — `2026-07-28 03:36` rendered in UTC is `2026-07-27 23:36` Eastern, a different date. That is the difference between "this went out Tuesday" and "this went out Monday night", which in a legal or scheduling record is the whole question.

This bit in practice on a headless deployment where nothing set `GOG_TIMEZONE`, so `time.Local` was UTC and the naive value silently meant UTC while every consumer read it as local.

Two things would fix it:

**1. Carry an explicit offset.** The compact `date` column is good for humans and should stay, but JSON consumers need an unambiguous value.

**2. Prefer `internalDate` over the `Date:` header.** `formatGmailDateInLocation` parses the header written by the *sending* client, which varies in format and offset and is occasionally just wrong. The Gmail API's `internalDate` is its own receipt timestamp (epoch ms, UTC) and is authoritative.

Proposal: add a `dateIso` field (RFC3339, from `internalDate`, rendered in the configured location) alongside the existing `date`, leaving current output and the human table untouched. Purely additive, so nothing that reads `date` today breaks.

Happy to send a PR — I have it implemented with tests.

chrischall:
Fixes #945.

## The problem

Message and thread listings render `date` with `listDateLayout` (`"2006-01-02 15:04"`), which carries **no offset**:

```json
{ "id": "19fa6c3d5a030951", "date": "2026-07-28 03:36", "subject": "…" }
```

A JSON consumer reading that has to already know which zone gog was configured for. Get it wrong and the value moves onto the wrong calendar **day**, not merely the wrong hour — that same instant rendered in Eastern is `2026-07-27 23:36`, a different date. That is the difference between "this went out Tuesday" and "this went out Monday night".

It bites hardest on headless deployments. Mine ran in a container with nothing setting `GOG_TIMEZONE`, so `time.Local` was UTC and the naive value silently meant UTC while every consumer read it as local time.

## The change

Add `internalDateIso` **alongside** `date` rather than reformatting it. `date` is the compact human column in the CLI table, and changing it would regress output for everyone reading it today. This is purely additive.

```json
{
  "id": "19fa6c3d5a030951",
  "date": "2026-07-27 23:36",
  "internalDateIso": "2026-07-27T23:36:00-04:00"
}
```

`internalDateIso` also switches **source**. `formatGmailDateInLocation` parses the `Date:` header written by the sending client, which varies in format and offset and is occasionally just wrong. `internalDateIso` derives from the API's `internalDate` — epoch milliseconds, UTC, Gmail's own internal message timestamp. For normal SMTP-received mail that is when Google accepted the message, which Gmail documents as more reliable than the `Date` header. (It is not universally independent of that header: for API-migrated mail the importing client may derive `internalDate` from it.) It is still rendered in the configured location, so `--timezone` / `GOG_TIMEZONE` / config all continue to apply.

An absent `internalDate` yields `""`, so the field stays `omitempty` rather than claiming the Unix epoch.

Applies to both `messageItem` (`gmail messages`) and `threadItem` (`gmail threads`).

## Compatibility

Nothing that reads `date` changes — same field, same layout, same location. The human table is untouched. No existing test needed modifying.

## Testing

`make ci` green. New `internal/cmd/gmail_date_iso_test.go` covers:

- the reported shape: an instant that is `03:36` UTC renders as `2026-07-27T23:36:00-04:00` in Eastern — the previous day — while the legacy `date` column keeps its existing output
- DST via the IANA database: January is `-05:00`, July `-04:00`
- absent `internalDate` (0 and negative) yields `""`
- a nil location falls back to `time.Local`, matching `formatMailDateInLocation`
- round-trips to the same instant across `America/New_York`, `UTC`, `Asia/Tokyo` and `Asia/Kolkata`, so the offset is carried rather than dropped

Tests `t.Skipf` when tzdata is unavailable rather than failing.

---

## Review round 2 (updated after ClawSweeper review)

**Renamed `dateIso` → `internalDateIso`.** Next to `date`, the old name read as "`date`, in ISO" — a representation change — when it is a different value from a different source. Not `receivedAt`: `internalDate` is the *send* time for messages you sent. See #945 for the full reasoning.

**Fixed a P1 that would have shipped the feature emitting nothing.** The message-summary fetch used a `fields` partial-response mask that never named `internalDate`, so `InternalDate` was always `0` and `omitempty` deleted the output field. A `fields` mask drops what it does not name *silently* — no API error — and unit tests that build `gmail.Message` values directly cannot see it. The selector now names the field and lives next to the metadata headers it travels with, and the new `gmail_message_summary_fields_test.go` uses a stub that honors the mask the way Gmail does. Removing `internalDate` from the selector fails that test with the exact production symptom (`internalDateIso = ""`).

The thread path needed no change — it has no `fields` mask, and `format=metadata` already returns `internalDate`. Live output below confirms it.

**Dropped the CHANGELOG entry** — `AGENTS.md` puts it in the maintainer's landing checklist.

## Real behavior proof

Built from this branch, run against a live Gmail account with `--timezone America/New_York`.

`gog gmail messages search "in:inbox" --max 2 --json` (message path, selector fixed):

```json
{
  "messages": [
    {
      "id": "19fb4605c40146e0",
      "threadId": "19fb4605c40146e0",
      "date": "2026-07-30 14:54",
      "internalDateIso": "2026-07-30T14:54:00-04:00",
      "from": "support@splitwise.com",
      "subject": "An update to the Splitwise API",
      "labels": ["UNREAD", "IMPORTANT", "CATEGORY_UPDATES", "INBOX"]
    }
  ]
}
```

`gog gmail search "in:inbox" --max 2 --json` (thread path, unmodified):

```json
{
  "threads": [
    {
      "id": "19fb44d34ac74f1f",
      "date": "2026-07-30 14:33",
      "internalDateIso": "2026-07-30T14:33:05-04:00",
      "subject": "CMS Inside Connection July 30, 2026",
      "messageCount": 1
    }
  ]
}
```

Note the second: `date` is `14:33`, `internalDateIso` is `14:33:05`. `listDateLayout` has no seconds, so even where the two agree on the instant the legacy column is lossy.

`date` is byte-identical to before on both paths.

**On how often the two sources disagree:** I scanned 60 live messages comparing the `Date` header against `internalDate` — zero disagreed to the minute. Well-behaved senders agree. The case for the new field is the unambiguous offset and the authoritative source, not a divergence I can demonstrate on demand.

## Review round 3

**Qualified the `internalDate` provenance claims (P3).** The finding is right and I'd overstated it. Gmail documents that for **API-migrated mail** the importing client may derive `internalDate` *from* the `Date` header, so my comments claiming the two are always distinct, and that `internalDate` is universally Google's receipt time, were wrong at the edges.

Corrected in all three places (`gmail_date.go`, `messageItem`, `threadItem`) and in this body: `internalDate` is described as Gmail's **internal message timestamp**, equal to Google's acceptance time for normal SMTP mail, with the migrated-mail exception stated explicitly. The field is now "separately sourced" rather than "a different value" — the distinction that actually matters to a consumer is that it doesn't depend on the sender's *formatting*, not that it always disagrees.

This doesn't change the case for the field: the offset is explicit either way, and where the two sources happen to coincide they coincide correctly.

## Review round 4

**Fixed the P2 millisecond truncation.** `internalDate` is epoch milliseconds, but the formatter used `time.RFC3339`, which has no fractional seconds — an instant ending in nonzero milliseconds serialized to a value that parsed back to the preceding whole second. Nonzero-millisecond instants now render with a fixed three-digit fraction (`2026-07-27T23:36:05.450-04:00`); whole-second instants keep the fraction-free form the field has emitted since it was introduced, so the live output shown above is byte-identical.

Coverage added first, red-then-green:

- `TestFormatGmailDateISO_MillisecondPrecision` pins the exact rendering at `.007`, `.450`, and `.999`, and that a whole second stays fraction-free
- `TestFormatGmailDateISO_RoundTripsToSameInstant` now round-trips a nonzero-millisecond instant across all four zones, and failed with the truncation symptom before the fix

`go test ./internal/cmd -run TestFormatGmailDateISO` and `make ci` both green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Gmail: report exact file-input byte counts in compose dry-run output, including forwarded-message notes with trailing newlines. (#936, #937) — thanks @chrischall.
 - Setup: replace obsolete Google Cloud API enablement links with direct API Library URLs while preserving project-scoped recovery hints. (#933) — thanks @LinXunFeng.
 - Docs: `docs write` now rejects an explicitly empty `--tab`/`--tab-id` value instead of silently targeting the whole document.
+- Gmail: add an offset-bearing `dateIso` to message and thread listings, sourced from the API's authoritative `internalDate` rather than the sender's `Date` header, so a JSON consumer is no longer left to guess which zone the naive `date` column is in. (#945)
 
 ## 0.34.1 - 2026-07-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@
 - Gmail: report exact file-input byte counts in compose dry-run output, including forwarded-message notes with trailing newlines. (#936, #937) — thanks @chrischall.
 - Setup: replace obsolete Google Cloud API enablement links with direct API Library URLs while preserving project-scoped recovery hints. (#933) — thanks @LinXunFeng.
 - Docs: `docs write` now rejects an explicitly empty `--tab`/`--tab-id` value instead of silently targeting the whole document.
-- Gmail: add an offset-bearing `dateIso` to message and thread listings, sourced from the API's authoritative `internalDate` rather than the sender's `Date` header, so a JSON consumer is no longer left to guess which zone the naive `date` column is in. (#945, #946)
 
 ## 0.34.1 - 2026-07-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - Gmail: report exact file-input byte counts in compose dry-run output, including forwarded-message notes with trailing newlines. (#936, #937) — thanks @chrischall.
 - Setup: replace obsolete Google Cloud API enablement links with direct API Library URLs while preserving project-scoped recovery hints. (#933) — thanks @LinXunFeng.
 - Docs: `docs write` now rejects an explicitly empty `--tab`/`--tab-id` value instead of silently targeting the whole document.
-- Gmail: add an offset-bearing `dateIso` to message and thread listings, sourced from the API's authoritative `internalDate` rather than the sender's `Date` header, so a JSON consumer is no longer left to guess which zone the naive `date` column is in. (#945)
+- Gmail: add an offset-bearing `dateIso` to message and thread listings, sourced from the API's authoritative `internalDate` rather than the sender's `Date` header, so a JSON consumer is no longer left to guess which zone the naive `date` column is in. (#945, #946)
 
 ## 0.34.1 - 2026-07-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.34.2 - 2026-07-27
 
+- Gmail: add source-specific `internalDateIso` timestamps to message and thread listings while preserving the legacy sender-header `date`. (#945, #946) — thanks @chrischall.
 - Gmail: prevent standalone draft updates from acquiring self-referential reply headers, with explicit recovery for affected drafts. (#942, #944) — thanks @chrischall.
 - Security: update gRPC-Go, PostCSS, and Sharp/libvips to patched releases, clearing three high-severity dependency alerts. (#940)
 - Drive: add opt-in `upload --replace --if-version` conflict protection using a fail-closed ETag precondition, without changing unconditional replacement. (#929) — thanks @karstenevers.
--- a/internal/cmd/gmail_date.go
+++ b/internal/cmd/gmail_date.go
@@ -48,9 +48,13 @@ func formatGmailDateInLocation(raw string, loc *time.Location) string {
 //     zone gog was configured for. Get that wrong and the timestamp can land on
 //     the wrong calendar DAY, which is what makes it worse than merely
 //     imprecise.
-//   - internalDate is the Gmail API's own receipt timestamp (epoch
-//     milliseconds, UTC). The Date header is written by the sending client and
-//     varies in format, offset, and honesty; internalDate does not.
+//   - internalDate is Gmail's own internal message timestamp (epoch
+//     milliseconds, UTC), independent of the sender's formatting. For normal
+//     SMTP-received mail it is when Google accepted the message, which Gmail
+//     documents as more reliable than the Date header. It is not universally
+//     independent of that header: for API-migrated mail the importing client
+//     may derive internalDate FROM the Date header, in which case the two
+//     agree by construction rather than by coincidence.
 //
 // Returns "" when internalDate is absent, so the field stays omitempty rather
 // than claiming the Unix epoch.
--- a/internal/cmd/gmail_date.go
+++ b/internal/cmd/gmail_date.go
@@ -39,6 +39,31 @@ func formatGmailDateInLocation(raw string, loc *time.Location) string {
 	return formatMailDateInLocation(raw, loc, listDateLayout)
 }
 
+// formatGmailDateISO renders a message's internalDate as RFC3339 in loc.
+//
+// Two things make this the value a machine consumer should read:
+//
+//   - It carries an explicit offset. listDateLayout ("2006-01-02 15:04") does
+//     not, so a caller that sees "2026-07-28 03:36" has to already know which
+//     zone gog was configured for. Get that wrong and the timestamp can land on
+//     the wrong calendar DAY, which is what makes it worse than merely
+//     imprecise.
+//   - internalDate is the Gmail API's own receipt timestamp (epoch
+//     milliseconds, UTC). The Date header is written by the sending client and
+//     varies in format, offset, and honesty; internalDate does not.
+//
+// Returns "" when internalDate is absent, so the field stays omitempty rather
+// than claiming the Unix epoch.
+func formatGmailDateISO(internalDateMillis int64, loc *time.Location) string {
+	if internalDateMillis <= 0 {
+		return ""
+	}
+	if loc == nil {
+		loc = time.Local
+	}
+	return time.UnixMilli(internalDateMillis).In(loc).Format(time.RFC3339)
+}
+
 // formatQuoteDate renders an original message's Date header in the Gmail-style
 // human form used for quoted replies and forwarded-message headers, converted
 // into loc (matching how Gmail shows the attribution in the reader's timezone).
--- a/internal/cmd/gmail_date.go
+++ b/internal/cmd/gmail_date.go
@@ -65,7 +65,14 @@ func formatGmailDateISO(internalDateMillis int64, loc *time.Location) string {
 	if loc == nil {
 		loc = time.Local
 	}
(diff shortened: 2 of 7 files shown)


chrischall:
Thanks — the P1 was real, and it was the kind of bug my tests were structurally unable to catch. Walking each finding:

## ✅ Request `internalDate` for message summary fetches — real, fixed

Confirmed. `Fields("id,threadId,labelIds,payload(headers)")` never asked Gmail for `internalDate`, so `msg.InternalDate` was always `0`, `formatGmailDateISO` returned `""`, and `omitempty` deleted the field. The feature would have shipped emitting nothing.

My unit tests could not have caught this: they construct `gmail.Message` values directly, so the field is populated by definition. A `fields` mask fails *silently* — the value arrives zeroed with no API error — so this only shows up against live Gmail.

Two changes rather than one. The selector now names the field:

```go
const gmailMessageSummaryFields = "id,threadId,labelIds,internalDate,payload(headers)"
```

and it moved next to `gmailMessageSummaryMetadataHeaders` in `gmail_metadata_headers.go` with a comment stating the coupling, so the selector and the struct it populates are read together.

And a regression test whose stub **honors the mask the way Gmail does** — it returns `internalDate` only when the selector asks for it:

```go
if strings.Contains(mask, "internalDate") {
    msg["internalDate"] = internalDateMillis
}
```

Verified load-bearing: removing `internalDate` from the selector fails it with exactly the production symptom.

```
--- FAIL: TestGmailMessagesSearch_RequestsInternalDateForISOOutput
    internalDateIso = "", want "2026-07-27T23:36:00-04:00"
    (fields mask sent: "id,threadId,labelIds,payload(headers)")
```

## ❌ Request `internalDate` for thread summary fetches — false positive

This one I can't reproduce, and I think the finding generalized from the message path. The thread path never had a `fields` mask:

```go
fullThread, err := svc.Users.Threads.Get("me", threadID).
    Format("metadata").
    MetadataHeaders(gmailMessageSummaryMetadataHeaders...).
    Context(ctx).Do()
```

`format=metadata` returns the full Message resource minus the body, and `internalDate` is part of that resource. Only a `fields` mask strips it. Live output below shows `internalDateIso` populated on this path with **no change to the call**.

## ✅ Describe it as receipt time, not `date`'s instant — fixed, and renamed

You were right that the comment was false, and chasing it exposed that the *field name* had the same problem. `dateIso` reads as "`date`, in ISO" — precisely what it is not. It is a different value from a different source, and a name that implies otherwise is a false friend that outlives any comment.

Renamed **`dateIso` → `internalDateIso`**, one of the two names suggested on #945.

I did not use `receivedAt`: `internalDate` is the send time for messages *you* sent, so "received" would be wrong for half the mailbox. `internalDateIso` names its provenance exactly and claims nothing further.

Honest note on the motivating divergence: I scanned 60 live messages and found **zero** cases where the `Date` header and `internalDate` disagreed to the minute. Well-behaved senders agree. The rename is justified by provenance and by the skewed/malformed tail, not by a discrepancy I can demonstrate on demand.

## ✅ Remove the release-owned changelog entry — fixed

Dropped. `AGENTS.md` puts `CHANGELOG.md` in the landing checklist ("update `CHANGELOG.md` (PR #/issue + thanks)"), so it is yours to write at merge, not mine. Same correction applied to #944.

## Real behavior proof

Built from this branch, run against a live Gmail account, `--timezone America/New_York`. Addresses and IDs are real-but-mine; nothing redacted was load-bearing.

`gog gmail messages search "in:inbox" --max 2 --json` — the message path, with the selector fix:

```json
{
  "messages": [
    {
      "id": "19fb4605c40146e0",
      "threadId": "19fb4605c40146e0",
      "date": "2026-07-30 14:54",
      "internalDateIso": "2026-07-30T14:54:00-04:00",
      "from": "support@splitwise.com",
      "subject": "An update to the Splitwise API",
      "labels": ["UNREAD", "IMPORTANT", "CATEGORY_UPDATES", "INBOX"]
    }
  ]
}
```

`gog gmail search "in:inbox" --max 2 --json` — the thread path, **unmodified**:

```json
{
  "threads": [
    {
      "id": "19fb44d34ac74f1f",
      "date": "2026-07-30 14:33",
      "internalDateIso": "2026-07-30T14:33:05-04:00",
      "subject": "CMS Inside Connection July 30, 2026",
      "messageCount": 1
    }
  ]
}
```

Note the second one: `date` is `14:33`, `internalDateIso` is `14:33:05`. The legacy column has no seconds — another reason a parser should read the new field.

`date` is byte-identical to before on both paths.

chrischall:
@clawsweeper re-review

chrischall:
✅ **Qualify the internalDate provenance claims (P3)** — fixed in `ecfa8091`. The finding is correct and I had overstated the claim.

Gmail documents that for **API-migrated mail** the importing client may derive `internalDate` *from* the `Date` header. So "internalDate is Google's receipt time" and "it is a different value from `date`" are both true of normal SMTP mail and false at that edge — the two can share a source, in which case they agree by construction.

Corrected in all three places (`gmail_date.go`, `messageItem`, `threadItem`) and in the PR body. It's now described as Gmail's **internal message timestamp**, with acceptance-time given as the normal-SMTP case and the migrated-mail exception stated explicitly.

The framing also changed from "a DIFFERENT value" to "SEPARATELY SOURCED". That's the claim that actually holds, and it's the one a consumer needs: `internalDateIso` doesn't depend on the sender's *formatting*. Whether it happens to agree with `date` on a given message isn't the point — where the two coincide, they coincide correctly.

This doesn't weaken the case for the field. The offset is explicit either way, and the naive `date` column remains unparseable without out-of-band knowledge of the configured zone regardless of where its value came from.

`make ci` green on `ecfa8091`. No code paths changed — comments and PR prose only, so the live output posted above still stands.

On the remaining decision — whether to guarantee `internalDateIso` as a supported additive field — that one's yours. Worth noting for it: the field is `omitempty` and absent when `internalDate` is, so a consumer with strict unknown-field handling sees a new optional key rather than a changed one, and nothing that reads `date` today is affected.

steipete:
Maintainer verification complete.

I reproduced the intended behavior in the branch: the Gmail messages search request includes `internalDate`, ISO output formats it through the shared date formatter, and the focused regressions `TestFormatGmailDateISO` and `TestGmailMessagesSearch_RequestsInternalDateForISOOutput` pass. The real branch binary builds and runs, the contributor supplied a live Gmail result demonstrating the corrected timestamp, and autoreview found no accepted/actionable findings.

Recommendation: **LAND**. This is a narrow fix for #945 with an appropriate regression test.

PR #947, opened 2026-07-29
feat(gmail): make inline attachment size limit configurable

ronny-rentner:
## Summary

`gmail attachment --inline` embeds the attachment content as base64 (`contentBase64`) in the response, but the size it will embed is capped at a hardcoded 3 MiB. This makes that cap configurable while keeping the current default, so nothing changes unless you ask for it.

- New flag `--inline-max-bytes` on `gmail attachment` (default `3145728`, i.e. the current 3 MiB).
- Also settable via `GOG_GMAIL_INLINE_MAX_BYTES`.
- Attachments larger than the limit keep the existing behavior: they fall back to the file path with the same explanatory `reason`.

## Motivation

Automation that consumes attachments *only* through `--inline` — for example a confined agent that cannot read gog's `0600` output files directly — needs to pull attachments larger than 3 MiB without a separate file read. A fixed 3 MiB cap forces those callers to give up on anything bigger.

## User-facing changes

- New flag: `--inline-max-bytes <bytes>` on `gmail attachment`.
- New env var: `GOG_GMAIL_INLINE_MAX_BYTES`.
- Default is unchanged (3 MiB), so existing behavior is preserved unless the limit is raised.

## Testing

- `go test ./internal/cmd/` — the oversized-fallback test now drives the boundary crossing through the flag (`--inline-max-bytes`) rather than a hardcoded constant.
- `make lint` — 0 issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.34.2 - 2026-07-27
 
+- Gmail: add configurable inline attachment byte limits while retaining the 3 MiB default and path fallback. (#947) — thanks @ronny-rentner.
 - Gmail: preserve single-object result envelopes for attachment-bearing `--results-only` output. (#943) — thanks @hashtag1974.
 - Sheets: keep positional updates within the requested range, preserving comma-bearing single-cell values and accurate named-range dry runs. (#941) — thanks @cathrynlavery.
 - Gmail: add source-specific `internalDateIso` timestamps to message and thread listings while preserving the legacy sender-header `date`. (#945, #946) — thanks @chrischall.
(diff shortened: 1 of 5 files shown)


steipete:
Maintainer follow-up: I fixed the dry-run observability gap on the contributor branch in 94a4431de1a334ac9c22bac62e07822561d4c99c.

`gmail attachment --inline --inline-max-bytes N --dry-run` now reports `inline_max_bytes: N`, so automation can audit the caller-selected output/memory boundary before authentication or download. The unchanged default (3145728), explicit values, changed values, and negative rejection are all covered.

Proof:

- `go test ./internal/cmd -run 'TestExecute_GmailAttachment_Inline'` — pass
- Built CLI with limit 42 — exit 0 and dry-run JSON reports `inline_max_bytes: 42`
- Source-blind behavior contract — explicit 42, default 3145728, changed 17, and negative exit-2 clauses all pass

My recommendation is LAND. The default remains byte-for-byte compatible, the configurable decision is entirely local after the normal Gmail attachment fetch, and the real binary now makes the new boundary inspectable without credentials.

PR #948, opened 2026-07-29
feat(gmail): auto-select reply From from the addressed send-as alias

ronny-rentner:
## Summary

When replying without an explicit `--from`, this selects the send-as alias the original message was addressed to as the reply's From identity, instead of always sending from the account default.

Precedence:

1. explicit `--from` (unchanged — always wins)
2. the send-as alias the original was addressed to (To before Cc, first verified match)
3. the account default (unchanged fallback)

Only verified send-as aliases are eligible, and the alias is resolved through the existing `resolveComposeFrom` path, so it can never produce an invalid From. If the send-as lookup fails, the account default stays in place.

Applies to both `gmail reply` / `gmail reply-all` and reply drafts.

## Motivation

On an account with multiple verified send-as aliases, a reply currently always goes out as the account default — even when the original was addressed to a specific alias. Someone who writes to `alias@example.com` then gets a reply from `primary@example.com`. This keeps the reply on the identity the sender actually wrote to, which is what most people expect a "reply" to do.

## Note on default behavior

This changes default reply behavior (no new flag required to trigger it). I think matching the addressed alias is the least-surprising default, but if you'd rather gate it behind an opt-in flag or a config setting, I'm happy to adjust — flagging it here since it's a behavior change rather than a purely additive one.

## Testing

- Unit test for the picker (`pickSendAsFromRecipients`): To-wins, Cc-used, To-beats-Cc, first-match, case-insensitive match, no-match → empty, unverified alias skipped.
- End-to-end reply test: addressed-to-alias → alias used as From; no matching alias → account default; explicit `--from` → override honored.
- `go test ./internal/cmd/`, `make lint` — 0 issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.34.2 - 2026-07-27
 
+- Gmail: add opt-in reply From alias selection for the verified send-as alias addressed by the original message. (#948) — thanks @ronny-rentner.
 - Gmail: add configurable inline attachment byte limits while retaining the 3 MiB default and path fallback. (#947) — thanks @ronny-rentner.
 - Gmail: preserve single-object result envelopes for attachment-bearing `--results-only` output. (#943) — thanks @hashtag1974.
 - Sheets: keep positional updates within the requested range, preserving comma-bearing single-cell values and accurate named-range dry runs. (#941) — thanks @cathrynlavery.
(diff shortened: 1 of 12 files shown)


steipete:
Maintainer verification complete. I pushed `55cc2efd` to preserve the existing default while keeping the contributor's useful alias-selection implementation as an explicit opt-in.

The feature is now available as `--auto-from-addressed-alias` on `gmail reply`, `gmail reply-all`, `gmail drafts create`, and `gmail drafts update`. Omitting the flag keeps the account-default From behavior; an explicit `--from` still wins. The alias is selected before signature resolution, so the chosen identity and signature remain consistent.

Proof:

- Focused reply and draft regression tests pass, including default-off, opt-in alias selection, explicit override, fallback, and signature selection.
- A source-blind built-CLI contract passed for all four commands: default dry runs report `auto_from_addressed_alias: false`, opt-in dry runs report `true`, explicit `--from` remains visible, help exposes the flag, and missing-body validation still exits 2.
- Full `make ci` passes, including all Go tests, lint/deadcode/format checks, generated command docs, docs-site validation, agent-skill checks, and script tests.
- Autoreview is clean with no accepted/actionable findings.

Recommendation: **LAND**.

steipete:
@clawsweeper re-review

The current head now preserves the existing default and exposes alias selection through the explicit `--auto-from-addressed-alias` flag; please review the updated behavior and maintainer proof above.

PR #962, opened 2026-08-07
feat(gmail): add --include-attachments to messages search

ronny-rentner:
## Summary

Adds `--include-attachments` to `gmail messages search`. With it, each result carries its attachment metadata — filename, size, mimeType (and the `attachmentId` in JSON) — so you can see what a message holds without switching to `--include-body`.

Today the only way to see a message's attachments in search results is `--include-body`, which also decodes and renders the full body. `--include-attachments` gives just the attachment list.

- New flag `--include-attachments` (env `GOG_GMAIL_INCLUDE_ATTACHMENTS`).
- Emits the existing `attachmentOutput` shape (the same one `--include-body` already produces), so the JSON is purely additive — no existing field changes.
- Text output gains an `ATTACHMENTS` column (`filename (mimeType, size)`); JSON gains the `attachments` array (with the `attachmentId`).
- Fetches `format=full` — the same complete fetch `gmail get` / `thread` / `drafts` already use — so attachments at any MIME nesting depth are listed. The body data is transferred but never rendered.

## Motivation

Lets a caller enumerate what a message carries — and, from the JSON, get the `attachmentId` to download — without switching to `--include-body`.

## User-facing changes

- New flag `--include-attachments` on `gmail messages search`.
- New env var `GOG_GMAIL_INCLUDE_ATTACHMENTS` (mirrors the existing `GOG_GMAIL_INLINE_MAX_BYTES` pattern).
- When set: an `ATTACHMENTS` column in text output and an additive `attachments` array in JSON.

## Testing

- `TestExecute_GmailMessagesSearch_IncludeAttachments`: lists an attachment nested several MIME levels down (proving the full fetch), does not render the body, and asserts the fetch uses `format=full` with no capping parts mask; the default search lists neither.
- `make ci` clean (fmt / lint / deadcode / tests / docs).

## Verification

Redacted terminal output from the built binary against a live Gmail account (ids masked, `attachmentId` truncated; sizes, mimeType, sender, subject verbatim).

Text output — the `ATTACHMENTS` column:

```
$ gog gmail messages search "from:purchase-noreply@twitch.tv has:attachment after:2026/08/01" --include-attachments
ID                THREAD            DATE              FROM                        SUBJECT          LABELS                         ATTACHMENTS
19fd72XXXXXXXXXX  19fd72XXXXXXXXXX  2026-08-06 14:41  purchase-noreply@twitch.tv  Kaufbestätigung  UNREAD,CATEGORY_UPDATES,INBOX  invoice_EU-DEXXXXXXX.pdf (application/pdf, 21.1 KB)
```

JSON — same metadata plus the per-fetch `attachmentId`:

```
$ gog gmail messages search "from:purchase-noreply@twitch.tv has:attachment after:2026/08/01" --include-attachments --json | jq '.messages[0].attachments'
[
  {
    "filename": "invoice_EU-DEXXXXXXX.pdf",
    "size": 21609,
    "sizeHuman": "21.1 KB",
    "mimeType": "application/pdf",
    "attachmentId": "ANGjdJ_5g7aKXXXX… (426 chars)"
  }
]
```

### DIFF
--- a/docs/commands/gog-gmail-messages-search.md
+++ b/docs/commands/gog-gmail-messages-search.md
@@ -34,6 +34,7 @@ gog gmail (mail,email) messages (message,msg,msgs) search (find,query,ls,list) <
 | `--gmail-no-send` | `bool` | false | Block Gmail send operations (agent safety) |
 | `-h`<br>`--help` | `kong.helpFlag` |  | Show context-sensitive help. |
 | `--home` | `string` |  | Override gogcli config/data/state/cache root (equivalent to GOG_HOME) |
+| `--include-attachments` | `bool` |  | Include each message's attachment metadata |
 | `--include-body` | `bool` |  | Include decoded message body (JSON is full; text output truncates only unusually large bodies) |
 | `-j`<br>`--json`<br>`--machine` | `bool` | false | Output JSON to stdout (best for scripting) |
 | `--local` | `bool` |  | Use local timezone (default behavior, useful to override --timezone) |
--- a/internal/cmd/execute_gmail_messages_include_attachments_test.go
+++ b/internal/cmd/execute_gmail_messages_include_attachments_test.go
@@ -1,0 +1,99 @@
+package cmd
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+func TestExecute_GmailMessagesSearch_IncludeAttachments(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		path := r.URL.Path
+		switch {
+		case strings.Contains(path, "/users/me/messages") && !strings.Contains(path, "/users/me/messages/"):
+			w.Header().Set("Content-Type", "application/json")
+			_ = json.NewEncoder(w).Encode(map[string]any{
+				"messages": []map[string]any{{"id": "m1", "threadId": "t1"}},
+			})
+		case strings.Contains(path, "/users/me/messages/m1"):
+			w.Header().Set("Content-Type", "application/json")
+			_ = json.NewEncoder(w).Encode(map[string]any{
+				"id": "m1", "threadId": "t1", "labelIds": []string{"INBOX"},
+				"payload": map[string]any{
+					"mimeType": "multipart/mixed",
+					"headers": []map[string]any{
+						{"name": "From", "value": "Example <no-reply@example.com>"},
+						{"name": "Subject", "value": "Receipt"},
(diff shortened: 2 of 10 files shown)


ronny-rentner:
On the "render MIME type and identifier in text output" finding: mimeType is now in the `ATTACHMENTS` column; the `attachmentId` deliberately is not. Two reasons:

**1. The full id doesn't fit a table column.** Gmail's `attachmentId` is ~426 bytes. Dropping that into a cell blows the table apart — it's unusable next to `filename`, `mimeType`, and `size`.

**2. A short id isn't feasible either.** gogcli already uses a short (first-8-char) form of the `attachmentId` to tag saved filenames, so the obvious idea was to show that same short id in the column. But the `attachmentId` is a **per-request opaque token** — Gmail mints a fresh ~426-char value on every `messages.get`, and all of them stay valid to download by. Verified against a live account, three fetches of the *same* message returned three *different* ids (leading bytes shown, the rest masked):

```
ANGjdJ-pXXXX…
ANGjdJ9oXXXX…
ANGjdJ9vXXXX…
```

and an id captured in one call still downloaded fine in a separate later call. Because the id changes on every call, its short form changes on every search too — it would be neither stable nor a match for a later download's filename, so it conveys nothing a caller can act on.

So the column shows `filename (mimeType, size)`. The full per-fetch id stays in `--json` (valid for that fetch), which is where retrieval belongs; the stable, compact reference is the 0-based index in the stacked #963.

steipete:
Landed in `9842e64b7682e748dd9a34c3df361fbb8c8298ca`; maintainer changelog follow-up is `1e9a712c9`.

Maintainer fix before landing:

- Gmail-controlled attachment filenames and MIME values now have tabs, line breaks, and other control characters replaced at the text/TSV presentation boundary. JSON metadata remains unchanged.

Proof:

- `go test ./internal/cmd -run 'TestExecute_GmailMessagesSearch_IncludeAttachments|TestGmailPresentationSchemas' -count=1` — passed, including the control-character regression case.
- Signed branch binary exercised read-only against a real Gmail account: the default search header had no `ATTACHMENTS` column; `--include-attachments` added it; flag and `GOG_GMAIL_INCLUDE_ATTACHMENTS=1` JSON each returned one real attachment with `attachmentId`, `filename`, `mimeType`, `size`, and `sizeHuman`; no decoded `body` field was emitted.
- GitHub CI run [31291954680](https://github.com/openclaw/gogcli/actions/runs/31291954680) passed Linux, worker, Windows, and Darwin jobs; Docker run [31291954693](https://github.com/openclaw/gogcli/actions/runs/31291954693) passed.

Thanks @ronny-rentner for the careful iteration and live evidence.

PR #963, opened 2026-08-07
feat(gmail): reference attachments by index (--use-indexed-attachment-ids)

ronny-rentner:
> **Stacked on #962.** This builds on `feat/gmail-include-attachments`; that PR's commit is shown here too and will drop out once #962 merges. Review/merge #962 first.

## Summary

Lets Gmail attachments be referenced by a stable **0-based index** instead of the long, opaque `attachmentId`, behind a single opt-in flag.

`--use-indexed-attachment-ids` (env `GOG_GMAIL_USE_INDEXED_ATTACHMENT_IDS`) is added to the six commands that render or accept an attachment id: `messages search`, `get`, `thread get`, `thread attachments`, `drafts get`, and `attachment`. When it's on:

- **Output** — those commands emit `attachmentIndex` (a number) in place of `attachmentId` (a ~200-char opaque string), in both JSON and text. The long id is not emitted at all.
- **Download input** — `gmail attachment <messageId> <index>` takes the 0-based index; the index resolves to the real attachment (the real id still drives the actual download API call). A non-numeric argument is rejected in this mode.
- **Saved filename** — the default download filename uses the index (`<messageId>_<index>_...`).

Default behaviour is unchanged: without the flag the argument and the output are the real `attachmentId`, and a numeric argument is treated as a real id, not an index (the index and the id are never silently mixed).

## Motivation

The `attachmentId` is a ~200-char opaque base64url string, so emitting one per attachment in a listing — and passing it back to download — is token-expensive for an agent. A 0-based position is 1–2 characters and is a stable reference: a message's MIME structure is fixed, so the Nth attachment is the same across calls, and `gmail attachment` re-resolves the index to the current `attachmentId` at download time. With the flag on, an agent lists attachments and downloads them purely by index and never handles a long id.

## User-facing changes

- New flag `--use-indexed-attachment-ids` (env `GOG_GMAIL_USE_INDEXED_ATTACHMENT_IDS`) on the six attachment-handling gmail commands.
- In that mode: output carries `attachmentIndex` (number) instead of `attachmentId`; `gmail attachment` takes the index; the saved filename uses the index.

## Implementation notes

- `attachmentInfo` carries both `AttachmentID` and a 0-based `AttachmentIndex`, stamped once when attachments are collected (`collectAttachments`).
- Output types (`attachmentOutput` and the download-result summaries) have **distinct** `attachmentId` (string, omitempty) and `attachmentIndex` (number, omitempty) fields; a stringified index is never written into an id field.
- The one index↔id resolution happens once, at the `gmail attachment` argument. The env var applies the setting globally through each command's env-defaulted flag, mirroring gogcli's existing `--include-passwords` / `GOG_ZOOM_INCLUDE_PASSWORDS`.

## Testing

- `attachmentByIndex` (valid / out-of-range / negative); index resolves and downloads; index out of range errors.
- Indexed mode: output surfaces `attachmentIndex` (unit + `gmail get` end-to-end), a non-numeric argument is rejected, the filename uses the index; a numeric argument **without** the flag is treated as a raw id (download fails); env-var enables the mode.
- `make fmt` / `make lint` clean; command docs regenerated. Verified against a live account.

### DIFF
--- a/docs/commands/gog-gmail-attachment.md
+++ b/docs/commands/gog-gmail-attachment.md
@@ -40,6 +40,7 @@ gog gmail (mail,email) attachment <messageId> <attachmentId> [flags]
 | `--readonly` | `bool` | false | Block mutating API requests at runtime; auth add also requests read-only OAuth scopes |
 | `--results-only` | `bool` |  | In JSON mode, emit only the primary result (drops envelope fields like nextPageToken) |
 | `--select`<br>`--pick`<br>`--project` | `string` |  | In JSON mode, select comma-separated fields (best-effort; supports dot paths). Desire path: use --fields for most commands. |
+| `--use-indexed-attachment-ids` | `bool` |  | Use 0-based indexes as attachment ids everywhere (output, the download argument, and saved filenames) |
 | `-v`<br>`--verbose` | `bool` |  | Enable verbose logging |
 | `--version` | `kong.VersionFlag` |  | Print version and exit |
 | `--wrap-untrusted` | `bool` | false | In JSON/raw output, wrap fetched text fields in external untrusted-content markers |
--- a/docs/commands/gog-gmail-drafts-get.md
+++ b/docs/commands/gog-gmail-drafts-get.md
@@ -37,6 +37,7 @@ gog gmail (mail,email) drafts (draft) get (info,show) <draftId> [flags]
 | `--readonly` | `bool` | false | Block mutating API requests at runtime; auth add also requests read-only OAuth scopes |
 | `--results-only` | `bool` |  | In JSON mode, emit only the primary result (drops envelope fields like nextPageToken) |
 | `--select`<br>`--pick`<br>`--project` | `string` |  | In JSON mode, select comma-separated fields (best-effort; supports dot paths). Desire path: use --fields for most commands. |
+| `--use-indexed-attachment-ids` | `bool` |  | Use 0-based indexes as attachment ids everywhere (output, the download argument, and saved filenames) |
 | `-v`<br>`--verbose` | `bool` |  | Enable verbose logging |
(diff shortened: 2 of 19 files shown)


steipete:
Landed in `8b44d0a81f9f8fd06f76d0150539d824e1a11449`; maintainer changelog follow-up is `8b33065cb`.

Maintainer work before landing:

- Merged current `main` so the stacked branch retained #962’s unlimited-depth attachment discovery, text column, and control-character sanitization.
- Validated indexed references before dry-run/authentication and emitted `attachment_index` in valid dry-run plans.
- Added usable curated message/index mappings to raw thread and draft JSON before stripping opaque attachment IDs.
- Added focused regression coverage for dry-run, draft, thread, and the stacked #962 behavior.

Proof:

- Focused attachment, message-search, presentation, draft, and thread tests passed.
- Signed current-head binary exercised against real Gmail with account/content identifiers redacted: default search emitted `attachmentId`; indexed search emitted `attachmentIndex: 0`; indexed message and thread JSON contained usable mappings and zero opaque-ID fields; `gmail attachment <messageId> 0` downloaded 34,710 bytes successfully.
- A clearly named temporary draft with one synthetic attachment returned index 0 and no opaque ID from indexed `drafts get`; it was deleted after proof. The first harness pass parsed the create response under the wrong key, so the cleanup trap missed it; that exact draft was immediately deleted before the corrected end-to-end rerun, whose cleanup also passed.
- GitHub CI run [31292547223](https://github.com/openclaw/gogcli/actions/runs/31292547223) passed Linux, worker, Windows, and Darwin jobs; Docker run [31292547230](https://github.com/openclaw/gogcli/actions/runs/31292547230) passed.

Thanks @ronny-rentner for the compact-reference design and persistent iteration.

PR #964, opened 2026-08-07
feat(gmail): env-configure --auto-from-addressed-alias via GOG_GMAIL_AUTO_FROM_ADDRESSED_ALIAS

ronny-rentner:
## Summary

Adds `GOG_GMAIL_AUTO_FROM_ADDRESSED_ALIAS` as an environment default for the existing `--auto-from-addressed-alias` option, so agents and wrappers can enable the policy without appending a flag to every invocation.

The environment default applies to the four command paths that already own the option and sender-selection behavior:

- `gmail reply`
- `gmail reply-all`
- `gmail drafts create`
- `gmail drafts update`

It does **not** add support to `gmail forward`. No behavior changes when the variable is unset, and an explicit `--from` retains its existing precedence.

## Motivation

A bridge that consistently wants replies to use the verified send-as alias addressed by the original message currently has to pass `--auto-from-addressed-alias` on every supported command. The environment default follows gogcli's existing env-backed Gmail option conventions without creating a new sender-selection path.

## Testing

- Added `TestGmailAutoFromAddressedAliasEnvDefaults`, covering env parsing for all four supported commands.
- Existing sender-selection tests cover addressed-alias selection, account-default fallback, and explicit `--from` precedence.
- `make ci` passes.

## Built-binary proof

The signed current-head binary was exercised source-blind with JSON dry runs:

- Env enabled: all four plans reported `auto_from_addressed_alias: true`.
- Env unset: all four plans reported `false`.
- Env enabled plus explicit `--from`: the plan retained both the enabled option and explicit sender input.
- No Google API call or message send occurred.

Provider-level alias proof is unavailable on the current test host because none of its healthy authenticated Gmail accounts exposes a verified non-primary send-as alias. The provider selection path itself is unchanged from the existing explicit flag.

### DIFF
--- a/internal/cmd/gmail_auto_from_env_test.go
+++ b/internal/cmd/gmail_auto_from_env_test.go
@@ -1,0 +1,41 @@
+package cmd
+
+import (
+	"encoding/json"
+	"testing"
+)
+
+func TestGmailAutoFromAddressedAliasEnvDefaults(t *testing.T) {
+	t.Setenv("GOG_GMAIL_AUTO_FROM_ADDRESSED_ALIAS", "1")
+
+	tests := []struct {
+		name string
+		args []string
+	}{
+		{name: "reply", args: []string{"gmail", "reply", "m1", "--body", "hello"}},
+		{name: "reply all", args: []string{"gmail", "reply-all", "m1", "--body", "hello"}},
+		{name: "draft create", args: []string{"gmail", "drafts", "create", "--to", "you@example.com", "--subject", "subject", "--body", "hello"}},
+		{name: "draft update", args: []string{"gmail", "drafts", "update", "d1", "--subject", "subject", "--body", "hello"}},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			args := append([]string{"--json", "--dry-run"}, tc.args...)
+			result := executeWithTestRuntime(t, args, nil)
+			if result.err != nil {
+				t.Fatalf("Execute: %v\nstderr=%q", result.err, result.stderr)
+			}
+			var output struct {
+				Request struct {
+					AutoFromAddressedAlias bool `json:"auto_from_addressed_alias"`
+				} `json:"request"`
+			}
+			if err := json.Unmarshal([]byte(result.stdout), &output); err != nil {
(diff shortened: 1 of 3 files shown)


steipete:
Current-head signed-binary proof (`b2068d8cd2baf361c6179968061fda120e6c0971`), with no account or message data involved:

```text
env_on_plans=4/4:true
env_unset_plans=4/4:false
explicit_from_preserved_with_env=true
google_api_calls=0; messages_sent=0
provider_alias_proof=blocked:no-verified-non-primary-alias; existing-flag-path-unchanged
behavior-validator: clauses 1-4 PASS; env-on/env-off and explicit-input probes PASS
```

Commands exercised the real built CLI with JSON `--dry-run` for `gmail reply`, `gmail reply-all`, `gmail drafts create`, and `gmail drafts update`, parsing each request object rather than trusting exit status. Focused service tests cover the unchanged addressed-alias selection, default fallback, and explicit-sender precedence paths.

Full local `make ci` and final Codex autoreview are clean. GitHub CI run [31293103238](https://github.com/openclaw/gogcli/actions/runs/31293103238) and Docker run [31293103228](https://github.com/openclaw/gogcli/actions/runs/31293103228) are green.

@clawsweeper re-review

steipete:
Landed in `d06fe6dc07202ac297f3b66bc01d182eb556e5c5`; maintainer changelog follow-up is `e32b1f73a`.

The permanent scope is the four existing option owners only: reply, reply-all, draft create, and draft update. The PR body was corrected to remove the unsupported forward claim, and a four-command env-default regression test was added.

Verification:

- Signed current-head binary JSON dry runs: env enabled produced `auto_from_addressed_alias: true` on all four commands; env unset produced `false`; explicit `--from` remained present with the env option enabled; zero API calls and zero sends.
- Provider-level alias selection could not be repeated because no healthy authenticated local account exposes a verified non-primary alias; that selection path is unchanged from the already-shipped explicit flag and remains covered by focused service tests.
- GitHub CI run [31293103238](https://github.com/openclaw/gogcli/actions/runs/31293103238) and Docker run [31293103228](https://github.com/openclaw/gogcli/actions/runs/31293103228) passed every job.

Thanks @ronny-rentner.

PR #965, opened 2026-08-07
feat(docs comments): resolve which tab a comment belongs to in one fetch

### LINKED ISSUE #961: feat(docs comments): resolve which tab a comment belongs to — `--tab` filter / `--locate` on `docs comments list`
### Motivation

In a multi-tab Google Doc, `gog docs comments list` returns every comment in the file with no indication of which tab each one belongs to. The Drive comments API is file-scoped, and the `anchor` field is an opaque `kix.*` token (verified against a live doc), so tab membership cannot be derived from the list payload alone.

`docs comments locate` (#687) already resolves a comment's `quotedFileContent.value` to `{startIndex, endIndex, paragraphIndex, tabId}` across all tabs — but only one comment per invocation, and each invocation performs its own `documents.get(includeTabsContent=true)`. Annotating a 30-comment doc therefore costs 30 full document fetches.

Concrete use case: reviewing feedback left on one tab of a large multi-tab planning doc. I want "the open comments on this tab, with their quoted text" in a single command.

### Repro / today's workaround

```bash
DOC=<doc id>
gog docs comments list "$DOC" --all -j --results-only \
  | jq -r '.[].id' \
  | while read -r id; do
      gog docs comments locate "$DOC" "$id" -j   # one full documents.get per comment
    done
```

…then join the two payloads by comment ID in jq.

### Proposed surface

Two shapes; I don't have a read on which one fits this CLI's taste better.

**Option A — extend `docs comments list`** (opt-in flags, default behaviour unchanged):

```
gog docs comments list <docId> [--locate] [--tab=<title|id>]
```

- `--locate`: attach a per-comment `location` object (`{matches, orphaned}`), resolved from a single `documents.get(includeTabsContent=true)` shared by every comment in the page
- `--tab`: implies `--locate` and filters to comments with at least one match in the target tab

**Option B — extend `docs comments locate`**:

```
gog docs comments locate <docId> --all [--tab=<title|id>]
```

- resolves every comment in one document fetch and emits an array of the existing `docsCommentLocateResult` shape
- callers join it with `comments list` themselves when they also need author/content/replies

A is more convenient for the "show me this tab's review feedback" workflow; B is the smaller, more orthogonal change. Either works for me.

### Notes / constraints I ran into

- **Tab attribution can only be quote-based.** `anchor` is a `"kix.ge89m06dp6ga"`-style opaque value carrying no tab or index information, so this reuses the same matching path as `locate` and inherits its semantics: a quote appearing in several tabs yields multiple matches, and an edited or deleted quote yields `orphaned: true` with zero matches. Proposal: `--tab` passes a comment when *any* match falls in the target tab, and `--locate` always reports the full `matches` array so callers can detect ambiguity themselves.
- **Extra scope.** `docs comments list` currently builds only the Drive service; both options additionally need the Docs service (`documents.readonly`). That is why I'd keep this strictly opt-in — users authorised for Drive only keep working exactly as today unless they pass a new flag.
- **Cost.** One extra `documents.get` per command invocation regardless of comment count, versus N today.

### Acceptance criteria

- `--locate` annotates each comment with the same range data `docs comments locate` returns for that comment, using a single document fetch.
- `--tab` accepts a tab title or tab ID through the existing `resolveTabArg` path and filters the list accordingly.
- Comments whose quote no longer resolves are reported as orphaned rather than dropped silently (and are excluded by `--tab`, since they resolve to no tab).
- A default invocation (no new flags) is byte-identical to today and still requires only the Drive scope.
- `--json`, `--plain`, and the expanded table all carry the resolved tab.

### Possible follow-up (out of scope here)

`--context <n>` to include the surrounding paragraph(s) of the quoted text — cheap once ranges are resolved, but a separate change.

### Question before I start

I'm happy to send a PR for this, with tests, `make ci`, `make docs-commands`, and a CHANGELOG entry. Before I do:

1. Is this wanted at all, or does it overlap with something already planned?
2. Option A or Option B?

If you'd rather shape this yourself, I'm equally happy to leave it as an issue — just let me know either way.

### References

- Drive `comments.list` (`quotedFileContent`, `anchor`): https://developers.google.com/workspace/drive/api/reference/rest/v3/comments/list
- #687 — `docs comments locate`
- #697 — `docs raw --tab`

ryo-touch:
closes #961

## What

Adds two opt-in flags to `gog docs comments list`:

- `--locate` attaches a per-comment `location` object (`matches`, `orphaned`), resolved from a **single** `documents.get(includeTabsContent=true)` shared by every comment.
- `--tab <title|id>` implies `--locate` and keeps comments with at least one match in that tab. Resolved through the existing `resolveTabArg` path, so the hidden deprecated `--tab-id` works like it does on the sibling commands.

Today `docs comments locate` answers this for one comment per invocation, each paying its own document fetch — annotating a 23-comment doc costs 23 fetches.

```
$ gog docs comments list <docId> --tab "Planning" --all --plain
TYPE     ID           AUTHOR  QUOTED  CONTENT  CREATED  RESOLVED  ACTION  TAB
comment  AAACANyRdjw  ...                                                 Planning
```

```jsonc
{
  "docId": "…",
  "tab": { "id": "t.abc", "title": "Planning" },   // only with --tab
  "comments": [
    { "id": "AAA…", "quotedFileContent": {…},
      "location": { "matches": [{"startIndex": 1, "endIndex": 19, "paragraphIndex": 0, "tabId": "t.abc"}],
                    "orphaned": false } }
  ],
  "nextPageToken": ""
}
```

## Design notes

- **Strictly opt-in.** Without the new flags not a line of the new code runs: same Drive-only path, same bytes on stdout, no Docs scope. A test installs a Docs service factory that calls `t.Fatal` to keep it that way.
- **Ambiguity stays visible.** `matches` always spans every tab, never truncated to the target tab, so a quote appearing in two tabs is detectable by the caller. `--tab` passes a comment when *any* match lands in the tab.
- **`--tab` drops anything that resolves to no tab** — orphaned quotes and comments with no quoted text (e.g. document-level ones). `--locate` alone reports them with `orphaned: true` rather than dropping them.
- **Exit codes unchanged.** `list --locate` deliberately does not inherit `locate`'s `exitCodeOrphaned` (11); orphaned is per-item state here. Only `--fail-empty` (3) and the usual auth codes apply.
- **Pagination.** With `--tab` and no `--all`, filtering can empty a page, so the command walks pages until one yields a match — mirroring the existing scan-for-open loop in `listDriveComments`, and reusing the one already-fetched document (with a seen-token guard against a repeating `nextPageToken`).
- `driveCommentWithLocation` embeds `*drive.Comment` **and overrides `MarshalJSON`**: the generated type declares `MarshalJSON` on a value receiver, so the promoted method would silently drop `location`. Same pattern as `eventWithCalendar` in `calendar_list.go`; there's a unit test that fails if the override is removed.

## Option A vs B

The issue asked which shape you'd prefer and I didn't want to sit on the patch, so this is **Option A**. The batch resolution lives behind `docsCommentLocator` in its own file, so **Option B (`docs comments locate --all`) would only need a different command layer** — no change to the resolver. Happy to reshape it, or to close this if you'd rather own the design.

## Testing

- `make ci` green.
- 11 new tests: default-path-unchanged (asserts no Docs service is created), single-fetch proof, orphaned/unquoted handling, `--tab` filtering and implied `--locate`, cross-tab matches preserved, page scanning, `--fail-empty`, unknown tab, tabless documents, plain-table columns, and the `MarshalJSON` regression guard.
- Verified the default table/plain rendering is byte-identical to `main` by running the same fixture (replies, truncation, nil entries) through both implementations.
- **Live**, against a 20+ tab doc with 23 comments: `--locate` resolved all 23 in one fetch (21 located across 5 tabs, 2 orphaned, 2 matching in more than one tab), and every `matches` array was identical to running `docs comments locate` per comment. `--tab` by ID and by (Japanese) title returned the same 5 comments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Docs: add opt-in `--locate` and `--tab` to `docs comments list`, resolving every comment's tab and index ranges from a single document fetch. (#961)
 - Gmail: add guarded single-message RFC822/EML import from a file or stdin, with labels, internal-date, spam, calendar-processing, and parse-only dry-run controls. (#956) — thanks @holgergruenhagen.
 - Gmail: warn before a draft update replaces an existing rich-text body with plain text only, while keeping JSON stdout clean. (#955) — thanks @mcinteerj.
 - Dependencies: update the Google API and OpenTelemetry stacks, Go developer tools, pnpm, and email-tracking worker toolchain to their latest policy-eligible releases.
--- a/docs/commands/gog-docs-comments-list.md
+++ b/docs/commands/gog-docs-comments-list.md
@@ -34,6 +34,7 @@ gog docs (doc) comments list (ls) <docId> [flags]
 | `--home` | `string` |  | Override gogcli config/data/state/cache root (equivalent to GOG_HOME) |
 | `--include-resolved`<br>`--resolved` | `bool` |  | Include resolved comments (default: open only) |
 | `-j`<br>`--json`<br>`--machine` | `bool` | false | Output JSON to stdout (best for scripting) |
+| `--locate` | `bool` |  | Attach each comment's tab and index ranges (one extra Docs fetch) |
 | `--max`<br>`--limit` | `int64` | 100 | Max results per page |
 | `--no-input`<br>`--non-interactive`<br>`--noninteractive` | `bool` |  | Never prompt; fail instead (useful for CI) |
 | `--page`<br>`--cursor` | `string` |  | Page token for pagination |
@@ -42,6 +43,7 @@ gog docs (doc) comments list (ls) <docId> [flags]
 | `--results-only` | `bool` |  | In JSON mode, emit only the primary result (drops envelope fields like nextPageToken) |
 | `--select`<br>`--pick`<br>`--project` | `string` |  | In JSON mode, select comma-separated fields (best-effort; supports dot paths). Desire path: use --fields for most commands. |
 | `--since` | `string` |  | Only return comments modified at or after this RFC3339 timestamp |
+| `--tab` | `string` |  | Only comments located in this tab by title or ID (implies --locate) |
 | `-v`<br>`--verbose` | `bool` |  | Enable verbose logging |
 | `--version` | `kong.VersionFlag` |  | Print version and exit |
 | `--wrap-untrusted` | `bool` | false | In JSON/raw output, wrap fetched text fields in external untrusted-content markers |
--- a/docs/docs-editing.md
+++ b/docs/docs-editing.md
@@ -135,6 +135,25 @@ Use `--occurrence N` when an anchor is ambiguous and `--match-case` when case
 must be exact. `docs comments locate` applies the same matching rules to a
 comment's quoted text and reports its tab plus UTF-16 range.
 
+`docs comments list --locate` does the same for every listed comment from a
+single document fetch, adding a `location` object (`matches`, `orphaned`) and a
+`TAB` column. `--tab` implies `--locate`, adds the resolved tab to JSON output
+as a top-level `tab` (`{id, title}`), and keeps only the comments with at least
+one match in that tab:
+
+```bash
+gog docs comments list <docId> --tab "Planning" --all --json
+```
+
+Anything that resolves to no tab therefore drops out under `--tab`: comments
+whose quote was edited away (`orphaned`), and comments with no quoted text at
+all, such as ones attached to the whole document. Use `--locate` alone to see
+them - it reports every comment, orphaned or not.
+
+Both flags are opt-in: without them the command still reads Drive alone.
+`matches` always spans every tab, so a quote that appears more than once stays
+visible as ambiguous.
+
 `insert` and `update` both accept `--markdown` to convert the content (headings,
 fenced code blocks, lists, tables, images) before placing it at the resolved
 position. `insert --markdown` adds the block without deleting anything; `update
--- a/docs/docs-editing.md
+++ b/docs/docs-editing.md
@@ -152,7 +152,8 @@ them - it reports every comment, orphaned or not.
 
 Both flags are opt-in: without them the command still reads Drive alone.
 `matches` always spans every tab, so a quote that appears more than once stays
-visible as ambiguous.
+visible as ambiguous. An explicitly empty `--tab` value is rejected instead of
+silently falling back to an unfiltered listing.
 
 `insert` and `update` both accept `--markdown` to convert the content (headings,
 fenced code blocks, lists, tables, images) before placing it at the resolved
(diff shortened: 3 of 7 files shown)


steipete:
Maintainer decision: adopt Option A (`docs comments list --locate/--tab`). The combined listing is the right ownership boundary for the stated review-feedback workflow, and both flags remain strictly opt-in.

Current-head signed-binary proof (`833dd764ce0d08c52daa6810890f2a62107826bf`), with account/document/tab/comment identities and all content redacted:

```text
document_tabs=3; default_comments=15; located_comments=15; matched_comments=3
default_location_keys=0; located_location_keys=15
tab_filter_count=3; all_results_touch_target_tab=true
plain_tab_column=true
explicit_empty_tab_rejected=true; explicit_empty_tab_id_rejected=true; pre_auth=true
one_docs_fetch_invariant=focused-http-test-pass
behavior-validator: clauses 1-6 PASS; default/locate comparison and derived-tab filter probes PASS
```

The real `--tab` value was derived from a located match, then every filtered comment was checked to contain that tab in its full match array. The focused HTTP-level test separately proves one `documents.get(includeTabsContent=true)` across all comments and page scanning.

Full local `make ci` and final Codex autoreview are clean. The contributor-owned changelog edit was removed; the maintainer will add the credited entry after landing.

@clawsweeper re-review

steipete:
Landed in `6ff1627e14ea5320569afab3c61a92f1987ce58b`; maintainer changelog follow-up is `b57df1c41`.

Maintainer work before landing:

- Adopted Option A (`docs comments list --locate/--tab`).
- Merged current `main`, removed the contributor-owned changelog edit for maintainer follow-up, and rejected explicit empty `--tab` / `--tab-id` values before service creation.
- Added regression coverage and documentation for that validation.

Proof:

- Signed current-head binary ran read-only against a real three-tab document with identities/content redacted: default listing returned 15 comments with zero location keys; `--locate` returned the same 15 with 15 location objects and three real range matches; a tab derived from those matches returned three comments, all touching the selected tab; plain output added `TAB`.
- Focused HTTP-level tests prove exactly one Docs fetch across all comments and page scanning, while the default path never creates a Docs service.
- Empty `--tab` and `--tab-id` values failed before auth/service access.
- Final Codex autoreview: clean, no accepted/actionable findings; ClawSweeper marked proof sufficient with no patch findings.
- First-time contributor gates were approved for exact runs [31293888385](https://github.com/openclaw/gogcli/actions/runs/31293888385) and [31293888410](https://github.com/openclaw/gogcli/actions/runs/31293888410); Linux, worker, Windows, Darwin, and Docker all passed.

Thanks @ryo-touch for the unusually thorough proposal, tests, and original live validation.

ryo-touch:
@steipete Thank you for reviewing and merging this pr. I am looking forward to release!

Issue #889, opened 2026-06-30
Intermittent 403 'insufficient scopes' that succeeds on retry with same token (per-call scope mint?)


ortonom:
## Summary

`gog` intermittently returns a **403 `insufficient scopes`** (`Request had insufficient authentication scopes`) on a call that then **succeeds on an immediate retry** with the **same stored refresh token, no re-auth and no consent change**. The error wording strongly implies a stale login / missing scope, which leads operators to re-authenticate — but re-auth changes nothing, because the scope was never actually missing.

## Observed

- Build: `0.31.1` (Homebrew), keyring backend `keychain`, single OAuth account.
- Two different services hit it in one session:
  - `gog calendar create <cal> --summary … --with-meet` → 403 `insufficient scopes`. Retried minutes later, **same token**, → success (Meet conference created).
  - `gog contacts search <q>` → 403 `insufficient authentication scopes`. Retried later, **same token**, → results returned.
- Sibling operations on the *same* service worked throughout (e.g. bare `calendar create` without `--with-meet`, and `contacts list`), so the token plainly held the scopes.
- Debug logs around the failing calls show per-call scoped-token minting:
  ```
  level=DEBUG msg="creating client options with custom scopes" serviceLabel=calendar …
  level=DEBUG msg="client options with custom scopes created successfully" …
  ```

## Hypothesis

The per-service custom-scope access-token mint occasionally yields a token that the API rejects as under-scoped on first use, and a fresh mint on retry resolves it (a mint/propagation race). The 403 is transient, not a genuine scope gap.

## Question / request

Should the per-service token-mint path **retry once internally** on a 403 `insufficient scopes` (or refresh-then-retry) before surfacing it? At minimum, could the surfaced error distinguish "token genuinely lacks scope X" from "transient mint failure" so callers don't reflexively re-authenticate?

Happy to capture fuller debug output if useful.

Issue #939, opened 2026-07-26
macOS 27 Tahoe beta: Keychain C API silently writes 0 bytes — file keyring workaround required


inamiy:
## Summary

On macOS 27.0 (Tahoe beta, build 26A5378n), `gog` cannot store tokens in macOS Keychain. The Keychain C API call (`SecKeychainAddGenericPassword` / `SecKeychainItemModifyContent`) appears to succeed (returns `errSecSuccess`), but the password data blob is silently empty (0 bytes) on read-back. `gog`'s built-in `verifiedSet` verification step catches this and reports:

```
Failed to store token: store token: token verification failed: keyring wrote 0 bytes: could not read back token after write: The specified item could not be found in the keyring
Workaround: switch to file-based keyring with: gog auth keyring file
```

## Environment

- **macOS**: 27.0 (build 26A5378n) — Tahoe beta
- **gog**: v0.32.0 (f20a26cd 2026-07-03T18:45:26Z) — installed via Homebrew
- **Architecture**: Apple Silicon (arm64)
- **Keyring backend**: `auto` (macOS Keychain / `99designs/go-keychain`)

## Reproduction

1. Ensure keychain backend is set to `auto`:
   ```bash
   gog auth keyring auto
   ```

2. Delete any stale gogcli keychain entries:
   ```bash
   security delete-generic-password -s "gogcli" -a "client/default/client-secret" 2>/dev/null
   security delete-generic-password -s "gogcli" -a "token:default:<email>" 2>/dev/null
   ```

3. Store OAuth credentials:
   ```bash
   gog auth credentials set <client_secret.json>
   ```
   → Fails with: `Secret not found in keyring (refresh token missing). Run: gog auth add <email>`

4. Check keychain — entry was created but password data is inaccessible:
   ```bash
   security find-generic-password -s "gogcli" -a "client/default/client-secret" -w
   ```
   → Exit code 36 (`errSecItemNotFound`) — the entry exists in `security dump-keychain` output but cannot be read by query.

5. Attempt browser login:
   ```bash
   gog auth manage
   ```
   → Browser shows: `Failed to store token: store token: token verification failed: keyring wrote 0 bytes`

## Key finding: `security` CLI works, Go C API doesn't

Apple's `security` CLI can write and read the same keychain entries perfectly:

```bash
security add-generic-password -s "gogcli" -a "test-account" -w "test-data"
security find-generic-password -s "gogcli" -a "test-account" -w
# → test-data (exit 0)
```

[...]

PR #973, opened 2026-08-09
feat(auth): auto-reauth on expired/revoked refresh tokens (invalid_grant)

inamiy:
>

## Summary

When the stored OAuth refresh token is expired or revoked (`invalid_grant`), `gog` currently fails with a hard error and requires the user to manually re-run `gog auth add`. This is in contrast to other CLI tools (e.g. Rust's `yup-oauth2` `InstalledFlowAuthenticator`) which automatically fall back to a browser-based re-authorization flow when the refresh token is invalid, making the failure transparent to the user.

This PR adds auto-reauth support to `gog`.

## How it works

```
API call → oauth2.Transport → refresh token exchange
  → invalid_grant (expired/revoked)
    → RetryTransport detects error
      → --no-input or non-TTY? → surface clear error: "run 'gog auth add'"
      → interactive? → call Reauth()
        → browser opens OAuth flow (with --force-consent)
        → new refresh token persisted to keyring
        → in-memory token source (resettableOAuthTokenSource) rebuilt with new token
        → retry original API request
```

## Key design decisions

- **`--force-consent` always used** in auto-reauth to ensure Google returns a new refresh token
- **Only retries once** (`retriedReauth` flag) — no infinite loops
- **Respects `--no-input` and non-TTY stdin** — CI/piped environments get a clear error with `gog auth add` hint, not a browser
- **Excluded for ADC, service accounts, and direct access tokens** — gated by concrete type assertion (`*persistingTokenSource`)
- **Scope preservation** — reauth loads the stored token's full scope/service set, preventing silent grant narrowing (e.g. a calendar command narrowing a gmail+calendar+drive grant)
- **Email verification** — authorized identity is compared with the expected email before persisting
- **Token source reset** — after successful reauth, `resettableOAuthTokenSource.ResetRefreshToken` rebuilds the source with the new refresh token, so the retried request doesn't reuse the revoked token

## Prior art

- `yup-o-auth2` (Rust): `InstalledFlowAuthenticator::find_token_info()` — on refresh failure, falls back to `auth_flow.token()` which opens a browser. Source: [authenticator.rs](https://github.com/dermesser/yup-o-auth2/blob/master/src/authenticator.rs)
- gogcli already has partial auth resilience:
  - v0.31.0: Recover from corrupt token payloads (#872)
  - v0.32.0: Retry on 403 insufficient scopes by refreshing credentials (#889)
  - v0.33.0: Trust Developer-ID-signed binaries for Keychain access
  - Auto-reauth on `invalid_grant` is the next gap in this progression.

## Files changed

| File | Description |
|---|---|
| `internal/googleauth/reauth.go` | `Reauth()` function — launches browser OAuth flow, verifies identity, persists new refresh token |
| `internal/googleauth/reauth_test.go` | Tests for `Reauth()`, email mismatch, scope preservation, `servicesFromScopes()` |
| `internal/googleapi/transport.go` | `Reauth` field on `RetryTransport`; `invalid_grant` detection + auto-reauth retry in `RoundTrip` |
| `internal/googleapi/client_auth.go` | `ResetRefreshToken` on `resettableOAuthTokenSource`/`persistingTokenSource`; `isInvalidGrantError()` detector; `refreshTokenResetter` interface |
| `internal/googleapi/client.go` | `reauthFunctionFromContext()` closure — loads stored token, calls `Reauth`, resets token source |
| `internal/googleapi/auth_dependencies.go` | `ReauthFunc` type, `Reauth` field, `WithNoInput()`/`NoInputFromContext()` |
| `internal/cmd/root.go` | Wires `reauthFn` into `AuthDependencies`; propagates `--no-input` + TTY detection |
| `internal/googleapi/auto_reauth_test.go` | Tests for `invalid_grant` detection, retry state machine, `NoInputFromContext` |
| `internal/googleapi/reauth_glue_test.go` | Integration tests for production closure + real token source chain |
| `docs/auto-reauth-issue-draft.md` | Issue draft (can be split out) |
| `CHANGELOG.md` | Unreleased entry |

## Testing

- `go build ./...` ✅
- `go vet ./...` ✅
- `go test ./internal/googleapi/... ./internal/googleauth/... -count=1` ✅
- `go test ./internal/cmd/... -count=1` ✅
- `go test -race ./internal/googleapi/... -run Reauth` ✅
- **Not yet tested against live Google APIs** (see disclaimer above)

## Related issue

Issue draft included in `docs/auto-reauth-issue-draft.md` — can be posted separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Auth: automatically re-authorize when the stored refresh token is expired or revoked (`invalid_grant`). In interactive sessions, `gog` opens a browser-based OAuth flow, persists the new refresh token to the keyring, resets the in-memory token source, and retries the original API request — mirroring `yup-oauth2`'s `InstalledFlowAuthenticator` fallback. The reauth preserves the original grant's full scope/service set (preventing silent grant narrowing) and verifies the authorized email matches. Suppressed under `--no-input` or non-TTY stdin so CI and piped runs surface a clear error with the manual `gog auth add` command instead. Excluded for ADC, service accounts, and direct access tokens.
+
 ## 0.35.0 - 2026-08-09
 
 - Install: move the Go module to `github.com/openclaw/gogcli`; new releases install with `go install github.com/openclaw/gogcli/cmd/gog@latest` instead of the former `github.com/steipete/gogcli` path.
--- a/docs/auto-reauth-issue-draft.md
+++ b/docs/auto-reauth-issue-draft.md
@@ -1,0 +1,92 @@
+# Feature: Auto-reauth on expired/revoked refresh tokens
+
+## Summary
+
+When a stored OAuth refresh token is expired or revoked (`invalid_grant`),
+`gog` fails with a hard error and requires the user to manually re-run
+`gog auth add`. This is in contrast to other CLI tools (e.g. Rust's
+`yup-oauth2` `InstalledFlowAuthenticator`) which automatically fall back to
+a browser-based re-authorization flow when the refresh token is invalid,
+making the failure transparent to the user.
+
+## Problem
+
+When the refresh token becomes invalid (Google revokes it for various
+reasons: 6-month inactivity, password change, token limit reached, app in
+Testing mode with 7-day expiry, user manually revoking access), any
+`gog` command that needs authentication fails:
+
+```
+$ gog calendar events --today
+Error: refresh access token: oauth2: "invalid_grant" "Token has been expired or revoked."
+```
+
+The user must then manually run:
+
+```
+gog auth add you@example.com --services gmail,calendar,drive
+```
+
+This is particularly frustrating because:
+
+1. **It's a hard stop** — no graceful degradation or recovery hint beyond the
+   raw OAuth error.
+2. **The error message doesn't tell the user what to do** — it's a wrapped
+   `golang.org/x/oauth2` error, not a user-facing diagnostic.
+3. **It happens repeatedly** — especially for users whose OAuth app is in
+   Testing mode (7-day refresh token expiry) or who don't use `gog` daily.
+4. **Other tools handle this transparently** — `yup-oauth2` (used by the
+   `today` Rust CLI) detects `invalid_grant` during token refresh and
+   automatically falls back to the Installed Flow (opens a browser, obtains a
+   new refresh token, persists it, and retries the original request).
+
+## Proposed behavior
+
+When `gog` detects `invalid_grant` during a token refresh attempt:
+
+1. **If running interactively** (stdin is a terminal, `--no-input` is not set):
+   - Print a message to stderr: "Refresh token expired or revoked. Re-authorizing…"
+   - Automatically launch the browser-based OAuth flow (same as `gog auth add`)
+     using the stored account's services and client.
+   - On success, persist the new refresh token to the keyring and retry the
+     original API request.
+   - On failure (user denies, browser doesn't open, etc.), surface a clear
+     error with the re-auth command to run manually.
+
+2. **If running non-interactively** (`--no-input`, CI, pipes):
+   - Do NOT auto-launch a browser.
+   - Surface a clear error message with the exact `gog auth add` command to run.
+
+## Design considerations
+
+- **Security**: Auto-reauth should only trigger when the refresh token is
+  specifically revoked/expired (`invalid_grant`), not for other OAuth errors.
+  The browser flow uses the same PKCE + state validation as `gog auth add`.
+- **Scope preservation**: The reauth should request the same services/scopes
+  as the stored token, not a broader or narrower set.
+- `--force-consent` should be used during auto-reauth to ensure Google
+  returns a new refresh token (without it, Google may omit the refresh token
+  for returning users).
+- **Keychain access**: The reauth flow needs keychain write access to persist
+  the new token. On macOS, this may trigger a Keychain permission prompt.
+- **Timeout**: The auto-reauth browser flow should have a reasonable timeout
+  (e.g. 2 minutes) to avoid hanging indefinitely in CI-like environments.
+
+## Prior art
+
+- `yup-o-auth2` (Rust): `InstalledFlowAuthenticator::find_token_info()` —
+  on refresh failure, falls back to `auth_flow.token()` which opens a browser.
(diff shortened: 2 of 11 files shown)


inamiy:
## Live behavior proof

Tested against real Google Calendar API with a revoked refresh token (revoked via `https://oauth2.googleapis.com/revoke`).

### Test 1: `--no-input` path (non-interactive)

Token revoked, run with `--no-input`:

```
$ gog --no-input calendar events --today
refresh token expired or revoked: base token source: resettable oauth token source: oauth2: "invalid_grant" "Token has been expired or revoked."; run 'gog auth add' to re-authorize
EXIT_CODE=1
```

✅ No browser launched; clear error with actionable `gog auth add` hint.

### Test 2: Interactive auto-reauth (TTY)

Same revoked token, run from a terminal (pseudo-TTY via `script`):

```
$ gog calendar events --today
Refresh token expired or revoked. Re-authorizing…
Opening browser for authorization…
If the browser doesn't open, visit this URL:
https://accounts.google.com/o/oauth2/auth?access_type=offline&client_id=…&prompt=consent&redirect_uri=http%3A%2F%2F127.0.0.1%3A56780%2Foauth2%2Fcallback&…
Authorization received. Finishing…
Re-authorization successful. Retrying request…

ID                              START                       END                         SUMMARY
ru09mkqs02fk3sr3jasauom9j0      2026-08-08T23:50:00+01:00   2026-08-09T00:10:00+01:00   バグレポート整理…
bmqd2e7keb4n13flt5hbrs6aqo      2026-08-09T00:00:00+01:00   2026-08-09T10:30:00+01:00   💤 睡眠
…

EXIT_CODE=0
```

✅ Full auto-reauth flow completed: browser opened → OAuth consent → new token persisted → request retried → calendar events returned.

### Test 3: Post-reauth token verification

```
$ gog auth list --check
inamiy@gmail.com  default  calendar,drive,gmail  2026-08-09T15:42:57Z  true  oauth
```

✅ Token valid after auto-reauth.

### What was verified against live Google APIs

1. ✅ `invalid_grant` detection works with real Google OAuth responses
2. ✅ Browser OAuth flow opens with `--force-consent` and `include_granted_scopes=true`
3. ✅ Scope preservation — OAuth URL contained all original scopes (calendar, drive, gmail)
4. ✅ New refresh token persisted to macOS Keychain
5. ✅ In-memory token source reset — retried request succeeded (did not reuse revoked token)
6. ✅ `--no-input` suppression — no browser launched, clear error with `gog auth add` hint
7. ✅ Non-TTY suppression — no browser when stdin is piped
8. ✅ Token revocation via Google's `/revoke` endpoint correctly triggers the flow

---

## UX caveats (known rough edges for PoC)

The current auto-reauth UX has some rough edges that should be polished before merging:

1. **Browser opens suddenly without confirmation.** When `invalid_grant` is detected in an interactive session, the browser immediately opens the Google OAuth consent screen. There is no prompt like *"Your session has expired. Re-authorize now? [Y/n]"* — the user might be surprised by a browser window appearing unexpectedly mid-command. A confirmation prompt (respecting `--no-input`) would make this feel more deliberate.

2. **Keychain password prompt may appear without explanation.** The `Reauth` function calls `EnsureKeychainAccess` before launching the browser, and `store.SetToken` writes the new token after the flow completes. On macOS, either of these can trigger a Keychain access prompt (especially for binaries not signed with keychain-access-groups entitlement). The user sees a system dialog asking for keychain approval with no context about *why* gog needs keychain access at this moment. This is partially related to existing issue #939 (macOS Tahoe Keychain C API issues) and may not appear in all configurations, but the UX could be smoother with a stderr message like *"Storing new credentials in keychain…"* before the write.

3. **The keychain access step (`EnsureKeychainAccess`) may be unnecessary in some cases.** If the keychain is already unlocked (common for interactive sessions where the user has recently used gog), the pre-flight `EnsureKeychainAccess` call is redundant and may trigger an extra permission prompt that would not appear during a normal `gog auth add`. This is worth investigating — it may be better to skip the pre-flight check and let `SetToken` handle keychain access naturally, with a clear error if it fails.

These are UX polish items, not correctness issues — the core auto-reauth mechanism (detect → reauth → persist → reset → retry) works correctly as demonstrated above.

steipete:
LAND verdict: I prepared the maintainer landing branch in #974 and preserved your authored feature commit.

Thank you for the unusually useful live Google proof—it made the recovery behavior much easier to validate. Before landing, the maintainer branch fixes the stale persistence cache, adds a post-reauth refresh regression, requires confirmation before browser launch, and coalesces recovery across separate service clients so one revoked account cannot trigger multiple concurrent browser flows. It also narrows detection to typed OAuth `invalid_grant` errors and keeps the non-interactive/manual recovery path.

Please leave this PR open while #974 receives final maintainer review. No action is requested from you.

steipete:
Landed — thank you, and your commit is preserved in the merge.

This shipped as #974 (`229da128`), which carries your authored commit plus a
maintainer hardening pass. Closing this one in favour of that branch rather
than merging here, since the follow-up work happened there.

Your diagnosis was right and the live Google Calendar proof you attached was
genuinely useful. Four things needed tightening before it was safe to land on
the auth path, and they are worth writing down:

1. **The revoked token could come back.** `ResetRefreshToken` updated only the
   inner source, so `persistingTokenSource.tok` still held the revoked token. A
   later refresh response that omits `refresh_token` — which is normal — would
   persist the revoked value again. Replacement metadata is now written before
   the in-memory source is swapped.
2. **Browser stampede.** Retry state was bounded per HTTP request, but one
   command can build several service clients for the same account, so a single
   revocation could open several browser flows at once. Recovery is now
   serialized across every client in the command, and waiters adopt the newly
   stored token instead of starting their own flow.
3. **Consent.** The interactive path opened a browser immediately. It now asks
   first, and declining fails cleanly:

   ```
   Refresh token for <account> expired or was revoked. Re-authorize now? [y/N]:
   … re-authentication failed: reauth: cancelled; run 'gog auth add' to re-authorize manually
   ```
4. **Detection and identity.** `invalid_grant` was matched on arbitrary error
   text; it now matches only the typed OAuth retrieval error. An empty identity
   email also fell back to the expected email, which quietly weakened the
   identity check — mismatched or missing emails are now rejected.

The `--no-input`, non-TTY, ADC, service-account, direct-access-token,
non-replayable-request and one-retry boundaries are all preserved, with
regression coverage for post-reauth refresh and concurrent clients.

You are credited in the 0.35.1 changelog. Good catch on a real gap.

PR #976, opened 2026-08-10
feat(safety-profile): lock flag values with locked-flags

ronny-rentner:
Safety profiles can allow or deny commands, but nothing in them can fix a *flag value*. Sanitized output happens only if the caller passes `--sanitize-content`, and a flag that does take a value from the environment is still overridden by the command line. When that command line is written by a model rather than a person, neither is a setting an operator can rely on.

This adds a `locked-flags` mapping to the profile format. A locked flag is applied before the command runs, and setting it on the command line is an error rather than an override.

```yaml
locked-flags:
  sanitize-content: true
  wrap-untrusted: true
  no-input: true
```

**How it works.** `parseRaw` reads `locked-flags` alongside `allow`/`deny` and keeps it out of command flattening; values may be bool, int or string. The generator emits `bakedSafetyLockedFlag` as a hashed switch with the same collision check as the rule matchers, so a locked flag name is no more patchable than a command rule. `enforceLockedFlags` runs beside `enforceBakedSafetyProfile`, applies each value through the flag's own parser, and rejects a command line that sets one. Locks apply per selected command, so a locked flag a command doesn't declare is silently inert.

Because a locked value can make a command reject a combination the caller never asked for, usage errors carry a note naming the locked flags and the profile. That's attached once in the error path, so no command carries profile knowledge.

**Not a behaviour change.** Profiles without `locked-flags` are unaffected, stock builds compile a stub returning "not locked", and `readonly.yaml`/`agent-safe.yaml` are untouched. `locked-flags` could not previously appear in a profile — an unknown top-level key was a parse error — so no existing profile changes meaning.

**On the two new profiles.** `readonly-locked.yaml` and `agent-safe-locked.yaml` are copies of their originals plus the locks above (and `readonly: true` on the read side, which also rejects a mutating request inside an allowed command). I added them as copies because I didn't know whether you want preset locked profiles shipped at all — they're easy to drop if the mechanism alone is what you want, or they could replace the existing presets instead of sitting beside them, if you'd rather the shipped agent profile be locked by default. That last option is a behaviour change for anyone building `agent-safe`, which is why I didn't do it. Note that locking `sanitize-content` makes `gmail get --format raw` unavailable, since that combination is rejected; both descriptions say so.

**Proof** — no mailbox or credentials involved. The lock check runs before account resolution, and the last command needs only an account string, failing at flag validation before any network call. `m1` is a placeholder that is never used.

```
$ make build-safe PROFILE=safety-profiles/agent-safe-locked.yaml OUTPUT=bin/gog-agent-safe-locked
v0.35.0-11-gfee967bd-safe (fee967bd12b5 2026-08-10T16:51:09Z)
built bin/gog-agent-safe-locked with baked safety profile safety-profiles/agent-safe-locked.yaml

$ ./bin/gog-agent-safe-locked gmail get m1 --json --sanitize-content=false
flag --sanitize-content is locked by baked safety profile "agent-safe-locked"

$ ./bin/gog-agent-safe-locked gmail get m1 --json --wrap-untrusted=false
flag --wrap-untrusted is locked by baked safety profile "agent-safe-locked"

$ ./bin/gog-agent-safe-locked gmail get m1 --json --no-input=false
flag --no-input is locked by baked safety profile "agent-safe-locked"

$ ./bin/gog-readonly-locked gmail get m1 --json --readonly=false
flag --readonly is locked by baked safety profile "readonly-locked"

$ ./bin/gog-agent-safe-locked gmail get m1 --format raw --json --account a@b.com
--sanitize-content cannot be used with --format raw
note: --no-input, --sanitize-content, --wrap-untrusted locked by baked safety profile "agent-safe-locked"
```

The last one is the lock taking effect: nothing on that command line mentions `--sanitize-content`, and the refusal only fires when it is true.

**Tests.** Sixteen. `internal/safetyprofile` covers value parsing, non-scalar rejection, and locked flags not leaking into command rules. `cmd/bake-safety-profile` covers the emitted hashed lookup, that the flag name never appears verbatim, and that profiles without locks still compile. `internal/cmd` covers every spelling of setting a locked flag (including setting it to the value it already has), inertness on commands without the flag, stock builds ignoring locks, unparsable values failing loudly, an int lock reaching the command, and the note text. `go test ./...` passes except `TestDocsWriteUpdate_FileInputErrors`, which fails in my sandbox because `/nonexistent` returns `permission denied` instead of `no such file`; unrelated, reproducible with `cat`.

**Docs.** `docs/safety-profiles.md` gains a "Locked Flags" section covering syntax, semantics, and which flags not to lock, with `--reply-all` as the worked counter-example.


ronny-rentner:
Fixed in `fd24ccc7` and `22e8ed95`.

Precedence now runs after `enforceLockedFlags`, and a locked mode outranks the competing one. Reordering alone wasn't enough: a locked flag isn't "provided", so `--plain` would still have cleared a locked `json`.

One behaviour decision worth surfacing — the competing flag is treated by origin:

- explicit `--plain` against a locked `json` is refused, since the caller asked for output the profile forbids
- `GOG_PLAIN` is overridden silently, since an environment default is a preference, not a request about this invocation

Two explicit modes remain a conflict error, as before.

**Proof**, with a throwaway profile locking only `json` (neither shipped profile locks an output mode):

```yaml
name: json-locked
locked-flags:
  json: true
version: true
```

```
$ make build-safe PROFILE=json-locked.yaml OUTPUT=bin/gog-json-locked
v0.35.0-13-g22e8ed95-safe (22e8ed956367 2026-08-10T18:05:08Z)
built bin/gog-json-locked with baked safety profile json-locked.yaml

$ ./bin/gog-json-locked --plain version
flag --plain conflicts with --json, locked by baked safety profile "json-locked"

$ GOG_PLAIN=1 ./bin/gog-json-locked version
{
  "commit": "22e8ed956367",
```

Both previously failed with `invalid output mode (cannot combine --json and --plain)`.

Four tests cover the CLI and environment combinations; I confirmed they're not vacuous by restoring the old ordering and watching them fail. Docs updated to state the environment-versus-explicit split.

Not guarded, deliberately: a profile locking both `json` and `plain` builds and then fails on every invocation. Catching that at bake time would mean the generator hard-coding which flags are mutually exclusive — happy to add it if you'd prefer.

The two P1s are unchanged and on your side; the PR description lists the preset options.

ronny-rentner:
All four addressed in `b3a42d56`.

**Presence** — `flagProvided` now counts a locked flag as given, so presence-sensitive commands (`calendar_edit.go` and similar) include the locked value in their patch requests. A separate `flagOnCommandLine` keeps the narrower question for lock enforcement and output-mode precedence; without the split, enforcement would see its own lock as an override and reject every invocation.

**Unknown names** — the generator emits a count alongside the hashed lookup, and a pre-enforcement pass walks the CLI model counting matches. A name matching nothing refuses to run:

```
$ ./bin/gog-typo version
baked safety profile "typo" locks 1 flag(s) but only 0 exist; check the locked-flags names

$ strings bin/gog-typo | grep -c readonlly
0
```

The count rather than a name list is deliberate: emitting the names would have put them in the binary as patchable text and undone the hashing the rule matchers rely on.

One difference from the finding as written: this refuses at first invocation rather than during profile validation, so a typo still builds. Validating at bake time would mean the generator constructing the Kong model to know which flags exist — happy to do it if you want the check there.

**`--home` and required flags** — refused, taking the "reject as unsupported" option, since both are consumed before locks run: `--home` by the pre-parse layout scan that picks config and credential roots, required flags by Kong's own validation.

Six tests added, docs updated with the behaviour and the three refusals.

steipete:
Maintainer verdict: **LAND**, using the hardened maintainer stack at `openclaw/gogcli:triage/t8-pr976` (`6d30196f`). No contributor revision is requested.

The core idea is sound, but a generic bool/int/string lock surface was too broad for a security boundary. The landing stack narrows `locked-flags` to boolean policy switches, which covers the motivating controls (`sanitize-content`, `wrap-untrusted`, `no-input`, and `readonly`) without compiling arbitrary paths, account names, tokens, or other values into policy binaries. It also removes the two copied preset files to avoid permanent policy drift; operators can copy an existing preset and add the few locks they need.

The stack additionally fixes false-valued output locks, makes `GOG_AUTO_JSON` subordinate to a lock, validates unknown/required/non-boolean/pre-parse flags before execution, rejects `help`, `version`, and `home` locks, preserves aliases, and keeps stock builds unchanged. Contributor authorship is preserved in the stack and maintainer commits carry Ronny's co-author credit.

Proof run on the final stack:

```text
$ make test
ok github.com/openclaw/gogcli/internal/cmd 76.889s
... all Go and Node tests passed

$ GOLANGCI_LINT_CACHE=/private/tmp/gogcli-t8-golangci-cache make lint
0 issues.

$ make docs-check
docs coverage ok: 709 command pages, 27 feature pages

$ ./gog-lock-json-true version
{ "commit": "a16e9334583b", ... }

$ ./gog-lock-json-true --json=false version
flag --json is locked by baked safety profile "live-json-true"
exit 2

$ ./gog-lock-json-true --machine=false version
flag --json is locked by baked safety profile "live-json-true"
exit 2

$ GOG_JSON=1 ./gog-lock-json-false version
v0.35.0-16-ga16e9334-dirty-safe (...)

$ GOG_AUTO_JSON=1 ./gog-lock-json-false version
v0.35.0-16-ga16e9334-dirty-safe (...)

$ make build-safe PROFILE=live-lock-string.yaml OUTPUT=bin/gog-lock-string
parse profile: locked-flags: json: expected boolean value, got string
exit 2

$ make build-safe PROFILE=live-lock-nonbool.yaml OUTPUT=bin/gog-lock-nonbool
baked safety profile "live-nonbool" locks --format, but only boolean flags can be locked
exit 2
```

Final structured review found no accepted/actionable P0/P1 findings (confidence 0.94). The original PR CI failure was the reported staticcheck selector issue; the landing stack fixes it and the local lint gate is clean.

PR #984, opened 2026-08-13
feat(gmail): add --count to search and messages search

### LINKED ISSUE #983: feat(gmail): report how many results a search really has — resultSizeEstimate saturates at 201 and is unusable as a count
## Problem

`gog gmail search` and `gog gmail messages search` give a caller no way to learn how large a result set is without walking it. The JSON payload is items + `nextPageToken`, so "are there more?" is answerable but "how many more?" is not.

That gap causes a specific, repeatable failure in agent/LLM callers: a capped page gets read as the complete answer, and the caller reports that a message does not exist when it does. A page token alone is easy to skim past; a magnitude is not.

## Why the obvious fix doesn't work

The Gmail API returns `resultSizeEstimate` on both `users.threads.list` and `users.messages.list`, and `gog` currently drops it — `gmail_search.go:~97` and `gmail_messages.go:~110` build the JSON payload by hand from items + `nextPageToken`.

**Surfacing that field would be a bug, not a fix.** It saturates. Measured on a live account, gog `v0.35.0 (402def5)`, 2026-08-12:

```
$ gog api call gmail v1 users.messages.list \
    --params '{"userId":"me","q":"<Q>","maxResults":1,"fields":"resultSizeEstimate"}'
```

| query | `resultSizeEstimate` | true count (`--all`) |
|---|---|---|
| `from:freshbooks.com` | 201 | **21** |
| `from:housecallpro.com` | 201 | **6** |
| `from:thumbtack.com newer_than:30d` | 201 | **3** |
| `from:honeybook.com` | 201 | **9** |
| `invoice` | 201 | >500 |
| `in:inbox` | 201 | large |
| `zzzznomatch` | 0 | 0 |

It returns **201 for every non-empty query** and 0 for an empty one — a has-results boolean wearing a number's clothes. It is also independent of `maxResults` (identical at 1, 10, and 100), so it isn't a per-page figure either. Emitting it would let a caller report "3 of ~201" when the truth is 3 of 6.

## Proposed fix

An opt-in `--count` flag on both search commands that reports a **real** number, obtained with one extra list call requesting a single maximal page of bare ids:

```
maxResults=500, fields="threads/id,nextPageToken"   (or messages/id)
```

- fits in one page → exact total, emitted as `totalMatches`
- fills the page with more behind it → honest lower bound, emitted as `totalMatchesAtLeast`

Verified against the same account:

| query | probe result | matches `--all`? |
|---|---|---|
| `from:freshbooks.com` | counted 21, no more pages | ✅ 21 |
| `from:housecallpro.com` | counted 6, no more pages | ✅ 6 |
| `invoice` | counted 500, more pages | ✅ lower bound |

The response is ids only, so it stays small, and the flag keeps it off the default path — no extra round-trip for callers who don't ask.

Exact whenever the set fits a page, which is the common case and always the case for the narrow queries a false negative is most likely to come from.

## Notes

- I have this implemented and validated in a downstream wrapper already; happy to send the `--count` PR here (branch in progress) if the shape looks right to you.
- Open question for a maintainer: whether `--count` should also print on the text/table path, or stay JSON-only.

chrischall:
Closes #983.

`gmail search` and `gmail messages search` emit items plus `nextPageToken`, so a caller can tell that *more* results exist but not *how many*. That gap is where agent/LLM callers go wrong: a capped page reads as the complete answer, and the caller reports that a message doesn't exist when it does.

## Why not `resultSizeEstimate`

It's the obvious source, and it saturates. Measured on a live account, `v0.35.0 (402def5)`, 2026-08-12:

| query | `resultSizeEstimate` | true count |
|---|---|---|
| `from:freshbooks.com` | 201 | **21** |
| `from:housecallpro.com` | 201 | **6** |
| `from:thumbtack.com newer_than:30d` | 201 | **3** |
| `from:honeybook.com` | 201 | **9** |
| `zzzznomatch` | 0 | 0 |

201 for every non-empty query, and identical at `maxResults` 1, 10 and 100 — so it isn't a per-page figure either. It's a has-results boolean wearing a number's clothes. Emitting it would let a caller report "3 of ~201" when the truth is 3 of 6, which is worse than reporting nothing.

## What `--count` does

One extra list call for a single maximal page of bare ids:

```
maxResults=500, fields="threads/id,nextPageToken"   (messages/id for messages search)
```

- set fits the page → exact total → `totalMatches`
- page fills with more behind it → honest lower bound → `totalMatchesAtLeast`

Two separate field names so a saturated probe can never be mistaken for a total. Exact for the narrow queries where a wrong count does the most damage.

Opt-in, so nobody pays the round-trip without asking. Text path prints to stderr, keeping stdout parseable.

## Live verification (`bin/gog` from this branch)

```
$ ./bin/gog gmail search from:freshbooks.com --max=3 --count --json
   returned: 3 | totalMatches: 21           ground truth (--all): 21  ✅

$ ./bin/gog gmail search from:housecallpro.com --max=3 --count --json
   returned: 3 | totalMatches: 6            ground truth (--all): 6   ✅

$ ./bin/gog gmail messages search from:honeybook.com --max=2 --count --json
   returned: 2 | totalMatches: 9            ground truth (--all): 9   ✅

$ ./bin/gog gmail search invoice --max=3 --count --json
   returned: 3 | totalMatchesAtLeast: 500   (large set — lower bound)  ✅
```

Opt-in respected — without `--count`, no count fields and no probe:
```
$ ./bin/gog gmail search from:housecallpro.com --max=2 --json
   count fields: NONE
```

Text path — hint on stderr, table clean on stdout:
```
$ ./bin/gog gmail search from:housecallpro.com --max=2 --count --plain
stderr: Showing 2 of 6 matches.
stdout: 3 rows, contains "matches." 0 times
```

## Tests

8 new tests in `internal/cmd/gmail_search_count_test.go` via the existing `httptest` idiom — exact count, lower bound, probe shape (asserts ids-only `fields`, `maxResults=500`, and that it reuses the search's own query), opt-in (probe must not run without the flag), zero-match, messages-search variant, and the stderr wording.

`make lint` clean (0 issues); `go test ./internal/cmd/` passes (81s).

## Notes

- `CHANGELOG.md` deliberately untouched — maintainer-owned per AGENTS.md landing flow.
- Open question from #983 still worth your call: whether the text path should print the count at all, or stay JSON-only. Currently it prints to stderr.

### DIFF
--- a/internal/cmd/gmail_messages.go
+++ b/internal/cmd/gmail_messages.go
@@ -33,6 +33,7 @@ type GmailMessagesSearchCmd struct {
 	Page                    string   `name:"page" aliases:"cursor" help:"Page token"`
 	All                     bool     `name:"all" aliases:"all-pages,allpages" help:"Fetch all pages"`
 	FailEmpty               bool     `name:"fail-empty" aliases:"non-empty,require-results" help:"Exit with code 3 if no results"`
+	Count                   bool     `name:"count" help:"Also report how many messages match in total (exact when the set fits one page, otherwise a lower bound)"`
 	Timezone                string   `name:"timezone" short:"z" help:"Output timezone (IANA name, e.g. America/New_York, UTC). Default: GOG_TIMEZONE, config, then local"`
 	Local                   bool     `name:"local" help:"Use local timezone (default behavior, useful to override --timezone)"`
 	IncludeBody             bool     `name:"include-body" help:"Include decoded message body (JSON is full; text output truncates only unusually large bodies)"`
@@ -82,10 +83,14 @@ func (c *GmailMessagesSearchCmd) Run(ctx context.Context, flags *RootFlags) erro
 
 	if len(messages) == 0 {
 		if outfmt.IsJSON(ctx) {
-			return writePagedJSONResult(ctx, map[string]any{
+			payload := map[string]any{
 				"messages":      []messageItem{},
 				"nextPageToken": nextPageToken,
-			}, 0, c.FailEmpty)
+			}
+			if c.Count {
+				payload["totalMatches"] = int64(0)
+			}
+			return writePagedJSONResult(ctx, payload, 0, c.FailEmpty)
 		}
 		u.Err().Println("No results")
 		return failEmptyExit(c.FailEmpty)
@@ -106,11 +111,23 @@ func (c *GmailMessagesSearchCmd) Run(ctx context.Context, flags *RootFlags) erro
 		return err
 	}
 
+	var matchCount gmailMatchCount
+	if c.Count {
+		matchCount, err = countGmailMessageMatches(ctx, svc, query)
+		if err != nil {
+			return err
+		}
+	}
+
 	if outfmt.IsJSON(ctx) {
-		return writePagedJSONResult(ctx, map[string]any{
+		payload := map[string]any{
 			"messages":      items,
 			"nextPageToken": nextPageToken,
-		}, len(items), c.FailEmpty)
+		}
+		if c.Count {
+			matchCount.apply(payload)
+		}
+		return writePagedJSONResult(ctx, payload, len(items), c.FailEmpty)
 	}
 
 	if len(items) == 0 {
@@ -125,6 +142,10 @@ func (c *GmailMessagesSearchCmd) Run(ctx context.Context, flags *RootFlags) erro
 		gmailMessageColumns(c.IncludeBody, c.IncludeAttachments, c.Full),
 	); err != nil {
 		return err
+	}
+	// stderr, so the table on stdout stays parseable.
+	if c.Count {
+		printGmailMatchCount(u, len(items), matchCount)
 	}
 	printNextPageHintWithAll(u, nextPageToken, "--all/--all-pages")
 	return nil
--- a/internal/cmd/gmail_messages.go
+++ b/internal/cmd/gmail_messages.go
@@ -33,7 +33,7 @@ type GmailMessagesSearchCmd struct {
 	Page                    string   `name:"page" aliases:"cursor" help:"Page token"`
 	All                     bool     `name:"all" aliases:"all-pages,allpages" help:"Fetch all pages"`
 	FailEmpty               bool     `name:"fail-empty" aliases:"non-empty,require-results" help:"Exit with code 3 if no results"`
-	Count                   bool     `name:"count" help:"Also report how many messages match in total (exact when the set fits one page, otherwise a lower bound)"`
+	Count                   bool     `name:"count" help:"Also report how many messages match the query in total, as totalMatches (exact) or totalMatchesAtLeast (lower bound). Always counts the WHOLE query, not the remainder after --page. Free with --all; no effect with --results-only"`
 	Timezone                string   `name:"timezone" short:"z" help:"Output timezone (IANA name, e.g. America/New_York, UTC). Default: GOG_TIMEZONE, config, then local"`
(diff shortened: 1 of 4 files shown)


chrischall:
Pushed `ca11466` addressing the two open questions from the review on #983.

Both turned out to be real defects rather than just undocumented behaviour, verified against a live account before fixing:

**`--results-only` was spending a request for nothing.** `unwrapPrimary` emits the bare result array, and `threads`/`messages` are both in its known-result list — so the count field was computed and then silently dropped. Confirmed live: `--count --results-only` returned the plain array with no count anywhere, having made the extra Gmail call regardless. The probe is now skipped and the caller is told on stderr instead of being left to guess.

**`--all` was asking Google a question it had just answered.** The walk has already exhausted every page, so the items in hand *are* the whole set. Confirmed live: `--all --count` on a 6-thread query reported `totalMatches: 6` — correct, but via a redundant round-trip. The total now comes from the walk, which makes it exact by construction rather than by probe.

**`--page` needed no code change but did need stating.** Confirmed live: on page 2 of a 6-thread query, `totalMatches` was `6`, not the 4 remaining. That is deliberate — "how many match this query" is the number that stops a caller reporting an absence, and keeping it page-independent means it does not drift while paging. Now said outright in the flag help.

On the product question in the review — **JSON-only vs. also text**: happy to defer entirely. Worth one data point for the decision: the text path currently prints `Showing 2 of 6 matches.` to **stderr**, which is the same channel and shape as the existing `printNextPageHintWithAll` hint right beside it, so stdout stays byte-identical and parseable either way. If you'd rather it were strictly JSON-only, deleting the `printGmailMatchCount` call is a two-line change and I'll push it on request.

Two new tests cover the skip paths, including one that would fail if a probe result could masquerade as an `--all` total. `make lint` clean (0 issues), `go test ./internal/cmd/` passing.

Noting the `clawsweeper:no-new-fix-pr` / `needs-product-decision` labels: this PR is not trying to jump the queue on that decision — it's here as a working reference implementation with live proof attached, and I'm happy for it to sit until a maintainer picks the contract (or to close it if the answer is no).

chrischall:
Live proof for `ca11466`, on a real account with `bin/gog` built from this branch (`v0.35.0`-based, 2026-08-13):

**`--results-only` — probe skipped, caller told, stdout unchanged**

```
$ ./bin/gog gmail search from:housecallpro.com --max=2 --count --json --results-only
stderr: --count has no effect with --results-only: the count is an envelope field,
        and --results-only emits only the result array. Skipping the extra request.
stdout: starts with '[' (bare array), occurrences of "totalMatches": 0
```

Before this commit the same command made the extra Gmail request and then discarded its result silently.

**`--all` — total from the walk**

```
$ ./bin/gog gmail search from:housecallpro.com --all --count --json
returned: 6 | totalMatches: 6
```

Same answer as before the change, now without the redundant round-trip.

**On what each layer proves.** The live runs above confirm the observable contract. They cannot themselves prove the *request did not fire* — gog has no HTTP-debug flag to count calls — so that is asserted deterministically in the unit tests instead, where the httptest handler increments a counter on any ids-only probe and both new tests require it to stay at zero. `TestGmailSearch_Count_SkipsProbeWithAll` additionally serves a probe response of 99 items, so if the probe ever fires again the reported total becomes 99 instead of 2 and the test fails loudly rather than silently passing.

Earlier live proof for the core feature is unchanged and still in the PR description: exact counts of 21, 6 and 9 all matching `--all` ground truth, and a lower bound of 500 on a large set.

PR #1003, opened 2026-08-18
feat(gmail): support custom API endpoint

sallyom:
Adds `GOG_GMAIL_BASE_URL` to route Gmail API requests through a compatible custom endpoint, such as a security proxy or test server.

The override applies to standard Gmail and batch-delete clients. Existing behavior is unchanged when the variable is unset.

## Testing

- Added custom-endpoint coverage for Gmail service creation.
- Added command-level coverage for authenticated Gmail get and batch-delete requests.
- `go test ./internal/googleapi`
- `make ci`
- Built `bin/gog` and ran both command paths against a live loopback-compatible endpoint using a synthetic token and synthetic message identifiers. The endpoint logged only the request method/path and whether bearer authentication was present:

```text
PROOF method=GET path=/gmail/v1/users/me/messages/proof-message authorization_present=true
PROOF method=POST path=/gmail/v1/users/me/messages/batchDelete authorization_present=true
```

No token value, mailbox content, or external Gmail account data was included in the proof.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -176,6 +176,7 @@ Environment:
 - `GOG_ENABLE_COMMANDS_EXACT=calendar.events,gmail.search` (optional exact allowlist; dot paths allowed; parent paths do not allow children)
 - `GOG_DISABLE_COMMANDS=gmail.send,gmail.drafts.send` (optional denylist; dot paths allowed)
 - `GOG_GMAIL_NO_SEND=1` (block Gmail send operations)
+- `GOG_GMAIL_BASE_URL=https://gmail-proxy.example/` (override the Gmail API endpoint; the endpoint receives OAuth bearer credentials and Gmail data, so use only a trusted HTTPS endpoint; plain HTTP is for loopback-only local development/testing)
 - `config.json` can also set `keyring_backend` (JSON5; env vars take precedence)
 - `config.json` can also set `default_timezone` (IANA name or `UTC`)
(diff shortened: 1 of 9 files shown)


sallyom:
@clawsweeper re-review

sallyom:
@clawsweeper re-review

sallyom:
wrt [P1]:
`GOG_GMAIL_BASE_UR` is an explicit operator-controlled trust boundary. It intentionally sends OAuth bearer credentials and Gmail request data to that endpoint when configured. The default remains the Google Gmail API; remote overrides require HTTPS, and HTTP is restricted to loopback. This behavior and its security implications are documented.

steipete:
Closing on security grounds rather than on effort — thank you for the work, but this can't land in its current shape.

The webhook/forwarding path sends Gmail OAuth **bearer credentials** and mailbox contents to operator-supplied endpoints, and the OAuth transport reattaches those credentials across redirects. That means a misconfigured or attacker-influenced target URL — or a redirect chain to one — receives live Google credentials and message contents, not just the intended payload. That's a credential-exfiltration shape, so it fails closed for us regardless of how the feature is configured.

If you want to pursue this, the design that would be reviewable: never attach the OAuth transport to outbound third-party requests (use a plain client with no credential provider), refuse to follow cross-origin redirects on that path entirely, and send only an explicitly constructed payload with no raw credential material. That plus live proof against a redirecting endpoint would make it a real conversation.

PR #1005, opened 2026-08-20
feat(appscript): add push, pull, deploy, and version management

haosdent:
## Scope

`gog appscript` could read a project (`get`, `content`) and run a deployed
function, but never change one. This adds the write half so an Apps Script
project can be developed and released from the CLI:

| Command | Purpose |
|---|---|
| `appscript push <scriptId> <dir>` | Replace project content from a local directory |
| `appscript pull <scriptId> <dir>` | Write project content into a local directory |
| `appscript deploy <scriptId>` | Cut a version and deploy it (prints the web app URL) |
| `appscript deployments <scriptId>` | List deployments |
| `appscript undeploy <scriptId> <deploymentId>` | Delete a deployment |
| `appscript versions <scriptId>` | List versions |

## Notes on the design

- **`deploy --deployment-id` updates in place**, so a redeploy keeps the same
  web app URL. Without it, every deploy mints a new URL, which is the main
  thing that makes scripted redeploys painful today.
- **Extension mapping.** Apps Script stores a file's extension in its `type`,
  not its `name`, so push/pull translate between `Code.gs` and
  `{name: "Code", type: SERVER_JS}`. Only `appsscript.json` is accepted as
  JSON, so a stray `package-lock.json` in the directory is skipped without
  being read.
- **The manifest is mandatory.** `UpdateContent` rejects a payload without
  `appsscript.json`; push checks up front and says so, instead of passing
  through an opaque API error.
- **Pagination.** Both list commands use `loadPagedItems` /
  `writePagedJSONResult` / `printNextPageHintWithAll`, so `--max`, `--page`,
  `--all`, and `--fail-empty` behave as they do elsewhere. This matters here:
  the API defaults to 50 per page and `deploy` cuts a version every time, so
  dropping the next page token would silently under-report versions after a
  few months of use.
- **`undeploy` goes through `dryRunAndConfirmDestructive`** — deleting a
  deployment permanently breaks its web app URL, so it honours `-y/--force`
  and fails closed under `--no-input`.
- **Safety profiles.** `readonly.yaml` and `agent-safe.yaml` are allow-lists
  that fail closed, so the six new leaves are listed explicitly. Without this,
  even the read-only `pull`, `deployments`, and `versions` would be rejected
  (and hidden from help) in a baked build.
- `--readonly` needs no per-command code: `script.googleapis.com` is not on
  the POST allowlist in `internal/googleapi/read_only.go`, so push, deploy,
  and undeploy are already refused at the transport. `pull` still writes
  local files under `--readonly`, matching `drive download`.

## New flags

- `appscript deploy`: `--description`, `--deployment-id`
- `appscript deployments`, `appscript versions`: the standard
  `--max` / `--page` / `--all` / `--fail-empty` set

## Testing

- `make ci` passes (`fmt-check`, `lint`, `deadcode`, `test`, `docs-check`,
  `agent-skills-check`).
- New tests in `appscript_sync_test.go` and `appscript_deploy_test.go` cover
  the extension mapping, manifest enforcement, non-manifest JSON skipping,
  push/pull round trips, pulled-file permissions and path-traversal
  containment, `--all` across two pages, `nextPageToken` retention in
  `--json`, and the `undeploy` confirmation gate.
- `dryrun_e2e_test.go` gains rows for `appscript push/pull/deploy/undeploy`;
  `TestReadonlySafetyProfileBlocksNestedMutations` gains the two new
  mutations.
- Generated docs and agent skills were regenerated (`make docs-commands`,
  `make agent-skills`).

## Two open questions for maintainers

1. **Should `push` also require confirmation?** `UpdateContent` replaces the
   project wholesale, so any remote file absent from the local directory is
   deleted — `drive sync push` deliberately advertises "no remote deletes",
   so this repo is more conservative here. I left push without a prompt so it
   stays usable non-interactively, and instead spelled the behaviour out in
   the help text. Happy to route it through `dryRunAndConfirmDestructive` if
   you'd prefer.
2. **Command shape.** `deploy` / `deployments` / `undeploy` are flat sibling
   verbs over one sub-resource, whereas the rest of the CLI nests
   (`drive comments {list,create,delete}`). A nested
   `appscript deployments {list,create,delete}` would also make the safety
   profile and `--enable-commands` entries one block instead of three
   scattered keys. I kept the flat shape to preserve `appscript deploy` as
   the ergonomic one-liner, but I'll restructure if you want consistency.

### DIFF
--- a/.agents/skills/gog-appscript/SKILL.md
+++ b/.agents/skills/gog-appscript/SKILL.md
@@ -30,8 +30,14 @@ gog --readonly --account user@example.com appscript --help
 | --- | --- |
 | `content` | Get Apps Script project content |
 | `create` | Create an Apps Script project |
+| `deploy` | Create a version and deploy it (web app URL is printed) |
+| `deployments` | List deployments |
 | `get` | Get Apps Script project metadata |
+| `pull` | Pull an Apps Script project into a local directory |
+| `push` | Push a local directory into an Apps Script project (replaces every remote file) |
 | `run` | Run a deployed Apps Script function |
+| `undeploy` | Delete a deployment |
+| `versions` | List versions |
 
 Run `gog appscript <command> --help` for flags and `gog schema appscript <command> --json`
 for the machine-readable contract. Do not guess command syntax.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Apps Script: add `push`, `pull`, `deploy`, `deployments`, `undeploy`, and `versions` so a project's source can be synced from a local directory and released to a web app without leaving the CLI.
 - Security: bound `gmail watch serve` and the local OAuth callback HTTP servers with read, idle, and header-size limits so a slow or oversized request cannot stall the listener.
 
 ## v0.37.0 - 2026-08-14
--- a/docs/commands.generated.md
+++ b/docs/commands.generated.md
@@ -32,8 +32,14 @@ Generated from `gog schema --json`.
   - [`gog appscript (script,apps-script) <command> [flags]`](commands/gog-appscript.md) - Google Apps Script
     - [`gog appscript (script,apps-script) content (cat) <scriptId>`](commands/gog-appscript-content.md) - Get Apps Script project content
     - [`gog appscript (script,apps-script) create (new) --title=STRING [flags]`](commands/gog-appscript-create.md) - Create an Apps Script project
+    - [`gog appscript (script,apps-script) deploy <scriptId> [flags]`](commands/gog-appscript-deploy.md) - Create a version and deploy it (web app URL is printed)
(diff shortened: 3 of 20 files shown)


steipete:
Closing this one — the Apps Script expansion is broad and currently conflicted against main, but the blocking issue is behavioral: the push path **silently deletes remote files** that aren't present locally, with no live-provider proof that the deletion set is what an operator would expect. Silent remote data loss is the failure mode we can least afford in a Google-surface CLI.

A version we could review: make destructive sync explicit and opt-in (e.g. a `--delete`/`--prune` flag that defaults off), report exactly what would be removed before doing it, and include live proof against a real Apps Script project showing both the non-destructive default and the explicit-delete path. Splitting it into a smaller first PR would also make the review tractable.

haosdent:
Thanks for your reivew! @steipete
I have split it into https://github.com/openclaw/gogcli/pull/1018 and fix according to your comments.

Issue #814, opened 2026-06-15
docs: surface that --fields is accepted as an alias for --select on most commands


privatenumber:
## Summary

The global `--fields` is accepted as an alias for `--select` (JSON output projection) on every
command that doesn't define its own API field-mask `--fields`. This is intentional and helpful
(agents often guess `--fields` for output projection), but it is only surfaced on the `--select`
flag help, so someone reaching for `--fields` directly won't discover it. A one-line note in the
agent / global-flags docs would close the discoverability gap.

## Details (v0.22.0, 197992a)

- The rewrite happens in `internal/cmd/root.go:251-274` (see the explanatory comment at `:251-255`);
  commands with a real field-mask `--fields` are exempted via `commandHasLocalFieldsFlagBefore`
  (`root.go:308`).
- Currently documented only on the `--select` help text (`root.go:44`: "Desire path: use --fields
  for most commands").

## Suggested change (docs only, behavior is correct)

Add a line to the agent-patterns / global-flags documentation, e.g.: "`--fields` is accepted as an
alias for `--select` (output field projection) on commands that don't take an API field mask."

## Environment

gog v0.22.0 (confirmed by reading source at `197992a`). Working as designed; this is a
discoverability/docs request, not a bug.

PR #1007, opened 2026-08-20
docs: document projection order for --results-only and --select

postoso:
## Scope

Documents the existing interaction between `--results-only` and `--select`. No behavior change.

The two global flags are documented separately today, so nothing states that `--results-only` is applied first, or how to project one field from every element of a list response. Reaching for the envelope path `--select items.id` returns `{}`, which is consistent with the best-effort projection contract but is not discoverable from the flag help.

Adds a short paragraph to the two places that describe these flags in prose:

- `docs/automation.md`, next to the existing `--fields` alias note
- `.agents/skills/gog/SKILL.md`, same

Content:

> `--results-only` is applied before `--select`, so a projection runs against the unwrapped primary result. To project each element of a list response, select fields relative to a single element, such as `--results-only --select id`. Dot paths descend through object keys and numeric array indexes; they do not broadcast across every element of a nested array, so an envelope path such as `--select items.id` selects nothing. Object paths that do not match are omitted from the projection rather than reported.

This follows #816, which documented the `--fields` alias in the same two files after #814. #816 also touched `README.md`, but the global-flags list it edited is no longer in the README, and the current "Automate safely" section defers to `docs/automation.md` for output contracts, so I left the README alone.

## Why not a help-string change

`--select` and `--results-only` help text lives in `internal/cmd/root.go`, and editing either regenerates 719 of the 720 pages under `docs/commands/`. That seemed like the wrong trade for a clarification, so this stays in the hand-maintained prose. Happy to move it if you would rather it live in the flag help.

## Testing

- `make docs-check` and `make agent-skills-check`
- Behavior claims verified against `v0.37.0` and a source build of `main` at `eb85a993`, using `gog api list` against the public Discovery directory:

| command | result |
| --- | --- |
| `--select 'items.name'` | `{}` |
| `--select 'items.0.name'` | `{"items.0.name": "..."}` |
| `--results-only --select 'name'` | one `{"name": ...}` per item |
| `--select 'kind,nope'` | `{"kind": "..."}`, unmatched path omitted |

## User-facing changes

Documentation only. No new flags, no behavior change.

### DIFF
--- a/.agents/skills/gog/SKILL.md
+++ b/.agents/skills/gog/SKILL.md
@@ -29,6 +29,13 @@ For JSON output projection, `--fields` is accepted as an alias for `--select` on
 commands that do not define their own API field-mask `--fields`; commands with a
 local field-mask flag keep that command-specific meaning.
 
+`--results-only` is applied before `--select`, so a projection runs against the
+unwrapped primary result. To project each element of a list response, select
+fields relative to a single element, such as `--results-only --select id`. Dot
+paths do not broadcast across the elements of a nested array, so an envelope
+path such as `--select items.id` selects nothing, and object paths that do not
+match are omitted rather than reported.
+
 Pick the account explicitly for API work:
 
 ```bash
--- a/.agents/skills/gog/SKILL.md
+++ b/.agents/skills/gog/SKILL.md
@@ -29,12 +29,10 @@ For JSON output projection, `--fields` is accepted as an alias for `--select` on
 commands that do not define their own API field-mask `--fields`; commands with a
 local field-mask flag keep that command-specific meaning.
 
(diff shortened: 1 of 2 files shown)


steipete:
Landed in 618c33b81409683c8cf3a6f1131d84089d6933c5 after tightening the explanation from 16 added contributor lines to 11 final lines across the automation guide and agent skill.

- Documentation and generated artifacts: `make docs-check agent-skills-check`.
- Independent structured review found no actionable issues.
- Live Google Gmail and Calendar both verified `--json --results-only --select id` returns only item-relative `id` fields.
- The fork's GitHub Actions run was first-time-contributor approval-gated; this docs-only change was independently verified instead of executing an untrusted fork workflow.

Thanks @postoso!

PR #1008, opened 2026-08-20
feat(gmail): reference sanitized body links as [link:N] markers, resolved on demand with gmail link

ronny-rentner:
## Problem

`--sanitize-content` loses every link destination: visible URLs are replaced with
`[url removed]`, and anchor `href`s are dropped with the markup. Sanitized mail therefore
cannot be acted on — a task like "open the invoice" or "unsubscribe me from this" needs the
destination. Keeping URLs inline is no answer either: they can be really long (400–1000
chars each) and would dominate the sanitized body.

## Change

Attachments already solve this shape of problem: the output carries a compact reference,
and the content is fetched on demand. This PR gives URLs the same treatment.

- `[url removed]` in the sanitized body is replaced with `[link:N]` instead, so you can
  reference the link by its index `N` when you actually want to fetch it.
- Indexes are assigned in document order; the same href at several sites shares one index.
- An anchor without visible text (an image-wrapper link) is named by its image's `alt`.
- Script/style content is dropped as before and never numbered.
- New command, mirroring `gmail attachment`:

  ```
  gog gmail link <messageId> <index>
  ```

  It re-runs the same deterministic body conversion, so its indexes are the ones in the
  sanitized body. JSON output is `{"url": ..., "text": ...}` with `text` being the
  anchor text, omitted for a bare URL; a bad index is a usage error.
- The `readonly` and `agent-safe` profiles allow `gmail link`; command docs regenerated;
  `docs/gmail-workflows.md` updated.

## Proof (real mailbox, redacted)

A payment notice whose actions are HTML buttons. Sanitized body on v0.37.0 — the button
texts survive, their destinations are silently gone:

```
... Falls Sie es noch nicht getan haben, fügen Sie einen Beleg hinzu, um Zeit bei der
Buchhaltung zu sparen.   Beleg hinzufügen   Möchten Sie diese Zahlung per Lastschrift
anfechten? Rufen Sie die Zahlung auf und prüfen Sie Ihre Möglichkeiten. ...
```

Same message with this branch:

```
$ gog --version
v0.34.2-38-g1b5fbc1f (1b5fbc1fdc64 2026-08-20T10:42:03Z)
$ gog gmail get MSG_ID --json --sanitize-content
... Falls Sie es noch nicht getan haben, fügen Sie einen Beleg hinzu, um Zeit bei der
Buchhaltung zu sparen.   Beleg hinzufügen [link:0]   Möchten Sie diese Zahlung per
Lastschrift anfechten? Rufen Sie die Zahlung [link:1] auf und prüfen Sie Ihre
Möglichkeiten. ...
$ gog gmail link MSG_ID 0 --json
{
  "url": "https://url2820.qonto.com/ls/click?upn=u001.PALJKS45jlsSqfTO[...redacted, 1023 chars total...]",
  "text": "Beleg hinzufügen"
}
```

The resolved URL is byte-identical to the anchor's `href` in the message's HTML. No URL
appears anywhere in the sanitized output; the original example message carries six long
tracked URLs which would otherwise add ~5 kB to every read.
Before this change, needing to open even a single one of these links meant giving up on
sanitizing entirely and reading the message unsanitized, with all URLs and the raw
payload, just to get at one `href`.

## Testing

- Unit: body conversion (anchor text + marker, bare-URL replacement, `mailto`, duplicate
  href sharing one index, image-anchor `alt` naming, script/style URLs never numbered).
- Execute-level: sanitized `gmail get` emits markers and no URLs; `gmail link` resolves
  index → exact URL and anchor text; non-numeric / out-of-range indexes are usage errors;
  index parity between `get` and `link` on one fixture; thread output numbers per message.
- Local gate: `make fmt-check`, `make lint`, `make deadcode`, `make docs-check`, `make
  test` all pass.

User-facing changes: the new `gmail link` command; sanitized bodies now mark links as
`[link:N]` instead of `[url removed]`/dropping them.

### DIFF
--- a/docs/commands.generated.md
+++ b/docs/commands.generated.md
@@ -422,6 +422,7 @@ Generated from `gog schema --json`.
       - [`gog gmail (mail,email) labels (label) modify (update,edit,set) <threadId> ... [flags]`](commands/gog-gmail-labels-modify.md) - Modify labels on threads
       - [`gog gmail (mail,email) labels (label) rename (mv) <labelIdOrName> <newName>`](commands/gog-gmail-labels-rename.md) - Rename a label
       - [`gog gmail (mail,email) labels (label) style (color,colour) <labelIdOrName> [flags]`](commands/gog-gmail-labels-style.md) - Change a user label color or visibility
+    - [`gog gmail (mail,email) link <messageId> <index>`](commands/gog-gmail-link.md) - Resolve a [link:N] marker from sanitized output to its full URL
     - [`gog gmail (mail,email) mark-read (read-messages) [<messageId> ...] [flags]`](commands/gog-gmail-mark-read.md) - Mark messages as read
     - [`gog gmail (mail,email) messages (message,msg,msgs) <command>`](commands/gog-gmail-messages.md) - Message operations
       - [`gog gmail (mail,email) messages (message,msg,msgs) modify (update,edit,set) <messageId> [flags]`](commands/gog-gmail-messages-modify.md) - Modify labels on a single message
--- a/docs/commands.generated.md
+++ b/docs/commands.generated.md
@@ -422,7 +422,7 @@ Generated from `gog schema --json`.
       - [`gog gmail (mail,email) labels (label) modify (update,edit,set) <threadId> ... [flags]`](commands/gog-gmail-labels-modify.md) - Modify labels on threads
       - [`gog gmail (mail,email) labels (label) rename (mv) <labelIdOrName> <newName>`](commands/gog-gmail-labels-rename.md) - Rename a label
       - [`gog gmail (mail,email) labels (label) style (color,colour) <labelIdOrName> [flags]`](commands/gog-gmail-labels-style.md) - Change a user label color or visibility
-    - [`gog gmail (mail,email) link <messageId> <index>`](commands/gog-gmail-link.md) - Resolve a [link:N] marker from sanitized output to its full URL
+    - [`gog gmail (mail,email) link <messageId> <index> [flags]`](commands/gog-gmail-link.md) - Resolve a [link:N] marker from sanitized output to its full URL
     - [`gog gmail (mail,email) mark-read (read-messages) [<messageId> ...] [flags]`](commands/gog-gmail-mark-read.md) - Mark messages as read
     - [`gog gmail (mail,email) messages (message,msg,msgs) <command>`](commands/gog-gmail-messages.md) - Message operations
       - [`gog gmail (mail,email) messages (message,msg,msgs) modify (update,edit,set) <messageId> [flags]`](commands/gog-gmail-messages-modify.md) - Modify labels on a single message
--- a/docs/commands/README.md
+++ b/docs/commands/README.md
@@ -2,7 +2,7 @@
 
 Every `gog` command has a generated docs page. The source of truth is the live CLI schema; run `make docs-commands` after changing command names, flags, help text, aliases, or arguments.
 
-Generated pages: 719.
+Generated pages: 720.
 
 ## Top-level Commands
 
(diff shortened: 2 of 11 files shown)


ronny-rentner:
Thanks for the review. Here is the status on each finding:

**[P1] Read links from the HTML alternative** — real, but pre-existing; proposed as
follow-up. Before this PR, sanitized output contained no reachable destination in any
mail: every URL was removed or dropped. The plain-first part selection is existing
behavior (`BestBodyForDisplay`), unchanged here, and a link present only in the HTML
alternative was equally unreachable before. This PR adds resolvability for every link in
the converted part and removes nothing. Extending coverage to the HTML alternative means
changing which part sanitize converts for multipart mail (or adding a selection option
whose value must match between `get` and `link`) — a sanitize-wide design decision that
deserves its own change rather than riding on this one.

**[P1] Reserve generated marker syntax** — disputed. A sender who can write a literal
`[link:0]` into their mail can equally write a real anchor with deceptive anchor text
and produce an indistinguishable result through the legitimate path — marker-adjacent
text is sender-controlled either way, with or without this feature. Resolution semantics
are unaffected by body text: index N always returns the N-th captured link, and a
fabricated out-of-range index is a clean usage error. The defense against deceptive mail
content is the existing untrusted-content handling, not marker syntax.

**[P1] Sanitize untrusted values before line output** — fixed in 20663861. The non-JSON
path now quotes both values with `tsvSafeValue`, as the attachment printer does; a
regression test covers an href carrying an entity-encoded newline and tab. The JSON path
already escapes control characters per the JSON spec.

**[P2] Preserve bare URL boundaries** — fixed in 4a0d97c2. `)` is URL-legal and now
stays in a captured bare URL like any other character. Nothing is trimmed: where a bare
URL ends inside prose is context no pattern can judge, and a reader can drop a trailing
character it recognizes as punctuation, while nobody can restore a character the pattern
cut off. When a text-captured URL ends in `.` or `)`, resolution adds a `note` that the
trailing character may be sentence punctuation; anchor hrefs are byte-exact and get no
note. Tests cover balanced parentheses, prose parentheses, and the note.

**Resolve security concern: untrusted link values can forge CLI records** — fixed by the
same change as the line-output finding above (20663861).

**Resolve merge risk: compatibility** — the shape change is disclosed in the PR body and
is the maintainer's call; if you want the markers gated behind an option instead, say so.

ronny-rentner:
On the punctuation finding: trimming before registering the marker would alter the
sanitized body and destroy characters nothing downstream can restore, and a trailing `.`
or `)` can be legitimate URL data, so `url` keeps the exact capture. The note is gone;
instead `gmail link --trim-punctuation` returns the directly usable value on request:
trailing sentence punctuation stripped (`.` `,` `;` `:` `!` `?`, and `)` only while the
URL's parentheses are unbalanced, since a balanced pair is likely path data). Anchor
hrefs are byte-exact and unaffected by the flag. Unit tests cover the punctuation set,
balanced parentheses, mixed trailing runs, and the degenerate cases.

steipete:
Closing: this changes the sanitized-output contract, and review found two concrete gaps. It misses links in common multipart messages (so the sanitized view is incomplete on exactly the mail shape people hit most), and the marker syntax it introduces can be forged by a sender — meaning message content can influence how our own output is parsed downstream.

Sanitized output is a trust boundary here, so the bar is that a sender must never be able to inject or imitate our markers, and coverage has to hold across multipart/alternative and multipart/related bodies. Happy to review a revision that pins both down with fixtures for the multipart cases.

ronny-rentner:
Hi Pete, thanks again for the review.

The forged-marker point is an easy fix and it's already done on the branch: a literal
`[link:N]` is now reserved the same way `internal/outfmt/untrusted.go` neutralizes a forged
marker, so message text can't pose as one of ours.

On the missed links, one thing I want to check before building it. This PR is really the
`[link:N]` references plus the `gmail link` command — the plain-first body selection it
inherits is the sanitized view's existing behavior, not something this PR introduced
(before it, that same plain body just showed `[url removed]`). So reaching links that live
only in the HTML part means changing what the sanitized view reads, which is a bit broader
than what this PR set out to do, and I'd rather not guess at the contract you want.

The way I'd lean is an opt-in switch on the sanitized reads (`gmail get` / `gmail thread
get` / `gmail link`) that prefers the HTML body — the same idea as the `--body-format html`
you already have on `gmail messages`. The default stays plain-first, so nothing changes for
existing reads, and anyone who wants the HTML links (and their markers) opts in.

Is that opt-in what you have in mind? Or do you actually want the sanitized view to prefer
HTML by default — i.e. change what it shows for plain+HTML mail? Happy to build either,
just want to match what you're after.

PR #1011, opened 2026-08-21
fix(docs): harden HTML tag stripping

vincentkoc:
## Summary

- replace two one-shot HTML-tag-removal regexes with one shared single-pass tag scanner
- preserve literal less-than text while removing complete, unclosed, and unterminated tag-like markup
- keep docs heading anchors and generated table-of-contents text on the same cleanup boundary
- document the security hardening in Unreleased

## Root cause

Both docs scripts removed multi-character HTML tag matches once. CodeQL 2.26.3 correctly treats that pattern as incomplete sanitization because removal can expose another dangerous tag sequence.

`scripts/html-text.mjs` now owns tag removal with a single-pass scanner instead of a replacement regex. It tracks quoted attributes, removes genuinely unterminated tag fragments, prevents discarded tags from rebuilding markup across adjacent `<` boundaries, and preserves ordinary text such as `1 < 2`. Both heading-anchor validation and generated TOC text use that owner, so the invariant cannot drift between the checker and renderer.

## CodeQL

Fixes:

- https://github.com/openclaw/gogcli/security/code-scanning/5
- https://github.com/openclaw/gogcli/security/code-scanning/6

## Validation

- `node --test scripts/check-docs-coverage.test.mjs`
- `node scripts/build-docs-site.mjs`
- rendered TOC probe across `dist/docs-site`
- `make fmt`
- `git diff origin/main...HEAD --check`
- privacy scrub of the complete PR diff

Production scripts: +36/-6 (net +30, shared security invariant with explicit malformed-input and reconstruction handling) | Tests: +15/-0 | Changelog: +1/-0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Security: remove overflow-prone capacity arithmetic from untrusted-output maps and versioned tracking ciphertext construction while preserving their wire formats.
 - Security: restrict the CI workflow's `GITHUB_TOKEN` to read-only repository contents instead of inheriting broader organization or repository defaults.
 - Security: generate live agent-evaluation session IDs with cryptographically random suffixes while preserving their fixed-width format.
+- Security: replace docs heading and table-of-contents HTML sanitization regexes with one shared single-pass tag scanner so removed markup cannot reconstruct unsafe tags.
 - Security: bound `gmail watch serve` and the local OAuth callback HTTP servers with read, idle, and header-size limits so a slow or oversized request cannot stall the listener.
 
 ## v0.37.0 - 2026-08-14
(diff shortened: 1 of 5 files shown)


vincentkoc:
@clawsweeper re-review

Exact head `083a5edc4577af56141f36dc6255b82cc819713e` replaces the regex loop with a single-pass tag scanner. Focused tests now explicitly cover the previously ambiguous unclosed element, a genuinely unterminated tag, quoted `>` attributes, nested markup, and literal `1 < 2` text.

vincentkoc:
@clawsweeper re-review

Exact head `a9fc21a6835b3bfb27f242bcfa63813e2c5a57a5` addresses the prior P1 boundary finding. Adjacent pending `<` characters are discarded when a recognized tag begins, and the focused regression proves `<<<script>img>` becomes `img>` rather than rebuilding ``. Unterminated-tag, quoted-attribute, nested-markup, and literal `1 < 2` cases still pass.

PR #1012, opened 2026-08-21
fix(evals): use secure session identifiers

vincentkoc:
## Summary

- replace the predictable `Math.random()` suffix with four cryptographic random bytes
- preserve the existing `gog-eval-<cli>-<timestamp>-<8 lowercase hex>` session ID shape
- cover the exact fixed-width format and random-byte count
- document the security hardening in Unreleased

## Root cause

The live eval harness used `Math.random()` in an OpenClaw session identifier. That identifier selects session state, so predictability crosses a security boundary even though the harness itself is an evaluation tool.

Four bytes from Node's built-in `crypto.randomBytes()` remove the weak generator without adding a dependency. Encoding those bytes as lowercase hex preserves the historical eight-character suffix, so existing OpenClaw session filenames and explicit `--session-id` handling keep the same shape instead of changing to a UUID.

## Compatibility proof

OpenClaw accepts explicit session IDs as trimmed non-empty strings, uses exact string equality for stored-session lookup, and embeds the supplied value into its synthetic explicit-session key. It does not require a UUID or impose a fixed length:

- https://github.com/openclaw/openclaw/blob/e66d7c5cac65fa9391904648eb81861276bc5725/src/agents/command/session.ts#L123-L129
- https://github.com/openclaw/openclaw/blob/e66d7c5cac65fa9391904648eb81861276bc5725/src/agents/command/session.ts#L198-L214
- https://github.com/openclaw/openclaw/blob/e66d7c5cac65fa9391904648eb81861276bc5725/src/agents/command/session.ts#L313-L318

The focused test also proves that the generated ID retains the historical `gog-eval-gog-1700000000000-01234567` shape while sourcing all eight suffix characters from four cryptographic bytes.

## CodeQL

Fixes:

- https://github.com/openclaw/gogcli/security/code-scanning/7

## Validation

- `node --test scripts/eval-gws-agents.test.mjs`
- `make fmt`
- `git diff origin/main...HEAD --check`
- privacy scrub of the complete PR diff

Production LOC: +6/-1 (net +5, security invariant and testable owner boundary) | Tests: +14/-0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Security: remove overflow-prone capacity arithmetic from untrusted-output maps and versioned tracking ciphertext construction while preserving their wire formats.
 - Security: restrict the CI workflow's `GITHUB_TOKEN` to read-only repository contents instead of inheriting broader organization or repository defaults.
+- Security: generate live agent-evaluation session IDs with cryptographically random suffixes while preserving their fixed-width format.
 - Security: bound `gmail watch serve` and the local OAuth callback HTTP servers with read, idle, and header-size limits so a slow or oversized request cannot stall the listener.
 
 ## v0.37.0 - 2026-08-14
--- a/scripts/eval-gws-agents.mjs
+++ b/scripts/eval-gws-agents.mjs
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 
 import { spawn } from "node:child_process";
+import { randomBytes } from "node:crypto";
 import { constants as fsConstants } from "node:fs";
 import { access, chmod, mkdir, mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
 import net from "node:net";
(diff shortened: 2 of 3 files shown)


vincentkoc:
@clawsweeper re-review

Exact head `bd032f644c062199174edae7fd9daa156f32fa40` now uses four cryptographic random bytes while preserving the historical fixed-width eight-character hexadecimal session suffix. Focused tests assert the exact session shape and byte count.

vincentkoc:
Maintainer resolution for the exact-head ClawSweeper review at `bd032f644c062199174edae7fd9daa156f32fa40`:

- accepted: the review confirms the cryptographic suffix implementation and focused regression test are correct
- accepted: the UUID compatibility risk is resolved by preserving the historical eight-character hexadecimal suffix; the PR body links the exact OpenClaw explicit-session handling
- skipped: removing the Unreleased changelog entry. Repository landing policy requires a meaningful changelog entry, and this security-relevant behavior change belongs in Unreleased

No code finding or compatibility risk remains.

PR #1019, opened 2026-08-24
Opened in error — please disregard

mm-yuya:
This PR was opened in error and has been closed. Please disregard it. Apologies for the noise.

### DIFF
--- a/internal/cmd/gmail_drafts.go
+++ b/internal/cmd/gmail_drafts.go
@@ -23,6 +23,7 @@ type GmailDraftsCmd struct {
 	Send     GmailDraftsSendCmd     `cmd:"" name:"send" aliases:"post" help:"Send a draft"`
 	Create   GmailDraftsCreateCmd   `cmd:"" name:"create" aliases:"add,new" help:"Create a draft"`
 	Update   GmailDraftsUpdateCmd   `cmd:"" name:"update" aliases:"edit,set" help:"Update a draft"`
+	Find     GmailDraftsFindCmd     `cmd:"" name:"find" help:"Find the reply draft addressed to a message (bounded owner-side recovery)"`
 	Reply    GmailDraftsReplyCmd    `cmd:"" name:"reply" help:"Save a reply as a draft"`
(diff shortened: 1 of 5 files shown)


mm-yuya:
Updated per the review findings (thank you — both were real):

1. **Recipients.** The create tool now wraps the dedicated `gmail drafts reply --no-quote` command instead of `drafts create --reply-to-message-id`. The previous form set only thread headers and saved a recipientless draft; the reply command derives recipients from the original message exactly as the send-side reply does.

2. **No silent surface widening.** Both tools are now exact-opt-in via a new `ExactOptIn` spec flag: they are invisible to `gmail`, `gmail.*`, `write`, `read`, `all`, `*`, and the default read set, and appear only when named literally in the allowlist (the write tool additionally requires `--allow-write`). Upgrading gog can no longer add a Gmail write tool to any existing configuration.

3. **Proof at the protocol level.** Added an in-process MCP client test showing unknown fields (`to`, `reply_all`, `attach`, `args`), wrong types, and missing required fields are rejected by the server before the tool handler runs — the handler call count is asserted.

Measured on the branch: no selector short of the literal names surfaces either tool; with exact names, the five Gmail tools appear and nothing matching send/delete/reply-all/forward exists anywhere in the registry.

mm-yuya:
Round 3 update — added owner-side recovery so a lost draft id can't cause a duplicate.

`gmail_get_draft` now takes either `draft_id` (read a known draft) or `recover_reply_to_message_id` (recover the single reply draft addressed to that message when its id was lost between create and recording). Recovery maps to a new bounded CLI command `gmail drafts find`, which scans at most `--max` drafts and matches on the original message's RFC Message-ID and thread — deterministic, owner-side, not a local cache. Exactly one input is required; both or neither is rejected.

Everything from the previous update stands: dedicated `drafts reply --no-quote` for addressed drafts, exact-opt-in visibility, and protocol-level unknown-field rejection.

mm-yuya:
Closing — opened without the required authorization on my side. Apologies for the noise.

PR #1020, opened 2026-08-25
feat(calendar): explicitly disable event reminders

### LINKED ISSUE #1002: Calendar: no way to create or update an event with NO reminders (useDefault:false + empty overrides)
### Problem

There is no way to create or update a calendar event with **no reminders**.

The Calendar API distinguishes three states:

| `reminders` payload | Meaning |
|---|---|
| absent | use the calendar's `defaultReminders` |
| `{"useDefault": false, "overrides": [{...}]}` | these specific reminders |
| `{"useDefault": false, "overrides": []}` | **no reminders at all** |

`gog` can express the first two. The third is unreachable, because an empty
`--reminder` is treated as "unspecified" rather than as "none".

### Why it matters

Google applies a calendar's single `defaultReminders` list to **all-day** events
created via the API, ignoring the separate "All-day event notifications" setting
shown in the web UI. So an all-day event created by `gog` on a calendar whose
timed default is `popup 10m` inherits that popup — and a 10-minute popup on an
all-day event fires at 23:50 the **evening before**.

There is no way to avoid this from `gog`: clearing the all-day setting in the UI
has no effect (the API path uses `defaultReminders`), and the timed default has
to stay for real meetings.

### Reproduction

```
# a calendar whose defaultReminders is [{popup, 10}]
gog calendar create primary --summary "all-day marker" --all-day \
    --from 2030-05-01 --to 2030-05-02 --json -n
```

The request body contains no `reminders` key, so the default is inherited:

```json
"event": {
  "end":     {"date": "2030-05-02"},
  "start":   {"date": "2030-05-01"},
  "summary": "all-day marker"
}
```

Attempting to suppress it:

```
gog calendar create primary ... --reminder ""    # -> no reminders field at all
gog calendar update primary <id> --reminder ""   # -> {"useDefault": true}
```

The update case is the surprising one: `--reminder ""` documents as "Set empty to
clear", but it sets `useDefault: true`, which restores the calendar default
rather than clearing reminders.

### Cause

`buildReminders` in `internal/cmd/calendar_build.go` returns `nil` for an empty
input, and `nil` means "use calendar defaults":

```go
//nolint:nilnil // nil return is intentional: nil means "use calendar defaults"
func buildReminders(reminders []string) (*calendar.EventReminders, error) {
	if len(reminders) == 0 {
		return nil, nil
	}
	var filtered []string
	for _, r := range reminders {
		if strings.TrimSpace(r) != "" {
			filtered = append(filtered, r)
		}
	}
	if len(filtered) == 0 {
		return nil, nil
	}
	...
```

`EventReminders{UseDefault: false, Overrides: overrides}` is only ever
constructed with at least one override, so `overrides: []` is unreachable. In
`calendar_update_patch_plan.go` a `nil` becomes `{UseDefault: true}`.

Confirmed identical on **0.12.0** and **0.37.0**.

### Suggested fix

A `--no-reminder` / `--reminders-none` flag (or a reserved `--reminder none`)
that produces:

```json
"reminders": {"useDefault": false, "overrides": []}
```

with `UseDefault` in `ForceSendFields`, exactly as the existing override path
already does.

### Workaround

Calling the API directly with that body works — Google accepts an empty
overrides list without complaint, so the limitation is purely in how the flag is
interpreted:

```
PATCH https://www.googleapis.com/calendar/v3/calendars/{calendarId}/events/{eventId}
{"reminders": {"useDefault": false, "overrides": []}}
```

### Version

```
gog v0.37.0 (Homebrew), macOS arm64
```

steipete:
Fixes #1002.

Add mutually exclusive `--no-reminders` flags to Calendar event creation and updates. Explicitly send `{"overrides":[],"useDefault":false}` to suppress all reminders while preserving empty `--reminder=` as the separately tested restore-calendar-defaults operation.

Verified against the designated live Google account with a disposable event: create without reminders, set a ten-minute custom reminder, update back to no reminders, restore the account's independently verified thirty-minute calendar default, and delete the event.

Includes exact JSON payload regressions, flag-conflict coverage, generated command documentation, and the full repository CI gate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Calendar: add `--no-reminders` to event creation and updates so calendar-default reminders can be explicitly disabled. (#1002)
 - Slides: add skip/unskip commands and expose slide visibility without changing existing plain-text output. (#1009) — thanks @marnunez.
 - Config: serialize concurrent updates with shared platform-native file locks, preventing intermittent access-denied failures on Windows.
 - Calendar: correctly clear custom reminder overrides when restoring calendar defaults. (#1016) — thanks @sorenisanerd.
--- a/docs/commands/gog-calendar-create.md
+++ b/docs/commands/gog-calendar-create.md
@@ -49,6 +49,7 @@ gog calendar (cal) create (add,new) <calendarId> [flags]
 | `--location` | `string` |  | Location |
 | `--location-search` | `string` |  | Resolve a Google Places text search and use the best match as event location |
 | `--no-input`<br>`--non-interactive`<br>`--noninteractive` | `bool` |  | Never prompt; fail instead (useful for CI) |
+| `--no-reminders` | `bool` |  | Disable all event reminders |
 | `--ooo-auto-decline` | `string` |  | Out of Office auto-decline mode: none, all, new |
 | `--ooo-decline-message` | `string` |  | Out of Office decline message |
 | `--place-id` | `string` |  | Resolve a Google Places ID and use it as event location |
--- a/docs/commands/gog-calendar-update.md
+++ b/docs/commands/gog-calendar-update.md
@@ -50,6 +50,7 @@ gog calendar (cal) update (edit,set) <calendarId> <eventId> [flags]
 | `--location` | `string` |  | New location (set empty to clear) |
 | `--location-search` | `string` |  | Resolve a Google Places text search and use the best match as event location |
 | `--no-input`<br>`--non-interactive`<br>`--noninteractive` | `bool` |  | Never prompt; fail instead (useful for CI) |
+| `--no-reminders` | `bool` |  | Disable all event reminders |
 | `--ooo-auto-decline` | `string` |  | Out of Office auto-decline mode: none, all, new |
 | `--ooo-decline-message` | `string` |  | Out of Office decline message (set empty to clear) |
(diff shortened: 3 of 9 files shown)


steipete:
Implemented and landed `--no-reminders` for both Calendar event creation and updates.

Proof:

- Exact disable payload: `{"overrides":[],"useDefault":false}`.
- Existing restore-defaults payload remains `{"overrides":null,"useDefault":true}`.
- Focused regressions: `go test ./internal/cmd -run 'Test(BuildReminders|CalendarUpdatePatch(ClearsReminders|DisablesReminders)|CalendarReminderFlagsAreMutuallyExclusive)' -count=1`.
- Full local repository gate: `make ci`, including lint, Go/JavaScript tests, generated command documentation, and agent skills.
- Live Google Calendar on the designated test account: independently fetched the primary calendar's default thirty-minute popup; created a disposable event without reminders; independently read back no reminders; added and verified a ten-minute custom reminder; disabled and independently verified it again; restored and independently verified the actual thirty-minute calendar default; deleted the disposable event.
- Independent structured review: no actionable findings.
- Hosted cross-platform CI: https://github.com/openclaw/gogcli/actions/runs/32830995351.
- Hosted Docker CI: https://github.com/openclaw/gogcli/actions/runs/32830995297.

Landed via https://github.com/openclaw/gogcli/pull/1020 at 5f4e545d0bc8f804ee089bd80925b1efcae23059.

Issue #938, opened 2026-07-23
feat(sheets): support Connected Sheets (BigQuery data sources)


ryo-touch:
I run several reporting spreadsheets wired to BigQuery via Connected Sheets. Keeping the underlying SQL in sync and forcing a refresh from a script (cron / CI / an agent) currently requires opening the Sheets UI by hand — `gog` can't touch the data source side of a spreadsheet at all.

## The gap

None of the ~35 `sheets` subcommands touch `DataSource`, `BigQueryDataSourceSpec`, or data execution status, and nothing in the docs mentions it either.

It's not just "no write path" — read is barely there either:

- `Spreadsheet` has top-level `dataSources[]` and `dataSourceSchedules[]` fields (per the Sheets API reference), but `SheetsMetadataCmd`'s `Fields(...)` mask only requests `spreadsheetId,title,locale,timeZone,sheets` — `dataSources` is silently dropped.
- The only way to even see a connected sheet's `dataSourceId` today is an unfiltered `gog sheets raw <id>` dump. There's no structured `list`, and no way to inspect `dataExecutionStatus` to tell whether a refresh succeeded, failed, or is still running.
- There's no command to create a data source, change its query/table, delete it, or force a refresh — so any Connected Sheet has to be built and maintained by hand in the browser.

## What already exists to build on

The underlying Sheets API v4 supports the full lifecycle via `spreadsheets.batchUpdate`:

- `AddDataSourceRequest` — attach a new BigQuery data source (custom SQL via `BigQueryQuerySpec`, or a native table via `BigQueryTableSpec`) and create the linked `DATA_SOURCE` sheet
- `UpdateDataSourceRequest` — change an existing data source's spec
- `DeleteDataSourceRequest` — remove a data source and its linked sheet
- `RefreshDataSourceRequest` — force/queue a re-execution of the data source's query

All of these types already exist in `google.golang.org/api/sheets/v4`, the SDK version gogcli is already pinned to (`v0.286.0`) — this is missing CLI surface, not an SDK gap.

## Proposed CLI surface

```
gog sheets datasource list <spreadsheetId>
gog sheets datasource add <spreadsheetId> --project <gcpProjectId> --query <sql>
gog sheets datasource add <spreadsheetId> --project <gcpProjectId> --table-project <p> --dataset <d> --table <t>
gog sheets datasource refresh <spreadsheetId> <dataSourceId> [--force]
gog sheets datasource delete <spreadsheetId> <dataSourceId>
```

[...]

PR #1023, opened 2026-08-25
docs(sheets): describe Connected Sheets BigQuery authorization accurately

ryo-touch:
Refs [#938](https://github.com/openclaw/gogcli/issues/938).

`docs/sheets-connected.md` makes two claims about BigQuery authorization that are not true, both shipped in 0.37. This corrects them and nothing else.

## What is wrong today

1. "Google requires `https://www.googleapis.com/auth/bigquery.readonly`". Google requires BigQuery read access. `bigquery.readonly` is the least-privilege scope that grants it, and that is why the Connected Sheets client asks for exactly that one — but it is not the only grant Google accepts.
2. Re-consent is presented as unconditional. An account whose stored grant already covers BigQuery read access needs no further consent round.

The second one matters in practice: a reader with a broadly-scoped token is told to re-authorize, and following that instruction on a token covering many services is a real cost for no gain.

## Why (2) is true

The Connected Sheets client performs no local scope check. `NewConnectedSheets` goes through `optionsForAccountScopes`, which passes `requireStoredGrant=false`, so the stored-grant comparison in `tokenSourceForAccountScopesWithStoredScopeCheck` (`internal/googleapi/client_auth.go:437`) is bypassed. The stored refresh token is used as-is and the request is authorized by Google against whatever the account already holds, not against the literal scope strings gog names. `InsufficientScopeError` is only ever constructed inside that skipped branch.

## On documenting accepted supersets

You asked for a reproducible redacted account-scope listing plus successful Connected Sheets output for the `cloud-platform` case. I could not produce evidence that actually isolates the variable, and I would rather say so than attach a listing that looks like proof and is not.

Google accumulates granted scopes. My account's token now holds `bigquery.readonly` alongside `cloud-platform`, so any listing from it is consistent with either explanation and proves nothing about `cloud-platform` on its own. Getting a token with `cloud-platform` but without `bigquery.readonly` means revoking the application's authorization and re-consenting from scratch — on a working token covering 22 services — and I have no separate account I can grant narrowly instead.

So this PR deliberately does **not** enumerate accepted supersets. It states the mechanism (gog does not gate locally; Google decides) and leaves `bigquery.readonly` as the documented least-privilege default. If you want the superset named explicitly, that needs evidence from an account that can be granted narrowly, and I would rather leave it undocumented than assert it from inference.

This also walks back part of what I claimed in [#938](https://github.com/openclaw/gogcli/issues/938#issuecomment-5312315264): that my read-path validation ran on a `cloud-platform` token without re-consent. I cannot now establish whether `bigquery.readonly` was already on that token at the time, so please treat the superset half of that report as unverified. The read-path findings themselves were reproduced and fixed in [#1001](https://github.com/openclaw/gogcli/pull/1001) and are unaffected.

## Scope

Docs only, one file, no command or generated-page changes. `make docs-check` passes (719 command pages, 27 feature pages).

The lifecycle PRs are unaffected by the above — a throwaway spreadsheet and scratch dataset do not need a separate account — and I am starting with `refresh` as you asked.

### DIFF
--- a/docs/sheets-connected.md
+++ b/docs/sheets-connected.md
@@ -22,7 +22,7 @@ gog auth add you@example.com \
 
 If the account token covers more services, keep that existing `--services` selection instead of narrowing it to `sheets`. Domain-wide delegated service accounts must also have both the Sheets read-only and BigQuery read-only scopes approved by the Workspace administrator; the Connected Sheets client requests only those two scopes.
 
-gog does not compare the stored grant against those scopes before calling Google, so a request is authorized by Google against whatever the account's existing grant covers rather than against the literal scope strings above. An account already holding BigQuery read access therefore needs no further consent round. Which broader scopes qualify is Google's determination; `bigquery.readonly` stays the documented least-privilege default.
+For a stored user OAuth token, the Connected Sheets client does not compare the grant against those scopes before calling Google, so the request is authorized against whatever the account already holds rather than against the literal scope strings above. Such an account, if it already holds BigQuery read access, needs no further consent round. Which broader scopes qualify is Google's determination; `bigquery.readonly` stays the documented least-privilege default. Domain-wide delegation works the other way: the scopes are asserted in the signed JWT, so the administrator must have approved exactly those.
 
 Looker data sources reuse the account's existing Looker link, but the same commands and output shape apply.
 
--- a/docs/sheets-connected.md
+++ b/docs/sheets-connected.md
@@ -9,9 +9,9 @@ description: Inspect BigQuery and Looker data sources, execution status, and anc
 
 ## Authorize BigQuery access explicitly
 
-Google requires `https://www.googleapis.com/auth/bigquery.readonly` whenever a Sheets API response contains BigQuery Connected Sheets data. Ordinary `sheets` authorization intentionally does not request that scope.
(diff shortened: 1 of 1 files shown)


steipete:
Landed in 1251bfc2c527779e003dffc59afcc66836e384b5. Thanks @ryo-touch!

The final wording keeps Google's explicitly documented `bigquery.readonly` requirement alongside Sheets authorization, makes reauthorization conditional on those permissions not already being present, and preserves the account's existing service selection. I removed the unverified broader-scope equivalence and the overly strict domain-wide-delegation assertion; Google's Connected Sheets guide remains the source of truth: https://developers.google.com/workspace/sheets/api/guides/connected-sheets.

Verification:

- `make docs-check` with the existing shared tool directory — generated documentation, site checks, and coverage all passed.
- `go test ./internal/googleapi -run '^TestNewConnectedSheetsRequestsReadOnlySheetsAndBigQueryScopes$' -count=1` — verifies the client requests both read-only Sheets and BigQuery scopes.
- Independent review passed with no actionable findings.
- Exact-head hosted CI passed Linux, Windows, macOS/cgo, and the worker: https://github.com/openclaw/gogcli/actions/runs/32867962894.

Issue #938 intentionally remains open for the separately scoped Connected Sheets lifecycle work.

PR #1026, opened 2026-08-25
feat(sheets): add BigQuery Connected Sheets data sources

steipete:
## Summary

Add the next independently scoped Connected Sheets lifecycle command:

```
gog sheets datasource add SPREADSHEET_ID --billing-project PROJECT --query SQL
gog sheets datasource add SPREADSHEET_ID --billing-project PROJECT --dataset DATASET --table TABLE [--table-project PROJECT]
```

- Require the caller to name the billing-enabled execution project explicitly; never infer one or confuse the existing global `--project` JSON-selection alias with billing.
- Support exactly one SQL or native-table source; omit the optional table-project field when Google should default it to the billing project.
- Reuse the dedicated isolated Connected Sheets writer and extract a small shared, readonly-enforced, single-attempt mutation path also used by refresh.
- Preview execution and billing risk without exposing SQL; never echo provider-returned raw SQL in normal JSON output.
- Preserve newly created source IDs and wrapped provider execution errors even when execution immediately fails; fail closed on ambiguous provider replies without retrying potentially successful creation.
- Credit @ryo-touch while replacing the original broad lifecycle patch with the requested focused creation stage.

Part of #938; issue remains open for update, delete, and a successful billed BigQuery provider proof.

## Validation

- Full local `make ci`.
- Targeted fixtures cover SQL sources, same-project and cross-project native tables, exact request shapes, malformed/ambiguous provider replies, structured failed execution, SQL-redacted output and dry runs, explicit billing selection, incompatible/empty arguments, readonly before account/authentication, and exactly one request through the real retry transport.
- Real `clawdbot@gmail.com` provider run created a disposable spreadsheet, verified a SQL-redacted billing-aware dry run, verified readonly rejection, submitted an actual `addDataSource` request with a deliberately nonexistent billing project and constant `SELECT 1`, received Google's real missing-`bigquery.readonly` 403 with actionable account-preserving guidance, confirmed the spreadsheet still contained zero data sources, and trashed the disposable spreadsheet.
- Positive BigQuery source creation is not claimed: it additionally requires user consent for `bigquery.readonly` and an explicitly approved billing-enabled BigQuery execution project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Sheets: add explicitly billed BigQuery Connected Sheets data sources from SQL queries or native tables without exposing SQL in command output. (#938) — thanks @ryo-touch.
 - Sheets: refresh individual Connected Sheets data sources with scoped BigQuery authorization, dry-run safety, and structured execution status. (#938) — thanks @ryo-touch.
 - Chat: include user-mention annotations and emoji-reaction summaries in message-list JSON without changing existing text output. (#1000) — thanks @Ben-Living.
 - Apps Script: safely pull project source and list deployments and versions without mutating remote projects. (#1018) — thanks @haosdent.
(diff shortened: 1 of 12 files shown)


steipete:
Landed as 540790794ed04e20cabbb640fc867564b7edc79a in #1026. The final version requires an explicitly selected billing project, safely supports SQL or native-table sources, suppresses SQL from output/dry runs, and never automatically retries creation. Full local/hosted CI passed; live Google testing reproduced the actual missing-`bigquery.readonly` 403 and verified disposable-sheet cleanup. Thanks @ryo-touch.

PR #1032, opened 2026-08-25
feat(gmail): send prebuilt RFC822 messages

higginz777:
## Summary
- add `gog gmail send --raw-file <path|->` for exact prebuilt RFC822 messages
- preserve raw bytes and return Gmail provider message/thread IDs through the existing send result
- keep the operation behind existing Gmail send policy, readonly, alias, and safety controls
- reject compose, reply, tracking, signature, and attachment flags in raw mode
- make dry-run offline and content-redacted (source, byte count, SHA-256 only)

## Why
Approval-gated clients need to freeze and authorize one exact MIME message, including a deterministic Message-ID, then reconcile the provider result without rebuilding the message inside the CLI.

## Tests
- `make ci`
- 728 generated command docs verified
- raw-file tests cover exact file/stdin bytes, pre-auth validation, dry-run redaction, conflicting flags, readonly, no-send policy, and alias enforcement

## User-facing changes
`gog gmail send` accepts `--raw-file`; optionally combine it with `--thread-id` for Gmail thread placement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 0.38.1 - Unreleased
 
+- Gmail: add guarded exact-byte RFC822 sending from a file or stdin, with optional thread binding, compose-mode isolation, and content-safe offline dry runs.
+
 ## 0.38.0 - 2026-08-25
 
 - Sheets: add the full BigQuery Connected Sheets lifecycle—create, update, refresh, and safely delete data sources—with explicit billing and authorization, protected SQL previews, correct source matching, and reliable extract reads. (#938, #1001) — thanks @ryo-touch.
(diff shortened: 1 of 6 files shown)


steipete:
Landed as 5aa64ca9f32cf1d2966ef1a48e2ff0903b2e4668 after resolving the contributor branch against current main and correcting the sender/safety/output boundaries.

Verification:

- `make ci` passed locally, including all Go/Node tests, lint, dead-code checks, 728 generated command pages, docs coverage, and generated agent skills.
- Focused RFC822 regressions cover exact file/stdin bytes, safe content-redacted dry-runs, malformed and duplicate `From` headers, invalid recipients, every incompatible compose flag, global and per-account no-send guards, accountless direct-token rejection, primary-sender zero-settings-call behavior, accepted/rejected verified aliases, Gmail web-URL thread normalization, and the existing `gmail send` JSON output contract.
- The built binary fetched an actual thread ID from the existing designated read-only Gmail account, then used its Gmail web URL in an isolated-home RFC822 dry-run; the result contained only source/byte-count/SHA-256/normalized thread metadata, created no auth files, and preserved the real thread ID.
- The same real-account invocation with `--gmail-no-send` was rejected before sending. No account grant was changed and no email was sent.
- Exact-head hosted CI: https://github.com/openclaw/gogcli/actions/runs/32913903021 (Linux, Windows, macOS, and worker all green).
- Exact-head hosted Docker: https://github.com/openclaw/gogcli/actions/runs/32913903036 (green).

The designated account has only `gmail.readonly`, so a successful live Google send still requires a separately authorized send-capable account and controlled email delivery.

Issue #1022, opened 2026-08-25
gmail attachment / drive download default to the config directory instead of the working directory


ronny-rentner:
When no `--out` is given, downloads are written under the config directory:

- `gmail attachment` → `layout_paths.go:62` `GmailAttachmentsDir()` = `ConfigDir/gmail-attachments`, used as the default at `gmail_attachment.go:101`
- `drive download` (and docs export, photos) → `layout_paths.go:58` `DriveDownloadsDir()` = `ConfigDir/drive-downloads`, used as the default at `drive_download.go:70`, `docs_tab_export.go:174`, `export_via_drive.go:65`, `photos.go:145`, `photos_picker.go:204`

Two reasons this default is wrong:

**1. A downloaded file isn't configuration.** The XDG spec's config dir (`$XDG_CONFIG_HOME`) is for an application's configuration. A file the user asked to download isn't configuration — it's user output. It isn't the app's data, state, or cache either: the XDG base dirs describe the *application's own* files, and a user's download isn't one of them (the only XDG notion of a download location, `XDG_DOWNLOAD_DIR`, is `~/Downloads`, from the separate user-dirs spec). So a downloaded file simply doesn't belong under the config dir.

**2. Every download tool writes to the working directory.** `curl -O`, `curl -o`, and `wget` all default to the current working directory; you pass a path to put it elsewhere. Users expect a downloaded file to appear where they ran the command — gogcli instead defaults it into its own config dir.

**Proposal:** default downloads to the **current working directory**, with `--out` overriding — matching `curl -O` / `wget`, and treating downloads as user output rather than app data. `--out <dir>` keeps working for anyone who wants a fixed location (including today's).

Happy to send the PR if this is the direction you'd want.

steipete:
Closing after maintainer review: the current download default is the documented and tested behavior and matches every released version, so this is working as designed rather than a bug. Use an explicit `--out` path for a different destination. Changing the default would silently move files for existing scripts, so we are keeping it. Thanks for the report.

PR #1036, opened 2026-08-26
fix(download): honor explicit new output directories

steipete:
## Summary

Fix output-directory handling consistently across Drive downloads, Docs/Sheets/Slides exports, Docs tab exports, Google Photos, and Photos Picker without changing default download locations.

- Treat an explicit trailing slash as directory intent even when that directory does not exist yet.
- Preserve the original `--out` value across home-directory expansion so quoted `~/new-directory/` still means a directory.
- Reuse the existing Gmail attachment directory-intent helper at the shared output-file boundary; continue creating private directories only during the real write.
- Preserve existing stdout behavior, provider-filename sanitization, export extensions, `0600` files, `0700` directories, and refusal to overwrite without `--overwrite`.
- Cover all affected command families, adversarial provider filenames, overwrite/no-overwrite behavior, and dry runs that must not create directories.

## Proof

- Every new regression first failed against current `main` across Drive, Docs export, Photos, Photos Picker, and Docs tab destination resolution.
- `make ci` passed, including formatting, lint, all Go packages, JavaScript tests, and all 728 generated command pages.
- Independent pre-commit review reported no actionable findings.
- Live Google Drive validation uploaded one disposable copy of the repository's public license, performed an accountless no-write dry run, downloaded into a previously nonexistent nested directory, compared bytes, verified `0600`/`0700` permissions and overwrite refusal, repeated with `--overwrite`, and permanently deleted only the disposable uploaded file.
- Existing default-directory behavior remains unchanged; this does not close or redefine #1022.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Gmail: send exact prebuilt RFC822 messages with verified sender/account guards, normalized thread IDs, and content-safe offline dry runs. (#1032, #1029) — thanks @higginz777.
 - Gmail: honor attachment-download dry runs for threads and drafts without opening account credentials, fetching messages, or writing files.
+- Downloads: honor explicit new output directories for Drive, Docs/Sheets/Slides exports, Docs tabs, Photos, and Photos Picker without changing existing defaults or overwrite protections.
 - Auth: add least-privilege Gmail `send` and `read-send` authorization, reject read-only/send conflicts, and preserve narrow Gmail grants during reauthorization. (#1033) — thanks @higginz777.
 - Security: redact authenticated Git remotes and configured API keys from backup/config command output and dry-run previews.
 - Backup: honor dry-run for all backup push commands without authenticating, fetching Google data, creating local caches or checkpoints, or changing Git repositories.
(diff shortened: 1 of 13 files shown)


steipete:
Landed in `ffda81fbb234b5b671c05253fc47f32a5c45dece`.

Drive downloads, Docs/Sheets/Slides exports, Docs tab exports, Photos, and Photos Picker now share the same explicit-directory behavior previously used by Gmail attachments. An existing directory or a trailing separator denotes a directory, even when `~/new-directory/` loses that separator during expansion. Destination creation remains exclusively owned by the private output writer, preserving default locations, sanitized filenames, raw stdout, export extensions, overwrite refusal, `0600` files, and `0700` directories.

Proof:

- New regressions failed first on current `main` across all five command families, then passed after the shared owner-boundary fix.
- Full local `make ci` passed, including every Go package, JavaScript checks, docs coverage, and all 728 generated command pages.
- Live Google Drive proof uploaded a disposable copy of the public project license, confirmed the accountless dry run created nothing, downloaded into a new nested directory, compared file bytes, verified private permissions and overwrite refusal, repeated with `--overwrite`, and permanently deleted only the disposable uploaded file.
- The first hosted Windows run exposed a test-only Unix assumption: Go reads `USERPROFILE`, not `HOME`, on Windows. The shared test-home helper now isolates both, and the corrected code also cross-compiles for Windows.
- Corrected hosted Linux/Windows/macOS/worker CI passed: https://github.com/openclaw/gogcli/actions/runs/32918960980
- Corrected hosted Docker passed: https://github.com/openclaw/gogcli/actions/runs/32918960978
- Independent pre-commit and exact-head pre-land reviews reported no actionable findings.

This intentionally does not change existing default download locations or close #1022.

PR #1037, opened 2026-08-26
fix(sheets): reject ambiguous Connected Sheets refresh replies

steipete:
## Summary

Fail safely when a Connected Sheets refresh receives HTTP 200 without the matching refresh execution reply.

- Reject null, empty, missing, nil, or wrong-variant batch-update replies before printing misleading success.
- Preserve successful handling of a real refresh reply with no optional execution statuses.
- Explain that the provider may already have started the refresh and that users should inspect the source before retrying.
- Never automatically repeat a potentially billable refresh request.
- Keep the validation inside the refresh-command owner; delete operations legitimately accept empty replies.

## Proof

- Seven distinct malformed HTTP-200 provider fixtures first reproduced false success on current `main`.
- Every malformed response now returns an error, emits no success JSON, and performs exactly one POST.
- An existing refresh reply with omitted optional statuses still succeeds with an empty JSON statuses array; existing running, failed, dry-run, readonly, permission-guidance, and no-retry coverage remains green.
- Full local `make ci` passed, including formatting, lint, every Go package, JavaScript tests, and all 728 generated command pages.
- Independent pre-commit review reported no actionable findings.
- The malformed-provider behavior is proven with the actual pinned Google client and HTTP fixtures. A successful live BigQuery refresh remains intentionally unclaimed because the test account lacks the required explicit BigQuery grant and an approved billing-enabled execution project; #938 stays open.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Downloads: honor explicit new output directories for Drive, Docs/Sheets/Slides exports, Docs tabs, Photos, and Photos Picker without changing existing defaults or overwrite protections.
 - Auth: add least-privilege Gmail `send` and `read-send` authorization, reject read-only/send conflicts, and preserve narrow Gmail grants during reauthorization. (#1033) — thanks @higginz777.
 - Security: redact authenticated Git remotes and configured API keys from backup/config command output and dry-run previews.
+- Sheets: reject missing Connected Sheets refresh replies without claiming success or retrying a potentially billable execution.
 - Backup: honor dry-run for all backup push commands without authenticating, fetching Google data, creating local caches or checkpoints, or changing Git repositories.
 - Auth: make remote authorization dry runs side-effect-free instead of creating OAuth state or exposing authorization URLs.
 - Config: preserve concurrent account aliases, client mappings, and security settings by keeping every configuration update inside its shared file lock.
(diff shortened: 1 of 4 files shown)


steipete:
Landed in `147464ad00a0bbc0016def3ffe0682a99f67da09`.

Connected Sheets refresh now validates that Google's successful batch-update response actually contains the matching refresh reply before reporting success. Empty, null, and wrong-variant replies are ambiguous because the provider may already have started a billable execution; the command fails with inspection guidance and never retries. A present refresh reply with omitted optional statuses still succeeds and retains the existing empty-array output contract. Validation stays inside the refresh owner because data-source deletion legitimately permits an empty reply.

Proof:

- Seven malformed HTTP-200 response variants first reproduced false success on `main`; every variant now fails without success output after exactly one provider POST.
- A valid refresh reply with no optional statuses remains successful; existing RUNNING/FAILED status behavior, wrapped failure JSON, dry-run, readonly, scope guidance, and no-retry regressions remain green.
- Built CLI proof confirmed an accountless readonly refresh dry run creates no local auth/config state.
- Full local `make ci` passed, including all Go packages, JavaScript checks, generated docs, and independent pre-commit/pre-land reviews.
- Hosted Linux/Windows/macOS/worker CI passed: https://github.com/openclaw/gogcli/actions/runs/32919700082
- Hosted Docker passed: https://github.com/openclaw/gogcli/actions/runs/32919700043
- No real positive BigQuery refresh is claimed: the designated test account lacks the explicit BigQuery grant and an approved billing-enabled execution project. #938 remains open pending that authorized live proof.

PR #1042, opened 2026-08-29
fix(gmail): preserve direct access token in watch serve

bill-starfoundry:
## What Problem This Solves

`gog gmail watch serve` recognizes `--access-token` and
`GOG_ACCESS_TOKEN` when it starts, but loses that token while processing
Pub/Sub notifications.

When no stored OAuth credential metadata exists, each notification fails with:

```text
watch: handle push failed: gmail options: read credentials:
read credentials metadata: oauth credentials missing
```

The push endpoint consequently returns HTTP 500 even though the command was
started with a valid direct access token.

## Root Cause

The direct access token is stored in the command context. Each HTTP request
arrives with its own request context, which does not contain that value.

The watch server already captures and restores the selected OAuth client before
constructing a Gmail service for a request, but it does not do the same for the
direct access token. Gmail client construction therefore falls back to the
stored-credential path.

## Change

Capture the direct access token when `watch serve` starts and reattach it to
each request context before constructing the Gmail service.

Only the token value is copied. The HTTP request context remains the parent, so
request cancellation and deadlines are preserved.

This does not change:

- Stored OAuth credential behavior.
- Gmail watch authorization or filtering.
- Direct-token expiration or refresh behavior.
- CLI flags, output formats, or persisted state.

## User Impact

`gog gmail watch serve` can process notifications using its documented direct
access-token authentication mode without requiring stored OAuth client
credentials or refresh tokens.

The token remains static and expires normally; this change does not add
automatic refresh.

## Regression Coverage

Added
`TestGmailWatchServeCmd_PreservesDirectAccessTokenForRequestContexts`
beside the existing client-override request-context regression.

The test starts the command with a direct token in its command context, invokes
the constructed Gmail factory with a fresh request context, and verifies the
factory receives the original token. It fails on unmodified `main` with
`expected direct access token, got ""`.

## Real Behavior Proof

Tested on Linux amd64 with a real Gmail account, a broker-issued direct access
token, and a Gmail Pub/Sub notification. Identifiers and credentials are
redacted.

Before the change:

```text
Note: Using direct access token (expires in ~1 hour; no auto-refresh)
watch: listening on 127.0.0.1:<port>/
watch: handle push failed: gmail options: read credentials:
read credentials metadata: oauth credentials missing
HTTP/1.1 500 Internal Server Error
```

After the change, using a binary built from this PR:

```text
Note: Using direct access token (expires in ~1 hour; no auto-refresh)
watch: listening on 127.0.0.1:<port>/
DEBUG using direct access token serviceLabel=gmail
HTTP/1.1 200 OK
```

The patched process fetched the pending Gmail history successfully. No OAuth
client metadata or refresh token was available to the process.

## Testing

- `go test ./internal/cmd -run '^TestGmailWatchServeCmd_Preserves(DirectAccessToken|ClientOverride)ForRequestContexts$' -count=1`
- `make ci`
- Live Gmail Pub/Sub push verification described above.

## What Was Not Tested

- Allowing one direct token to expire inside a continuously running server.
  Direct tokens remain static and non-refreshing as documented.
- Stored-refresh-token authentication, which is unchanged.
- Pub/Sub OIDC verification, which is independent of Gmail API authentication.

## Maintainer follow-up

Kept the four-line production repair and strengthened the regression to use the real Gmail client against a local HTTP endpoint. It now verifies the outgoing bearer header as well as the incoming request's deadline and cancellation. Added watch documentation that distinguishes Gmail API authentication from push-sender authentication and makes static-token expiration explicit.

Independent before/after verification used two fresh, isolated gog homes with no stored OAuth client metadata. The only credential supplied to each watcher was a short-lived direct token, which authenticated real Gmail profile/history reads; the incoming push was synthetic and delivered only over loopback. Unchanged main returned HTTP 500 with the missing-credentials error. The repaired binary returned HTTP 202 for empty history and persisted the unique push ID and update timestamp, with the real history cursor unchanged and no auth-recovery state. Those state assertions rule out a stale/duplicate-push acknowledgment being mistaken for a successful authenticated request.

This verification did not send or modify mail, register a Gmail watch, or change Pub/Sub/IAM configuration. It was not an end-to-end Pub/Sub delivery test and did not wait for token expiration. The contributor's separate live Pub/Sub proof above is preserved as contributor evidence.

Verification:

- The direct-token regression fails against the original main service factory and passes with this repair.
- All watch-serve tests pass; the focused direct-token regression also passes with the race detector.
- Full local `make ci` passes on the patch applied to current main.
- Independent Codex autoreview found no findings in its requested P0 scope; the request-context and transport paths were also reviewed directly.

Thanks @bill-starfoundry for the fix and original live reproduction. Contributor credit is retained.

Exact-head GitHub verification also passed: [Linux, macOS, Windows, and worker CI](https://github.com/openclaw/gogcli/actions/runs/33607492291), plus the [Docker build](https://github.com/openclaw/gogcli/actions/runs/33607492284). Both fork workflows initially required maintainer approval; after approval, their first executed attempts passed. No CI repair or test rerun was needed.

### DIFF
--- a/docs/watch.md
+++ b/docs/watch.md
@@ -181,6 +181,15 @@ Schema (v1):
 - `--max-bytes`: hard cap on body bytes (default `20000`).
 - If over cap: truncate + set `bodyTruncated=true`.
 
+## Gmail API authentication
+
+`gog gmail watch serve` preserves the selected OAuth client and any direct
+`--access-token` (or `GOG_ACCESS_TOKEN`) when handling incoming push requests.
+Direct-token mode does not require stored OAuth client credentials or a refresh
+token. The token remains static and expires normally; restart the server with a
+fresh token when needed. This is separate from authenticating the push sender
+below.
+
 ## Auth (push)
 
 Preferred:
--- a/internal/cmd/gmail_watch_cmds.go
+++ b/internal/cmd/gmail_watch_cmds.go
@@ -401,11 +401,15 @@ func (c *GmailWatchServeCmd) Run(ctx context.Context, kctx *kong.Context, flags 
 	}
 
 	selectedClient := strings.TrimSpace(flags.Client)
+	directToken := authclient.AccessTokenFromContext(ctx)
 	gmailFactory, err := gmailServiceFactory(ctx)
 	if err != nil {
 		return err
 	}
 	serviceFactory := func(ctx context.Context, account string) (*gmail.Service, error) {
+		if directToken != "" {
+			ctx = authclient.WithAccessToken(ctx, directToken)
+		}
 		if selectedClient != "" {
 			ctx = authclient.WithClient(ctx, selectedClient)
 		}
--- a/internal/cmd/gmail_watch_serve_test.go
+++ b/internal/cmd/gmail_watch_serve_test.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"io"
 	"net/http"
+	"net/http/httptest"
 	"os"
 	"path/filepath"
 	"testing"
@@ -16,6 +17,7 @@ import (
 	"google.golang.org/api/idtoken"
 
 	"github.com/openclaw/gogcli/internal/authclient"
+	"github.com/openclaw/gogcli/internal/googleapi"
 )
 
 func TestGmailWatchServeCmd_DryRunDoesNotTouchStateOrListen(t *testing.T) {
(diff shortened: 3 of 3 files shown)


steipete:
Merged as [f78486ada64e80d5d09431493d513777c3a6271a](https://github.com/openclaw/gogcli/commit/f78486ada64e80d5d09431493d513777c3a6271a). Thanks @bill-starfoundry for the focused fix and original reproduction; contributor credit is preserved.

Kept the four-line production repair. The follow-up uses the real Gmail transport in the regression, checks its bearer header and request deadline/cancellation, and documents static-token expiry separately from push-sender authentication.

Verified head: `4345993bd53c4d9eec96011f64a5ba39489847bc`.

- **Red:** overlaid only `internal/cmd/gmail_watch_cmds.go` from main `874fd09b36a9fc8ab94e7fe054bd1d8cc2070964`; the direct-token regression fails with `expected direct access token, got ""`.
- **Green:** `go test ./internal/cmd -run '^TestGmailWatchServeCmd_' -count=1 -timeout=60s -v` and `go test -race ./internal/cmd -run '^TestGmailWatchServeCmd_PreservesDirectAccessTokenForRequestContexts$' -count=1 -timeout=90s`.
- **Full gate:** `make ci` with the repository's pinned tools, applied onto current main; independent Codex autoreview was clean in its requested P0 scope.
- **GitHub:** [Linux, macOS, Windows, and worker CI](https://github.com/openclaw/gogcli/actions/runs/33607492291) and the [Docker build](https://github.com/openclaw/gogcli/actions/runs/33607492284) passed after fork-workflow approval. No CI repair or rerun was needed.
- **Independent live proof:** fresh isolated homes, a short-lived direct token, a real Gmail history read, and a synthetic loopback push. Original main returned 500; the fix returned 202 for empty history, persisted the unique push marker and timestamp, retained the real cursor, and did not enter auth recovery. Neither home contained OAuth client metadata. No mail, watch registration, or Pub/Sub/IAM settings were changed. This was not an end-to-end Pub/Sub delivery or token-expiration test.

The PR body retains the contributor's separate live Pub/Sub evidence and records the independent verification above. Main has been pulled with `--ff-only` and is clean.

PR #1004, opened 2026-08-18
fix(paging): stop calendar and Gmail listing on a repeated page token

SebTardif:
## What Problem This Solves

`gog calendar events --all` (and any other path that lists every calendar) walks `CalendarList.List` with no memory of page tokens it has already seen. Gmail backup ID listing does the same on `users.messages.list`. If Google repeats `nextPageToken`, both loops never finish.

## Why

The repo already has this guard. `collectAllPages` errors with `pagination loop: repeated page token`. Photos Picker and Drive changes do the same with a local seen-token map. `listCalendarList` is in package `cmd`, so it now calls `collectAllPages` instead of growing a second helper. Gmail backup lives under `internal/backup/gmail` and cannot import `cmd`, so it adds the same local `seenTokens` map Photos Picker already uses.

## User Impact

A stuck Google page token now fails with `repeated page token` instead of hanging `gog calendar events --all` or a Gmail backup list until the process is killed.

## Evidence

terminal output from the unpatched loops versus this patch, using a Calendar `httptest` server and a Gmail source that always return `nextPageToken=stuck`.

Unpatched (300ms context; the HTTP peer answers immediately, so the deadline is the only stop):

```console
$ go test ./internal/cmd/ -count=1 -timeout 5s -v -run TestListCalendarListRejectsRepeatedPageToken
=== RUN   TestListCalendarListRejectsRepeatedPageToken
    calendar_list_test.go:115: err = context deadline exceeded
--- FAIL: TestListCalendarListRejectsRepeatedPageToken (0.30s)
FAIL

$ go test ./internal/backup/gmail/ -count=1 -timeout 5s -v -run TestListMessageIDsRejectsRepeatedPageToken
=== RUN   TestListMessageIDsRejectsRepeatedPageToken
    fetch_test.go:146: err = list Gmail backup messages: context deadline exceeded
--- FAIL: TestListMessageIDsRejectsRepeatedPageToken (0.30s)
FAIL
```

Patched (same commands, same stuck token, returns immediately):

```console
$ go test ./internal/cmd/ -count=1 -timeout 10s -v -run TestListCalendarListRejectsRepeatedPageToken
=== RUN   TestListCalendarListRejectsRepeatedPageToken
    calendar_list_test.go:117: err = pagination loop: repeated page token "stuck"
--- PASS: TestListCalendarListRejectsRepeatedPageToken (0.00s)
PASS

[...]

PR #1044, opened 2026-08-29
fix(people): reject repeated page tokens in raw email resolve

SebTardif:
## What Problem This Solves

`gog people raw user@host` and `gog contacts raw user@host` resolve an email by paging `People.Connections.List` with `pageToken = connections.NextPageToken` and no seen-set. If Google repeats `nextPageToken`, that loop never finishes and the CLI hangs until the process is killed.

## Why

The repo already has this guard. `collectAllPages` errors with `pagination loop: repeated page token`. Calendar listing uses it after [#1004](https://github.com/openclaw/gogcli/pull/1004). Email resolve lives in the same `cmd` package, so it now calls that helper instead of growing a second loop.

## User Impact

A stuck People connections page token now fails with `repeated page token` instead of hanging `gog people raw` / `gog contacts raw` on an email identifier.

## Evidence

terminal output from the unpatched email-resolve loop versus this patch. A People connections `httptest` server always returns `nextPageToken=stuck`. `gog people raw ada@example.com` (via `PeopleRawCmd`) pages that list, then would call `People.Get`.

Unpatched (the HTTP peer answers immediately; the hang guard stops the loop after 8 list calls):

```console
$ go test ./internal/cmd -run TestPeopleRaw_EmailResolveRejectsRepeatedPageToken -count=1 -timeout 30s -v
=== RUN   TestPeopleRaw_EmailResolveRejectsRepeatedPageToken
    people_raw_test.go:185: repeated page token looped: 9 list calls
    people_raw_test.go:185: repeated page token looped: 10 list calls
    people_raw_test.go:204: err = Get "http://127.0.0.1:62467/v1/people/me/connections?alt=json&pageSize=1000&pageToken=stuck&personFields=names%2CemailAddresses%2Cmetadata&prettyPrint=false": EOF after 10 list calls
--- FAIL: TestPeopleRaw_EmailResolveRejectsRepeatedPageToken (0.00s)
FAIL
```

Patched (same command, same stuck token, returns immediately):

```console
$ go test ./internal/cmd -run TestPeopleRaw_EmailResolveRejectsRepeatedPageToken -count=1 -timeout 15s -v
=== RUN   TestPeopleRaw_EmailResolveRejectsRepeatedPageToken
    people_raw_test.go:206: err = pagination loop: repeated page token "stuck" after 2 list calls
--- PASS: TestPeopleRaw_EmailResolveRejectsRepeatedPageToken (0.00s)
PASS
ok  	github.com/openclaw/gogcli/internal/cmd	0.948s
```

Live CLI still accepts an email as the raw identifier:

```console
$ /tmp/gog-f008 people raw --help
Usage: gog people (person) raw <userId> [flags]
Dump raw People API response as JSON (People.Get; lossless; for scripting and
LLM consumption)
Arguments:
  <userId>    Person resource name (people/...) or email
```

`go test ./internal/cmd -count=1` also completed on this tree (ok, 62.568s). `gofmt` is clean. `golangci-lint run ./internal/cmd/ --new-from-rev=upstream/main` reported 0 issues.

## Real behavior proof

- **Behavior or issue addressed:** Repeated Google `nextPageToken` values could hang `gog people raw user@host` / `gog contacts raw user@host` while resolving the email through `People.Connections.List`.
- **Real environment tested:** macOS Darwin 25.6.0 arm64, Go go1.27.0, full clone at `/tmp/oc-pr-gogcli-F008` on `fix/f008-people-raw-page-token` from `upstream/main` at `be4037f7`.
- **Exact steps or command run after this patch:** From `/tmp/oc-pr-gogcli-F008`, ran `go test ./internal/cmd -run TestPeopleRaw_EmailResolveRejectsRepeatedPageToken -count=1 -timeout 15s -v` against a People connections `httptest` that always returns `nextPageToken=stuck`, then ran `/tmp/gog-f008 people raw --help` on the binary built from this branch.
- **Evidence after fix:** terminal output above. Email resolve now returns `pagination loop: repeated page token "stuck"` after 2 list calls (0.00s). The unpatched loop made 9+ list calls and only stopped when the hang guard closed the server.
- **Observed result after fix:** The stuck token no longer pages forever. `PeopleRawCmd` returns the repeated-token error on the next page instead of hanging.
- **What was not tested:** A live `people.connections.list` response that actually repeats a token. That requires a faulty Google page, which we cannot force from a healthy account.

## Related

- Same-repo guard already used by `collectAllPages` in `internal/cmd/paging.go` and by calendar listing in [#1004](https://github.com/openclaw/gogcli/pull/1004).
- Unguarded People raw email paging dates to [`e16241bc`](https://github.com/openclaw/gogcli/commit/e16241bc9a1b6348c7898c63b450c5e321cbf54e) (`fix(contacts): resolve raw email identifiers`, 2026-06-01, 89 days ago).

## Maintainer follow-up

The resolver now filters and deduplicates matching resource names inside each page fetch, then passes only those identifiers to the shared pagination collector. This preserves the original lookup's memory behavior while adding the repeated-token guard.

The command-level regression now uses an HTTP 400 sentinel on a third request instead of failing inside the HTTP handler. With only the production resolver replaced by the unpatched implementation from main, the regression fails after three requests; with this fix it returns the repeated-token error after exactly two requests. Additional cases cover matches on later pages, duplicate records for one resource, distinct-resource ambiguity, and a later-page API failure without partial output. The raw API documentation records these semantics.

Focused proof: `go test ./internal/cmd -run '^(TestPeopleRaw|TestContactsRaw)' -count=1 -timeout=45s -v` passes. This is injected-fault command-path proof using the real generated People client, not a claim that a live Google server returned a malformed page token.

### DIFF
--- a/docs/raw-api.md
+++ b/docs/raw-api.md
@@ -40,6 +40,12 @@ legacy top-level `Document` fields such as `body`, `lists`, and
 populates its recursive `tabs` tree. Without either flag, the command keeps the
 existing first-tab response.
 
+`gog people raw` and `gog contacts raw` also accept email identifiers. They
+scan contact pages before selecting a unique resource: repeated records for
+the same resource count once, while distinct contacts sharing an email are
+ambiguous. If Google repeats a page token, lookup stops with a pagination
+error instead of paging indefinitely.
+
 Use service-native field masks when available:
 
 ```bash
--- a/internal/cmd/people_raw.go
+++ b/internal/cmd/people_raw.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"fmt"
 	"strings"
-
-	"google.golang.org/api/people/v1"
 )
 
 // defaultPeopleRawMask is the field mask used when the user does not
@@ -65,7 +63,8 @@ func runPeopleRaw(ctx context.Context, flags *RootFlags, id, fields string, pret
 
 	resource := normalizePeopleResource(identifier)
 	if strings.Contains(identifier, "@") && !strings.HasPrefix(identifier, "people/") {
-		fetch := func(pageToken string) ([]*people.Person, string, error) {
+		seen := make(map[string]bool)
+		fetch := func(pageToken string) ([]string, string, error) {
 			call := svc.People.Connections.List(peopleMeResource).
 				PersonFields("names,emailAddresses,metadata").
 				PageSize(1000).
@@ -77,20 +76,19 @@ func runPeopleRaw(ctx context.Context, flags *RootFlags, id, fields string, pret
 			if listErr != nil {
 				return nil, "", wrapPeopleAPIError(listErr)
 			}
-			return connections.Connections, connections.NextPageToken, nil
+			var matches []string
+			for _, person := range connections.Connections {
+				if person == nil || !personHasEmail(person, identifier) || person.ResourceName == "" || seen[person.ResourceName] {
+					continue
+				}
+				seen[person.ResourceName] = true
+				matches = append(matches, person.ResourceName)
+			}
+			return matches, connections.NextPageToken, nil
 		}
-		connections, listErr := collectAllPages("", fetch)
+		matches, listErr := collectAllPages("", fetch)
 		if listErr != nil {
 			return listErr
-		}
-		matches := make([]string, 0, 1)
-		seen := make(map[string]bool)
-		for _, person := range connections {
-			if person == nil || !personHasEmail(person, identifier) || person.ResourceName == "" || seen[person.ResourceName] {
-				continue
-			}
-			seen[person.ResourceName] = true
-			matches = append(matches, person.ResourceName)
 		}
 		switch len(matches) {
 		case 0:
--- a/internal/cmd/people_raw.go
+++ b/internal/cmd/people_raw.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"strings"
+
+	"google.golang.org/api/people/v1"
 )
 
 // defaultPeopleRawMask is the field mask used when the user does not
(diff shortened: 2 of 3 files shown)


steipete:
Landed as e1f678fa0a262a437b2d80587f901a6eca31b11e. Thanks @SebTardif for the focused pagination repair; contributor credit is preserved in the squash commit.

The shared guard now stops repeated tokens without retaining unrelated People records. Matching resource IDs are filtered and deduplicated per page, and lookup still scans all pages before selecting a unique contact.

Verification on the complete PR applied to current main:

- `go test ./internal/cmd -run '^(TestPeopleRaw|TestContactsRaw)' -count=1 -timeout=45s -v` — passed.
- `go test -race ./internal/cmd -run '^(TestPeopleRaw|TestContactsRaw)' -count=1 -timeout=60s` — passed.
- `make ci TOOLS_DIR=/Users/steipete/Projects/gogcli/.tools` — passed: formatting, lint, deadcode, Go/Node tests, generated documentation, and agent-skill checks.
- Codex Auto Review found no accepted/actionable findings in its requested scope; maintainer source review also completed.
- [GitHub CI 33601097515](https://github.com/openclaw/gogcli/actions/runs/33601097515) passed Linux, Windows, macOS, and worker checks at fead06248aeb570d0059fc9ab71fe8e154b0921a; the Docker image check also passed.

The bounded regression fails after three requests when only the unpatched production resolver from main is overlaid, then passes after exactly two requests with the fix. Additional cases cover later-page matches, duplicate resources, ambiguity, and later API errors with no partial stdout. This is injected-fault proof through the command and generated People client, not a claim that a live Google server returned a malformed token.

Issue #640, opened 2026-05-22
docs: --page-size=A4|Letter|... named-size shortcut for write/page-layout


sebsnyk:
## Summary

v0.19 added `--page-width` / `--page-height` on `gog docs write` and `gog docs page-layout` (closes #629). Setting A4 today requires the explicit numbers:

```
gog docs write <id> --page-width=595.275pt --page-height=841.89pt --margin-left=36pt --margin-right=36pt --margin-top=36pt --margin-bottom=36pt
```

Would be more ergonomic to accept named sizes:

```
gog docs write <id> --page-size=A4
gog docs page-layout <id> --page-size=Letter
```

## Suggested set

Common print sizes the CLI could map to PT:

| Name   | Width (pt) | Height (pt) | Notes                          |
|--------|------------|-------------|--------------------------------|
| A4     | 595.275    | 841.890     | ISO 216                        |
| A5     | 419.528    | 595.275     | ISO 216                        |
| Letter | 612        | 792         | US 8.5×11 in                   |
| Legal  | 612        | 1008        | US 8.5×14 in                   |
| Tabloid| 792        | 1224        | US 11×17 in                    |

Behaviour: `--page-size=<name>` sets both width and height in one go; mutually exclusive with `--page-width` / `--page-height` (or, if both passed, explicit dims override the named preset).

## Why

- Removes the magic-number footgun (`841.890` is easy to mistype)
- Matches how customers actually think about page layouts ("A4" vs "841.89pt")
- Trivial to add — internal lookup table, no API change needed

## Workaround

Pass explicit `--page-width=595.275pt --page-height=841.89pt` until shipped.

PR #1050, opened 2026-08-31
feat(docs): add A3 to --page-size presets

ryo-touch:
## Summary

`--page-size` (added in [#640](https://github.com/openclaw/gogcli/issues/640)) covers A4 and A5 but not A3, so the next size up in the same ISO 216 series still needs the magic numbers the flag exists to remove:

```bash
gog docs page-layout <id> --page-width=841.890pt --page-height=1190.551pt
```

The gap looks incidental rather than deliberate — the suggested set in #640 simply didn't list A3.

Only A3 is added, not the rest of the A series. A3 is the common large-format sheet (and the one I hit in practice); A2 and up are rare enough for Docs that the explicit-dimension form seems the better default for them. Happy to widen the set if you'd rather have it complete.

## Changes

- Add `a3` (841.890 × 1190.551 pt = 297 × 420 mm) to `docsPageSizePresets`
- Update the two `A4, A5, Letter, Legal, Tabloid` strings (flag help + the `resolveDocsPageSize` usage error)
- Regenerate `docs/commands/gog-docs-page-layout.md` and `gog-docs-write.md` via `make docs-commands`
- Show `--page-size` in the `docs/docs-editing.md` Page Layout guide, which until now only demonstrated the explicit-dimension form and so never surfaced the flag to anyone reading the prose docs

A3's width is A4's height, and the existing `a4` entry already spells that 841.890, so the table stays internally consistent.

## Testing

`make ci` passes.

- New `TestResolveDocsPageSize` — the resolver had no direct table test. Covers A3 (`A3` / `a3` / padded), pass-through when `--page-size` is empty, an unknown name, and the "cannot be combined with `--page-width` / `--page-height`" guard.
- New `TestDocsPageLayoutCmd_NamedPageSizeA3` — asserts the emitted `updateDocumentStyle` carries 841.890 × 1190.551 PT with `fields = "pageSize.width,pageSize.height"` and leaves `documentFormat` untouched (i.e. `--page-size` alone still preserves pageless/pages mode).

  Note on overlap: `TestDocsPageSizePresetBuildsDimensions` and `TestDocsPageSizeRejectsExplicitDimensions` already cover the preset→dimensions and conflict paths, but both call `buildUpdateDocumentStyleRequest` directly. Nothing exercised `--page-size` through Kong argument parsing, so this is the first test that does. Glad to drop it if you consider that layer out of scope.
- Manual check against a live doc: `--page-size A4` then `--page-size A3` round-trip to 595.2755905511812 × 841.8897637795277 and 841.8897637795277 × 1190.5511811023623 PT respectively.

## User-facing

`--page-size A3` is now accepted by `gog docs write` and `gog docs page-layout`. No behaviour change for existing values.

## Noticed while here

`dryRunExit` ran before `resolveDocsPageSize`, so `--dry-run --page-size=A2` exited 0 and reported `pageSize A2` while the real run failed with a usage error. Pre-existing and orthogonal to this change, so it was filed separately as [#1051](https://github.com/openclaw/gogcli/issues/1051) rather than widening this PR — and it is now fixed by [#1052](https://github.com/openclaw/gogcli/pull/1052), which this branch has been rebased onto.

## Maintainer integration and proof

Integrated current main and added a `docs write --page-size a3` regression, explicit paged examples, and a credited Unreleased entry. Named sizes alone preserve the current page mode.

Verified candidate `27b067bf64643722468122d382769f175aed2103` with `GOTOOLCHAIN=go1.27.1 make ci`, focused Docs tests, and independent Codex review (no actionable P0/P1 findings). Live Google Docs proof verified A3 dimensions, unchanged state after dry-run, pages and pageless preservation, the write path, and an explicit switch to pages. The synthetic fixture was trashed and its cleanup read back; temporary credentials were removed. No new Google grants or email sends.

The live-tested binary SHA-256 was `a4c0383001e31b30e801521a122eb17d511ff49a374871146bedbe5acd989031`; the reviewed/tested diff was verified unchanged before committing.


steipete:
Landed in 62a7a208a7ca9b8c90ef3b10755a8e409f97bde4. Thanks @ryo-touch!

The final PR includes the A3 preset, mode-preservation coverage for both `docs page-layout` and `docs write`, clearer usage docs, and a credited Unreleased entry. Full local `make ci`, independent review, and GitHub CI/Docker passed. Independent live Google Docs checks verified the dry-run, dimensions, paged/pageless preservation, and explicit mode changes; the synthetic document was trashed and cleanup verified. No new grants or mail sends.

CI: https://github.com/openclaw/gogcli/actions/runs/33716954837

PR #1052, opened 2026-08-31
fix(docs): validate page layout before dry-run

### LINKED ISSUE #1051: docs page-layout: --dry-run reports success for an invalid --page-size
## Summary

`gog docs page-layout` validates `--page-size` only on the real execution path, so `--dry-run` reports success for a name the command would reject:

```console
$ gog docs page-layout <docId> --page-size A2 --dry-run --json
{
  "dry_run": true,
  "op": "docs.page-layout",
  "request": {
    "documentId": "<docId>",
    "pageSize": "A2"
  }
}
$ echo $?
0

$ gog docs page-layout <docId> --page-size A2
set page layout: --page-size must be one of A3, A4, A5, Letter, Legal, Tabloid
$ echo $?
2
```

`gog docs write` shares the same `DocsLayoutFlags` but is **not** affected — it already validates first:

```console
$ gog docs write <docId> --text x --page-size A2 --dry-run --json
--page-size must be one of A3, A4, A5, Letter, Legal, Tabloid
$ echo $?
2
```

## Cause

`DocsWriteCmd` calls `validateDocumentStyle()` (`internal/cmd/docs_edit.go:79`) before its `dryRunExit` (`:148`). That helper (`:96`) runs `buildUpdateDocumentStyleRequest` purely for its error and discards the request, which resolves the named size and surfaces an invalid name early.

`DocsPageLayoutCmd` has no equivalent step: `dryRunExit` (`internal/cmd/docs_set_page_layout.go:63`) builds its payload straight from the raw flag values, and `resolveDocsPageSize` only runs afterwards via `setDocumentStyle` → `buildUpdateDocumentStyleRequest` (`:80`), which the dry-run path exits before reaching.

So the fix is narrow, and there is a working precedent to copy inside the same package.

## Why it matters

Dry-run is the documented way to check a write before making it, and here it predicts the wrong outcome. It also echoes back the unresolved name (`"pageSize": "A2"`) rather than the dimensions that would actually be sent, so the preview does not show what the batchUpdate would contain even when the name is valid.

## Proposal

Give `DocsPageLayoutCmd` the same pre-`dryRunExit` validation `DocsWriteCmd` already has, and put the resolved dimensions in the dry-run payload.

Acceptance criteria:

- `gog docs page-layout <docId> --page-size <invalid name> --dry-run` exits non-zero with the same usage error as the real run
- `gog docs page-layout <docId> --page-size A3 --dry-run` payload contains the resolved `841.890pt` × `1190.551pt` rather than the bare name
- `gog docs page-layout <docId> --page-size A3 --page-width 100 --dry-run` errors on the mutual-exclusion guard
- `gog docs write` behaviour is unchanged

## Context

Noticed while adding the A3 preset in #1050. Pre-existing and independent of that change, so filing separately.

Correction: an earlier revision of this issue claimed `docs write` had the same bug. It does not — verified above. Only `docs page-layout` needs the fix. Thanks to ClawSweeper for catching it.

steipete:
`docs page-layout --dry-run` reported success for unknown page sizes, conflicting size flags, and invalid dimensions because it exited before building the document-style request.

Build and validate that request before dry-run, expose its resolved dimensions and field mask as `request.updateDocumentStyle`, and execute that same request after resolving the optional tab. Existing preview fields remain available. Add command-level regressions and document the preview. Thanks @ryo-touch for the diagnosis.

Closes #1051.

Validation:

- `go test ./internal/cmd -run 'TestDocsPageLayoutCmd|TestDocsPageSize|TestDocsWrite' -count=1` — passed.
- `golangci-lint run ./internal/cmd/ --new-from-rev HEAD` — 0 issues (run before committing).
- Codex autoreview — scoped clean at its default P0 threshold.
- Built CLI, isolated temporary `GOG_HOME`, no credentials or Google requests:

```console
$ gog docs page-layout doc1 --page-size=A2 --dry-run --json
set page layout: --page-size must be one of A4, A5, Letter, Legal, Tabloid
exit 2
$ gog docs page-layout doc1 --page-size=A4 --page-width=100 --dry-run --json
set page layout: --page-size cannot be combined with --page-width or --page-height
exit 2
$ gog docs page-layout doc1 --page-size=A4 --margin-left=0 --dry-run --json
exit 0
```

The valid preview includes width `595.275 PT`, height `841.89 PT`, zero left margin, and field mask `pageSize.width,pageSize.height,marginLeft`. Both invalid page-layout commands exited 0 on main before this fix. `docs write --text=x --page-size=A2 --dry-run --json` still exits 2.

This fix uses the presets currently on main and does not depend on the separate A3 addition in #1050.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.38.2 - Unreleased
 
+- Docs: validate page-layout sizes and margins before dry-run, and preview the resolved document-style request used for execution. (#1051) — thanks @ryo-touch.
 - Sheets: preserve zero-valued sheet, row, and column indexes in Connected Sheets refresh status references, keeping A1 anchors identifiable. (#938) — thanks @ryo-touch.
 - Dependencies: prefer Go 1.27 while retaining Go 1.26 compatibility, refresh Google and MCP SDKs, update tracking worker dependencies and Go tooling, and refresh Docker build actions.
 
--- a/docs/docs-editing.md
+++ b/docs/docs-editing.md
@@ -366,6 +366,12 @@ flags are supplied; pass `--layout` when you also want to toggle pageless/pages.
 `--pageless` preserves Google Docs' existing width unless `--page-width` is set
 explicitly.
 
+`docs page-layout --dry-run --json` validates page sizes, dimensions, and
+margins before accessing Google. Its `request.updateDocumentStyle` preview
+includes the resolved point dimensions and field mask used for execution;
+the original flag values remain available alongside it. A requested tab is
+shown by name and resolved only when the command executes.
+
 Command page:
 
 - [`gog docs page-layout`](commands/gog-docs-page-layout.md)
--- a/internal/cmd/docs_set_page_layout.go
+++ b/internal/cmd/docs_set_page_layout.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/alecthomas/kong"
+	"google.golang.org/api/docs/v1"
 
 	"github.com/openclaw/gogcli/internal/outfmt"
 	"github.com/openclaw/gogcli/internal/ui"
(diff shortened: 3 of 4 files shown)


steipete:
Landed in https://github.com/openclaw/gogcli/commit/8f56f0afca9b0b715f34d97ae9e1baf67d7d5796. Invalid page-layout previews now fail before Google access, and valid previews include the resolved request. Full local `make ci` passed; [CI on the reviewed head](https://github.com/openclaw/gogcli/actions/runs/33373142553) is green. Thanks @ryo-touch for the diagnosis.

PR #1065, opened 2026-09-02
fix(drive): reject repeated page tokens in sync push listing

SebTardif:
## Summary

`gog drive sync push` previously copied each `nextPageToken` into the next folder-list request without detecting repeated tokens. A faulty continuation could keep the recursive preflight running indefinitely and repeatedly accumulate the same children.

Keep @SebTardif's narrow fix: route `listDriveSyncChildren` through the existing `collectAllPages` helper, preserving the query, selected fields, shared-drive flags, request context, ordering, and incomplete-search rejection. Repeated tokens and the shared page limit now fail preflight. Ordinary provider failures still discard partial children; remote writes begin only after the entire recursive plan succeeds.

The maintainer follow-up strengthens the existing regression test, adds populated multi-page success and later-page failure cases, and exercises the real command path with a nested pagination failure after an earlier upload could already have been planned. The command must make zero writes and emit no success output. `docs/examples.md` documents that boundary.

## Validation

Tested the complete candidate integrated on main at `eaea84227d4f0d8b7ebb5beccaca80ede0876677`.

- The maintained repeated-token test fails against the old production adapter after a bounded third request; the corrected adapter rejects the repeated token after two requests and returns no partial children.
- Populated two-page results preserve order. A second-page API error or incomplete-search response returns no partial listing.
- The full `drive sync push` regression, without `--dry-run`, verifies no HTTP writes and no success output on a nested pagination error.
- Focused tests and the full `make ci` gate passed. One initial test-lint finding was corrected by marking the independent table subtests parallel; no lint rule was disabled.
- Independent Codex review of the complete integrated candidate was scoped-clean at P0/P1.

```sh
go test ./internal/cmd -run '^(TestDriveSyncPush|TestGoogleDriveSyncClientUsesSharedDriveFlags|TestListDriveSyncChildren|TestExecuteDriveSyncPush)' -count=1 -timeout=90s
go test -race ./internal/cmd -run '^(TestDriveSyncPush|TestGoogleDriveSyncClientUsesSharedDriveFlags|TestListDriveSyncChildren|TestExecuteDriveSyncPush)' -parallel=1 -count=1 -timeout=120s
make ci
```

The serialized race run passed. The default-parallel run exposed a separate preexisting global locked-flag-state race; the same failure was reproduced on unchanged main using only the existing dry-run/help tests. That ownership defect is being handled separately, not hidden by deleting concurrent tests or attributed to this pagination change.

[...]

PR #1063, opened 2026-09-02
fix(backup): reject repeated Chat and Classroom page tokens

SebTardif:
## Summary

Use the existing shared paginator for Chat spaces/messages and Classroom courses, topics, announcements, coursework, materials, and submissions. Repeated or cyclic continuation tokens now stop collection instead of repeatedly fetching the same pages; the shared 10,000-page limit also applies.

Preserve the existing Classroom child-list best-effort contract: an ordinary API failure ends that child listing without discarding earlier pages. The maintainer follow-up handles that policy inside the fetch callback, leaves the shared paginator strict, and removes error-string classification. Chat and top-level course listing errors remain strict.

## User-visible behavior

Pagination guard errors stop the affected service collection. Default `--best-effort` records them in an encrypted service error shard and lets other services continue; `--no-best-effort` returns the error. No flags, scopes, list filters, or provider permissions changed. The handwritten backup guide documents the boundary.

## Verification

- Maintained Chat-spaces and Classroom-courses tests against the old main implementation made an unwanted third request before the synthetic peer's safety cap returned HTTP 400. The guarded implementation stops after two requests with the pagination error.
- Added tests failed on contributor head `c0238cce`: a page-two 403 erased earlier rows from all five Classroom child collections; provider error text containing `pagination loop` was also incorrectly treated as a guard error. Both cases pass after the follow-up.
- Restored generated-client coverage for Chat messages and all five Classroom child types, including propagation through the aggregate child collector.
- Shared helper coverage checks ordinary fetch errors, multi-token cycles, and the page cap without making 10,000 HTTP requests.
- `go test ./internal/cmd -run '^(TestFetchBackup(Chat|Classroom)|TestCollectAllPages)' -count=1 -timeout=90s -v` passed.
- `go test -race ./internal/cmd -run '^(TestFetchBackup(Chat|Classroom)|TestCollectAllPages)' -count=1 -timeout=120s` passed.
- `make ci` and `git diff --check` passed with Go 1.27.0 on macOS arm64.

Codex autoreview of the complete integrated candidate was scoped-clean with no actionable P0/P1 findings. Exact-head [CI](https://github.com/openclaw/gogcli/actions/runs/33648974807) and [Docker](https://github.com/openclaw/gogcli/actions/runs/33648975008) passed with the restored adapter coverage.

[...]

PR #1066, opened 2026-09-02
fix(drive): reject repeated page tokens in audit permission listing

SebTardif:
## Problem

`listDrivePermissionsForAudit` accepted every returned next-page token without remembering which tokens had already been used. A repeated token could keep both `drive audit` commands waiting indefinitely. The same helper feeds the preflight scan for both bulk permission commands, so those callers must also reject incomplete permission listings before acting on earlier plans.

## Change

Keep @SebTardif's production fix: delegate permission paging to the existing strict `collectAllPages` helper. Preserve request context, field selection, page size, shared-drive support, ordered successful results, and nil-on-error behavior. Repeated tokens and the existing collector page bound now stop the scan with an error.

Retain main's stronger shared-paginator tests to resolve the add/add conflict introduced by the earlier backup fix. Extend the existing Drive audit test file with exact token handoff and nil-on-loop assertions, ordered two-page success, a populated first page followed by a provider error, and four full command-path cases. The documentation states the scan-before-write boundary without claiming rollback after writes begin.

The command-path fixture first returns a complete actionable permission page for one file, then repeats a token while listing a second file. Both audits run with `--fail-found`; both bulk commands run with `--force`, without `--dry-run` or `--readonly`. All four must report the second file's pagination error with exit code 1, emit no success output, and send zero non-GET requests. This checks the actual CLI and generated Drive client rather than an illustrative pagination loop.

## Proof

The maintained regression suite was run against old production source using a Go overlay. The unguarded adapter made a third request, which the synthetic peer deliberately refused to bound the failure; all four CLI cases and the direct adapter regression failed as expected. With the production fix, the selected tests pass: the adapter stops after two calls, successful pages retain order, later provider errors discard partial permissions, and the four command cases return the file-specific error without output or writes.

```sh
go test ./internal/cmd -run '^(TestListDrivePermissionsForAudit|TestExecuteDrivePermissionScanFailurePreventsOutputAndWrites|TestDriveAudit|TestDriveBulk|TestCollectAllPages)' -count=1 -timeout=120s -v
go test -race ./internal/cmd -run '^(TestListDrivePermissionsForAudit|TestExecuteDrivePermissionScanFailurePreventsOutputAndWrites|TestDriveAudit|TestDriveBulk|TestCollectAllPages)' -count=3 -timeout=180s
make ci
```

The focused suite passed both on the contributor branch and on the integrated current-main candidate. Three focused race runs, full `make ci` (including lint, Node tests, generated docs and skill checks), and independent P0/P1 Codex review passed on the integrated candidate.

These are synthetic provider-fault tests through real generated clients and command execution, not a claim that a live Google service returned a faulty page. No Google account request, permission mutation, new grant, IAM setting, or billing change was needed.

Final head `6070f0f9204167858a4d1e6dd383b9e22cdbac34` passed [GitHub CI](https://github.com/openclaw/gogcli/actions/runs/33659723395) (Linux tests, Windows, Darwin/cgo, and worker checks) and [Docker](https://github.com/openclaw/gogcli/actions/runs/33659723301). GitHub confirms the add/add merge conflict is resolved.

## Context and credit

Thanks @SebTardif for the production repair and original adapter regression. The unguarded helper was introduced in [`d949466d`](https://github.com/openclaw/gogcli/commit/d949466d21c49022a5684af19d1a29297fd2d234), relative to raw parent `c84a09192c8092224bbe785184b885de491b99f4`; the introducing patch was inspected directly. The original body's attribution to #914 was incorrect.

Related guard work: [calendar #1004](https://github.com/openclaw/gogcli/pull/1004), [backup #1063](https://github.com/openclaw/gogcli/pull/1063), and [Drive sync #1065](https://github.com/openclaw/gogcli/pull/1065). The integrated candidate is based on main after [#1067](https://github.com/openclaw/gogcli/pull/1067), which separately fixes locked-flag-state ownership.

### DIFF
--- a/docs/drive-audits.md
+++ b/docs/drive-audits.md
@@ -102,6 +102,11 @@ gog drive bulk remove-public --parent <folderId> --dry-run
 gog drive bulk update-role --parent <folderId> --from writer --to reader --target contractor@example.com --dry-run
 ```
 
+Permission paging must finish for every file included in the scan. API failures,
+repeated page tokens, and page-limit errors fail without partial audit success
+output. Bulk operations complete this permission scan before writing changes, so
+a listing failure prevents writes; this does not roll back later write failures.
+
 ## Shared Drives
 
 The audit commands include shared drives by default where the underlying Drive
--- a/internal/cmd/drive_audit.go
+++ b/internal/cmd/drive_audit.go
@@ -233,9 +233,7 @@ func driveAuditItems(ctx context.Context, svc *drive.Service, fileIDRaw, parentR
 }
 
 func listDrivePermissionsForAudit(ctx context.Context, svc *drive.Service, fileID string) ([]*drive.Permission, error) {
-	out := make([]*drive.Permission, 0, 8)
-	var pageToken string
-	for {
+	fetch := func(pageToken string) ([]*drive.Permission, string, error) {
 		call := svc.Permissions.List(fileID).
 			SupportsAllDrives(true).
 			PageSize(100).
@@ -246,14 +244,11 @@ func listDrivePermissionsForAudit(ctx context.Context, svc *drive.Service, fileI
 		}
 		resp, err := call.Do()
 		if err != nil {
-			return nil, err
+			return nil, "", err
 		}
-		out = append(out, resp.Permissions...)
-		if resp.NextPageToken == "" {
-			return out, nil
-		}
-		pageToken = resp.NextPageToken
+		return resp.Permissions, resp.NextPageToken, nil
 	}
+	return collectAllPages("", fetch)
 }
 
 func driveSharingFinding(item driveTreeItem, perm *drive.Permission, internalDomains map[string]struct{}) (driveSharingAuditFinding, bool) {
--- a/internal/cmd/drive_audit_test.go
+++ b/internal/cmd/drive_audit_test.go
@@ -2,11 +2,14 @@ package cmd
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"io"
 	"net/http"
 	"strings"
+	"sync/atomic"
 	"testing"
+	"time"
 )
 
 func TestDriveAuditSharingFindsPublicAndExternal(t *testing.T) {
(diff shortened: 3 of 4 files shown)


steipete:
Landed in [8f0e9d35](https://github.com/openclaw/gogcli/commit/8f0e9d35ac7f861ac85565bb47ab40543436794e). Thanks @SebTardif for the production fix and original regression; your co-author credit is preserved. Main is clean, and all four candidate files (including the retained shared-paginator tests) match the reviewed integration.

The old-production Go-overlay control failed after the deliberately capped third request, both in the direct generated-client adapter test and in all four real CLI cases. After the fix, these checks passed:

```sh
go test ./internal/cmd -run '^(TestListDrivePermissionsForAudit|TestExecuteDrivePermissionScanFailurePreventsOutputAndWrites|TestDriveAudit|TestDriveBulk|TestCollectAllPages)' -count=1 -timeout=120s -v
go test -race ./internal/cmd -run '^(TestListDrivePermissionsForAudit|TestExecuteDrivePermissionScanFailurePreventsOutputAndWrites|TestDriveAudit|TestDriveBulk|TestCollectAllPages)' -count=3 -timeout=180s
make ci
```

The tests prove exact page-token handoff, ordered successful pages, nil partial results after a later API error, and two-call loop termination. Each command fixture first produces an actionable finding/permission plan, then fails on the second file: both audits return exit 1 rather than findings exit 3, and both forced bulk commands send zero writes. All four emit no success output. The focused suite also passed on the contributor branch; the full gate, race checks, and scoped-clean P0/P1 Codex review used the integrated current-main candidate.

Exact head `6070f0f9204167858a4d1e6dd383b9e22cdbac34` passed [CI](https://github.com/openclaw/gogcli/actions/runs/33659723395), including Linux, Windows, Darwin/cgo and worker checks, plus [Docker](https://github.com/openclaw/gogcli/actions/runs/33659723301). This is synthetic provider-fault proof through the real generated client, not a live faulty Google response. No Google account call, permission mutation, or new grant was made.

PR #38, opened 2026-01-04
feat(calendar): feature parity with google calendar api

salmonumbrella:
## Summary
Google Calendar feature parity with what's missing from the API (20 items):

1. --color flag (event color ID)
2. --visibility flag (default/public/private/confidential)
3. --transparency flag (busy/free aliases)
4. --send-updates flag (all/externalOnly/none)
5. Guest permissions: --guests-can-invite / --guests-can-modify / --guests-can-see-others
6. --with-meet (create Google Meet)
7. calendar focus-time subcommand
8. calendar out-of-office (ooo) subcommand
9. calendar working-location (wl) subcommand
10. Extended attendee format (email;optional;comment=)
11. --source-url / --source-title
12. Recurring scope support: --scope single/future/all + --original-start
13. --attachment (file URL)
14. --private-prop / --shared-prop (extended properties)
15. --private-prop-filter / --shared-prop-filter (events list)
16. --fields (field selection)
17. Enhanced event output for new fields
18. Real --scope=future handling (new series + truncate parent)
19. Recurrence truncation helpers + tests
20. Lint cleanup across cmd/config/secrets/tracking/googleauth/Gmail

## Testing
- golangci-lint run ./...
- go test ./...

PR #1084, opened 2026-09-05
fix(calendar): reject repeated instance page tokens

SebTardif:
## What Problem This Solves

`gog calendar delete` and `gog calendar update` with `--scope single` or `--scope future` resolve the target instance through `Events.Instances`. That helper rebuilt each page request from `NextPageToken` and never recorded tokens it had already used. If Google returned the same token again, instance lookup never finished, so the command never deleted or patched the event.

The same unguarded rebuild has already been fixed for CalendarList (`#1004`) and other `cmd` list helpers. Recurring instance lookup was still on the old loop, introduced with calendar recurrence parity in `#38` (`1d595747`, 2026-01-08).

## Evidence

Public path: `CalendarDeleteCmd.Run` and the update/edit commands call `resolveRecurringScopeResolution`, which calls `resolveRecurringInstanceID` whenever `--scope` is `single` or `future` and `--original-start` is set.

A local Calendar Instances peer that always returns `nextPageToken=stuck` (and a non-matching item) shows the hang versus the fail-closed walk. The unguarded rebuild kept requesting until the peer refused after 40 calls. The seen-token walk stopped after two calls.

```
$ go run C:/Users/sebta/.grok/tmp/instance-page-token-proof.go
OLD: unguarded Events.Instances rebuild
old calls=41 err=googleapi: got HTTP response code 429 with body: stopped after 40 requests
NEW: collectAllPages seen-token walk
new calls=2 err=pagination loop: repeated page token "stuck"
```

The same peer through `CalendarDeleteCmd` with `--scope single` and `--original-start 2025-01-02T10:00:00Z` now returns that pagination error, makes two instance list calls, writes no JSON, and never sends DELETE.

## Real behavior proof

- **Behavior or issue addressed:** Recurring instance lookup for calendar edit and delete hangs if `Events.Instances` repeats a page token.
- **Real environment tested:** Windows, Go 1.27.1, openclaw/gogcli at `25703c78` plus this patch, generated Calendar client against a local Instances peer.

  ```
  go run C:/Users/sebta/.grok/tmp/instance-page-token-proof.go
  ```

- **Evidence after fix:** terminal output from the patched walk:

  ```
  OLD: unguarded Events.Instances rebuild
  old calls=41 err=googleapi: got HTTP response code 429 with body: stopped after 40 requests
  NEW: collectAllPages seen-token walk
  new calls=2 err=pagination loop: repeated page token "stuck"
  ```

- **Observed result after fix:** The unguarded rebuild issued 41 instance list requests. After the patch, the same peer stops after two requests with `pagination loop: repeated page token "stuck"`. `calendar delete --scope single` does not emit a result or send DELETE once lookup fails.
- **What was not tested:** A live Google Calendar account returning a repeated Instances token.

Related: [#1004](https://github.com/openclaw/gogcli/pull/1004), [#38](https://github.com/openclaw/gogcli/pull/38), [#1066](https://github.com/openclaw/gogcli/pull/1066).

### DIFF
--- a/internal/cmd/calendar_delete_test.go
+++ b/internal/cmd/calendar_delete_test.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net/http"
 	"strings"
+	"sync/atomic"
 	"testing"
 
 	"google.golang.org/api/calendar/v3"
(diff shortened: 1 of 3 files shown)


steipete:
Superseded by #1087, which incorporates recurring-instance cycle protection while preserving immediate success when the requested occurrence is found. Failed lookup performs no event deletion. Thanks @SebTardif. Superseded by #1087, which incorporates cycle protection for Drive tree, inventory, and disk-usage scans. All three stock CLI commands were verified to fail without partial success output. Thanks @SebTardif. Superseded by #1087, which incorporates Gmail query-cycle protection, including repeated empty pages, while retaining existing result limits. Failure occurs before downstream message changes. Thanks @SebTardif.

PR #1046, opened 2026-08-29
fix(contacts): reject repeated page tokens in export group listing

SebTardif:
## What Problem This Solves

`gog contacts export` lists `ContactGroups` with `pageToken = resp.NextPageToken` and no seen-set when memberships need category names. Contact listing already uses `loadPagedItems`. Groups do not. If Google repeats `nextPageToken`, export never finishes and the CLI hangs until the process is killed.

## Why

The repo already has this guard. `collectAllPages` errors with `pagination loop: repeated page token`. Calendar listing uses it after [#1004](https://github.com/openclaw/gogcli/pull/1004). People raw email resolve uses it in [#1044](https://github.com/openclaw/gogcli/pull/1044). The Gmail `--from-contact` fallback uses it in [#1045](https://github.com/openclaw/gogcli/pull/1045). Group listing lives in the same `cmd` package, so it now calls that helper instead of growing a second loop.

## User Impact

A stuck ContactGroups page token now fails with `repeated page token` instead of hanging `gog contacts export` when a contact has group memberships.

## Evidence

terminal output from the unpatched group listing loop versus this patch. A People `httptest` server returns one connection with a group membership (so category names are fetched) and always returns `nextPageToken=stuck` from `ContactGroups.List`.

Unpatched (the HTTP peer answers immediately; the hang guard stops the loop after 8 list calls):

```console
$ go test ./internal/cmd -run TestContactsExport_GroupListingRejectsRepeatedPageToken -count=1 -timeout 30s -v
=== RUN   TestContactsExport_GroupListingRejectsRepeatedPageToken
    contacts_export_test.go:102: repeated page token looped: 9 list calls
    contacts_export_test.go:102: repeated page token looped: 10 list calls
    contacts_export_test.go:120: err = list contact groups: Get "http://127.0.0.1:50484/v1/contactGroups?alt=json&groupFields=groupType%2Cmetadata%2Cname&pageSize=1000&pageToken=stuck&prettyPrint=false": EOF after 10 list calls
--- FAIL: TestContactsExport_GroupListingRejectsRepeatedPageToken (0.04s)
FAIL
```

Patched (same command, same stuck token, returns immediately):

```console
$ go test ./internal/cmd -run TestContactsExport_GroupListingRejectsRepeatedPageToken -count=1 -timeout 15s -v
=== RUN   TestContactsExport_GroupListingRejectsRepeatedPageToken
    contacts_export_test.go:122: err = list contact groups: pagination loop: repeated page token "stuck" after 2 list calls
[...]

PR #1045, opened 2026-08-29
fix(gmail): reject repeated page tokens in --from-contact fallback

SebTardif:
## What Problem This Solves

`gog gmail search --from-contact NAME` falls back to paging `People.Connections.List` when `SearchContacts` misses. That fallback assigned `pageToken = resp.NextPageToken` with no seen-set. If Google repeats `nextPageToken`, search never finishes and the CLI hangs until the process is killed.

## Why

The repo already has this guard. `collectAllPages` errors with `pagination loop: repeated page token`. Calendar listing uses it after [#1004](https://github.com/openclaw/gogcli/pull/1004). People raw email resolve uses it in [#1044](https://github.com/openclaw/gogcli/pull/1044). The Gmail `--from-contact` fallback lives in the same `cmd` package, so it now calls that helper instead of growing a second loop.

## User Impact

A stuck People connections page token now fails with `repeated page token` instead of hanging `gog gmail search --from-contact` after a SearchContacts miss.

## Evidence

terminal output from the unpatched `--from-contact` fallback versus this patch. A People `httptest` server returns an empty `SearchContacts` result (so the list fallback runs) and always returns `nextPageToken=stuck` from `People.Connections.List`.

Unpatched (the HTTP peer answers immediately; the hang guard stops the loop after 8 list calls):

```console
$ go test ./internal/cmd -run TestGmailFromContactFallbackRejectsRepeatedPageToken -count=1 -timeout 30s -v
=== RUN   TestGmailFromContactFallbackRejectsRepeatedPageToken
    gmail_from_contact_test.go:71: repeated page token looped: 9 list calls
    gmail_from_contact_test.go:71: repeated page token looped: 10 list calls
    gmail_from_contact_test.go:92: err = resolve --from-contact fallback list: Get "http://127.0.0.1:57606/v1/people/me/connections?alt=json&pageSize=200&pageToken=stuck&personFields=names%2CemailAddresses&prettyPrint=false": EOF after 10 list calls
--- FAIL: TestGmailFromContactFallbackRejectsRepeatedPageToken (0.00s)
FAIL
```

Patched (same command, same stuck token, returns immediately):

```console
$ go test ./internal/cmd -run TestGmailFromContactFallbackRejectsRepeatedPageToken -count=1 -timeout 15s -v
=== RUN   TestGmailFromContactFallbackRejectsRepeatedPageToken
    gmail_from_contact_test.go:94: err = pagination loop: repeated page token "stuck" after 2 list calls
[...]

PR #1085, opened 2026-09-05
fix(drive): reject repeated tree inventory page tokens

SebTardif:
## What Problem This Solves

`gog drive tree`, `gog drive inventory`, and `gog drive du` walk folder children through `listDriveChildren` in `internal/cmd/drive_reporting.go`. That helper copied `nextPageToken` into the next `Files.List` request with no seen-set.

When Drive repeats a continuation token, the loop never terminates. Tree, inventory, and disk-usage scans keep requesting the same page and never return.

The same hang class is already closed for Drive sync push listing (#1065), Drive audit permission listing (#1066), calendar and Gmail listing (#1004), and People/Gmail-from-contact/contacts-export listing (#1044, #1045, #1046). Folder children used by the read-only reporting commands were still on the unguarded loop.

## Evidence

terminal output from the compiled `internal/cmd` listing binary after the patch. A stuck continuation token is rejected after two list calls, with no third request. The same error reaches the public commands:

```console
$ gogcli-drive-reporting.exe -test.run TestListDriveChildrenRejectsRepeatedPageToken -test.v
    drive_reporting_test.go:381: err = pagination loop: repeated page token "stuck" after 2 list calls

$ gogcli-drive-reporting.exe -test.run TestExecuteDriveReportingRejectsRepeatedPageToken -test.v
    drive_reporting_test.go:496: list Drive folder root: pagination loop: repeated page token "stuck"; no success output
--- PASS: TestExecuteDriveReportingRejectsRepeatedPageToken/tree
--- PASS: TestExecuteDriveReportingRejectsRepeatedPageToken/inventory
--- PASS: TestExecuteDriveReportingRejectsRepeatedPageToken/du
```

Before the patch, the same stuck token made a third list request and returned HTTP 400 from the safety cap (`unexpected extra page request after 3 list calls`) instead of stopping on the repeated token.

## Real behavior proof

- **Behavior or issue addressed:** Drive tree, inventory, and du hung when Google repeated a folder-list page token. Those commands never finished a folder walk.
- **Real environment tested:** Windows, Go 1.27.1, branch `fix/drive-children-page-token` from `origin/main` at `25703c78`, compiled `internal/cmd` listing binary.
- **Exact steps or command run after this patch:** Compiled the listing binary with `go`, then ran `gogcli-drive-reporting.exe` with `-test.v` on the helper hang-guard and the public `drive tree` / `drive inventory` / `drive du` command path.
- **Evidence after fix:** terminal output from that binary shows `pagination loop: repeated page token "stuck"` after 2 list calls for `listDriveChildren`, and `list Drive folder root: pagination loop: repeated page token "stuck"` with empty stdout for tree, inventory, and du.
- **Observed result after fix:** A repeated Drive children page token now fails closed with the shared paginator error after two requests. Distinct pages still concatenate in order. A later provider error still discards partial children. Query, page size 1000, `orderBy=folder,name`, and `driveFilesListCallWithDriveSupport` (`corpora=allDrives`) are unchanged.
- **What was not tested:** A live Google Drive account with a real repeated token.

## Summary

Route `listDriveChildren` through existing `collectAllPages`. Keep the original query, fields, page size, order, shared-drive flags, and empty-parent default to `root`.

Related: #1065, #1066, #1004, #1044, #1045, #1046.

Introduced in `e9c496ef` (2026-05-05, `feat(drive): add read-only reporting commands`, #554).

### DIFF
--- a/internal/cmd/drive_reporting.go
+++ b/internal/cmd/drive_reporting.go
@@ -364,10 +364,7 @@ func listDriveChildren(ctx context.Context, svc *drive.Service, parentID string,
 		parentID = driveRootID
 	}
 	q := buildDriveListQuery(parentID, "")
-	out := make([]*drive.File, 0, 64)
-	var pageToken string
-
-	for {
+	fetch := func(pageToken string) ([]*drive.File, string, error) {
 		call := svc.Files.List().
 			Q(q).
 			PageSize(driveDefaultPageSize).
@@ -380,16 +377,11 @@ func listDriveChildren(ctx context.Context, svc *drive.Service, parentID string,
 		).Context(ctx)
 		resp, err := call.Do()
 		if err != nil {
-			return nil, err
+			return nil, "", err
 		}
-		out = append(out, resp.Files...)
-		if resp.NextPageToken == "" {
-			break
-		}
-		pageToken = resp.NextPageToken
+		return resp.Files, resp.NextPageToken, nil
 	}
-
-	return out, nil
+	return collectAllPages("", fetch)
 }
 
 func driveOwners(f *drive.File) []string {
--- a/internal/cmd/drive_reporting_test.go
+++ b/internal/cmd/drive_reporting_test.go
@@ -1,10 +1,14 @@
 package cmd
 
 import (
+	"context"
 	"encoding/json"
+	"io"
 	"net/http"
 	"strings"
+	"sync/atomic"
 	"testing"
+	"time"
 )
 
 func TestExecuteDriveTreeJSON(t *testing.T) {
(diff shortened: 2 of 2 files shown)


steipete:
Superseded by #1087, which incorporates cycle protection for Drive tree, inventory, and disk-usage scans. All three stock CLI commands were verified to fail without partial success output. Thanks @SebTardif. Superseded by #1087, which incorporates Gmail query-cycle protection, including repeated empty pages, while retaining existing result limits. Failure occurs before downstream message changes. Thanks @SebTardif.

PR #1091, opened 2026-09-06
fix(appscript): keep content filenames in one output field

haosdent:
`appscript content` printed API filenames directly into TSV rows, so a valid filename containing tabs or line breaks could create extra fields or rows. Escape those characters in text/plain output while preserving the exact filename in JSON. Move the existing single-line formatter into shared output helpers and preserve the trimmed comment variant's behavior. Thanks @haosdent for the original output-sanitization change.

The proposed `appscript push` command is deferred and is not part of this revision. Google's `projects.updateContent` replaces the whole project. A live disposable-project probe found no `GetContent` ETag and confirmed that `UpdateContent` accepted a deliberately stale `If-Match` header. A read/merge/write sequence can therefore overwrite concurrent remote edits or delete files absent from its preview. A second read only narrows that race; an explicit write-ownership/overwrite contract needs a maintainer decision before this mutation command is added.

Validation:

- Text, plain, and JSON regression cases pass; shared notes/comment formatting remains equivalent.
- Isolated Codex review is scoped clean through P2 with full helper context.
- Real Google API proof: created a disposable project containing a filename with tabs/newlines, observed the released CLI split its TSV row, and verified the final compiled CLI emits a single field while JSON retains the original name. The project was trashed.
- A separate live conditional-write probe confirmed the API limitation described above; its disposable project was also trashed.
- The final binary does not expose the deferred push command. Full local and exact-head CI gates are required before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.40.1 - Unreleased
 
+- Apps Script: keep file names within one TSV field in `appscript content`, escaping line breaks instead of splitting output rows. (#1091) — thanks @haosdent.
 - Search Console: add `searchconsole inspect` for per-URL index status via the URL Inspection API (coverage state, indexing/page-fetch/robots.txt state, canonical, sitemaps, last crawl time), using the existing `webmasters` OAuth scope. (#1094) — thanks @laihenyi.
 - Search Console: preserve permission-denied exit codes when adding API setup or scope guidance. (#1094)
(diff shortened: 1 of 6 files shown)


steipete:
Thanks @haosdent — the content-output fix landed in [fa705714](https://github.com/openclaw/gogcli/commit/fa70571484e51e4124f12e164413bf1cd8dc709d), with human contributor credit preserved. Real Google data reproduced the broken TSV row; the final CLI escapes it and preserves JSON. Local checks, isolated P2 review, and [exact-head CI](https://github.com/openclaw/gogcli/actions/runs/34720380520) passed. The push proposal remains deferred pending an explicit write-ownership contract; its conditional-write limitation and provider probe are documented in the PR body.

PR #1114, opened 2026-09-11
fix(sheets): preserve zero indices when inserting dimensions

### LINKED ISSUE #1107: sheets insert cannot insert before first row or column
`gog sheets insert ID Sheet cols 1`

fails with:

`range.startIndex must be set`

Looks like JSON serialization is eating our zero. We need zeros in this case. They are important. Don't eat them. Zeros are heroes.

steipete:
Inserting before row 1 or column A failed because the Google client omitted `startIndex: 0` from the request. Reuse the dimension-range serialization helper already used by resize and delete, including explicit zero sheet IDs.

Regression coverage checks raw request field presence for both dimensions and sheet IDs 0 and 7. The tests fail on main and pass with the fix. The real built CLI reproduced Google's 400 response before the fix; afterward both insertions succeeded and a synthetic sentinel moved from A1 to B2 in a disposable spreadsheet.

Closes #1107. Thanks @gurgeous for the report. The changelog entry is reserved for the final notes/dependency PR to avoid conflicts between independent branches.

### DIFF
--- a/internal/cmd/sheets_insert.go
+++ b/internal/cmd/sheets_insert.go
@@ -98,16 +98,18 @@ func (c *SheetsInsertCmd) Run(ctx context.Context, flags *RootFlags) error {
 		return usagef("unknown sheet %q", sheetName)
 	}
 
+	dimRange := &sheets.DimensionRange{
+		SheetId:    sheetID,
+		Dimension:  apiDimension,
+		StartIndex: startIndex,
+		EndIndex:   endIndex,
+	}
+	forceSendDimensionRangeZeroes(dimRange)
 	req := &sheets.BatchUpdateSpreadsheetRequest{
 		Requests: []*sheets.Request{
 			{
 				InsertDimension: &sheets.InsertDimensionRequest{
-					Range: &sheets.DimensionRange{
-						SheetId:    sheetID,
-						Dimension:  apiDimension,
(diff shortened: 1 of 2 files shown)


steipete:
Maintainer verification at ecd76454cbeab4e958af9f877a80f1d086673e09:

- Added raw JSON regression cases for both dimensions on sheet IDs 0 and 7. All four failed against main because `startIndex` was absent; all pass with the fix.
- Ran the real built CLI against a disposable Google spreadsheet. Before the fix, inserting before row 1 and column A each returned Google's `range.startIndex must be set` 400 error. With the fix, both succeeded and a synthetic sentinel moved from A1 to B2, verified by `sheets get`.
- Full local Go/JavaScript tests, formatting, lint, deadcode, docs coverage and agent-skill checks passed. The documentation generator needed system `/bin/bash`: this host's Homebrew Bash stalls writing a 7 KB heredoc. No tests were skipped or changed for that workaround.
- Local and committed-branch Codex autoreviews are clean through P2.

The changelog bullet will be carried by the final notes/dependency PR. This PR is prepared for orchestrator approval; no merge has been performed.

Cross-platform CI is green on this exact head: https://github.com/openclaw/gogcli/actions/runs/34649488751 (Linux tests/lint/deadcode/docs, Windows, macOS cgo, tracking worker).

PR #1115, opened 2026-09-11
feat(sheets): select a tab in raw API reads

### LINKED ISSUE #1104: `gog sheets raw --sheet <some_tab>`
`raw` returns a lot of data. Let us request a single sheet. I have some big sheets. Real big.

steipete:
Add `sheets raw --sheet <title>` so large spreadsheets can be inspected one tab at a time. The Sheets API filters the response before download; spreadsheet-level metadata and the existing grid-data opt-in remain intact. Quoting the exact title also handles apostrophes, whitespace and A1-like names without interpreting them as ranges.

Validated the built binary against a disposable two-tab Google spreadsheet: unfiltered reads returned both tabs and selected reads returned only Data, both with and without grid data. Selected grid readback retained the expected synthetic cell. Request-level tests cover title escaping and grid selection.

Closes #1104. Thanks @gurgeous for the request. The changelog line is reserved for the final notes/dependency PR.

### DIFF
--- a/docs/commands/gog-sheets-raw.md
+++ b/docs/commands/gog-sheets-raw.md
@@ -39,6 +39,7 @@ gog sheets (sheet) raw <spreadsheetId> [flags]
 | `--readonly` | `bool` | false | Block mutating API requests at runtime; auth add also requests read-only OAuth scopes |
 | `--results-only` | `bool` |  | In JSON mode, emit only the primary result (drops envelope fields like nextPageToken) |
 | `--select`<br>`--pick`<br>`--project` | `string` |  | In JSON mode, select comma-separated fields (best-effort; supports dot paths). Desire path: use --fields for most commands. |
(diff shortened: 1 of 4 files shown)


steipete:
Maintainer proof at b487598ba31be002f746adbbbe823fa813c3aeb2:

- The built CLI read a disposable two-tab spreadsheet with and without `--sheet Data`, both with and without `--include-grid-data`. Unfiltered reads returned both tabs; selected reads returned only Data while retaining spreadsheet metadata. The selected grid contained the expected synthetic sentinel.
- Request regressions cover A1-like titles, apostrophes, whitespace, grid-data selection and raw metadata preservation.
- `make ci` passed, using system Bash for this host's documented heredoc issue. Local and committed-branch Codex autoreviews are clean through P2.
- Exact-head cross-platform CI passed: https://github.com/openclaw/gogcli/actions/runs/34650155185

Prepared for squash after orchestrator approval. Release notes are carried by the final notes/dependency PR to keep sibling branches independent.

PR #1118, opened 2026-09-11
chore(deps): refresh dependencies and stabilize backup tests

steipete:
Refresh Go networking/authentication/cryptography support modules and tracking-worker tooling. Wrangler's updated dependency graph brings Sharp to 0.35.4, fixing GHSA-rgj7-g3m4-5g8c; the worker audit now reports zero known vulnerabilities. Go 1.26 compatibility, the Go 1.27.1 toolchain, pnpm 11.25.0 and the 24-hour worker release-age policy remain unchanged.

This is the final notes PR for the phase-four batch. Land it after #1114, #1115, #1116, #1117 and #1119. It carries the complete ordered Unreleased section so the independent code PRs do not conflict on CHANGELOG.md. No version, tag or release is created.

Validation includes a rebuilt CLI making an authenticated Google Sheets read, frozen worker installation, worker lint/build and all 15 tests, plus live local workerd responses for `/health` and the invalid-token pixel fallback. The full local Go gate and local/branch Codex autoreviews passed. Fresh exact-head CI is green for the fixture repair below: https://github.com/openclaw/gogcli/actions/runs/34651955204.

The newly published Wrangler 4.131.1 and today's genproto snapshots are held for maturity; all pinned GitHub Actions match their current releases. Other graph-only modules remain controlled by their upstream dependencies.

The first CI run exposed a Git 2.55 cleanup race in the real-Git async checkpoint test: detached automatic maintenance could write into the bare fixture after the test completed. The fixture now keeps maintenance enabled in the foreground for both the local caller and bare receiver. A repeated trace verifies no detached maintenance remains; the push/drain assertions are unchanged.

PR #1119, opened 2026-09-11
feat(slides): delete multiple elements in one batch

### LINKED ISSUE #1109: feat(slides): no way to delete a page element — and it is the only route to clearing a wordArt headline
### What is missing

There is no way to delete a page element. `gog slides element` has `create-shape`, `create-line`, `transform` and `style`; `gog slides` has `delete-slide` but no `delete-element`. The Slides API request is `deleteObject`, which takes a single `objectId`.

### Why it matters

Adapting a slide copied from a template almost always means removing the elements whose slot the new content takes. Without a delete, the options are to leave stale placeholder elements on the slide or to drop the whole slide and lose its design.

It is also the **only** way to clear a `wordArt` element. `presentations.batchUpdate` has no request that sets WordArt text, `insert-text` does not accept it, and `replace-text --object` refuses with *"object-scoped replace-text currently supports shape text objects only"* while `replace-text --all` reports `replaced=0` against text that `raw` shows as `{"wordArt": {"renderedText": "..."}}`. So a deck built from a template whose headlines are WordArt cannot be finished through the CLI at all, and the element cannot even be removed to substitute a text box.

### Suggested shape

```
gog slides element delete <presentationId> <objectId>... [--force]
```

Accepting several ids matters, because the realistic case is clearing three or four elements from one slide, and each extra invocation is another `presentations.batchUpdate` against the 60-write-per-minute quota (see #1097).

### Notes

`raw` already surfaces `wordArt` elements, so a `replace-text` that matched a `renderedText` and could not act on it would be worth a warning rather than a silent `replaced=0`. That silence is what makes the failure hard to find: a walker that reads `shape.text.textElements` never sees the element, so the deck looks finished while every headline still carries template copy.

steipete:
Extend `slides element delete` to accept several object IDs and submit their DeleteObject requests together in one atomic API batch. Existing single-ID JSON/text output, confirmation and dry-run behavior remain intact; multiple targets return `objectIds`. Reuse the existing target validation to reject duplicate or empty IDs before making a request.

Single-element deletion was already available. This completes the requested multi-target operation and documents the WordArt limitation and delete/recreate workflow; it does not add automatic WordArt warnings to text replacement.

Live built-CLI proof: created two synthetic elements in a disposable presentation, verified non-interactive refusal without force, inspected the two-request dry run, deleted with force, then read back both elements absent with the slide preserved. Tests cover request grouping, single-ID output compatibility, validation, confirmation and dry runs. Local Codex autoreview is clean through P2.

Closes #1109. Thanks @sebsnyk for the request. The changelog bullet is in #1118, which must merge after this and the other code PRs.

### DIFF
--- a/docs/commands.generated.md
+++ b/docs/commands.generated.md
@@ -680,7 +680,7 @@ Generated from `gog schema --json`.
       - [`gog slides (slide) element alt-text <presentationId> <objectId> [flags]`](commands/gog-slides-element-alt-text.md) - Set or clear element accessibility text
       - [`gog slides (slide) element create-line <presentationId> <slideId> [flags]`](commands/gog-slides-element-create-line.md) - Create a native line on a slide
       - [`gog slides (slide) element create-shape <presentationId> <slideId> [flags]`](commands/gog-slides-element-create-shape.md) - Create a native shape on a slide
-      - [`gog slides (slide) element delete (rm) <presentationId> <objectId>`](commands/gog-slides-element-delete.md) - Delete one page element
+      - [`gog slides (slide) element delete (rm) <presentationId> <objectId> ...`](commands/gog-slides-element-delete.md) - Delete one or more page elements
       - [`gog slides (slide) element group <presentationId> <objectId> ... [flags]`](commands/gog-slides-element-group.md) - Group two or more elements
       - [`gog slides (slide) element style <presentationId> <objectId> [flags]`](commands/gog-slides-element-style.md) - Style a shape fill/outline or a line
       - [`gog slides (slide) element transform (move,resize,rotate) <presentationId> <objectId> [flags]`](commands/gog-slides-element-transform.md) - Move, resize, rotate, or replace an element transform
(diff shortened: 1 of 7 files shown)


steipete:
Maintainer verification at a0fb32d553dc0cb7f0b72dd599b62b68b8ad9e41:

- Ran the built CLI against a disposable Google presentation containing two synthetic elements. Non-interactive deletion without force returned usage exit 2. Dry run contained both DeleteObject requests. Forced deletion returned both IDs, and raw API readback confirmed both absent with the slide preserved.
- Tests cover multi-target batching, legacy single-ID JSON, duplicate/empty-ID rejection, confirmation and auth-free dry runs.
- Full local `make ci` passed. Local and branch Codex autoreviews are clean through P2.
- Exact-head cross-platform CI passed: https://github.com/openclaw/gogcli/actions/runs/34651236383

The changelog entry and contributor thanks are in #1118, which must land last. WordArt deletion/recreation is documented; automatic replacement warnings remain outside this change. Ready for orchestrator-approved squash; no merge has been performed. The disposable presentation was trashed and cleanup verified.

PR #1121, opened 2026-09-11
feat(api): cache Discovery documents with bounded TTL

### LINKED ISSUE #1106: `gog api call` needs a cache
Right now `api call` downloads and parses the schema on each request. That's, uh, really slow. Adds 0.5-1s per request on my machine, so you can imagine that gog enthusiasts like me would suffer horribly if we adopted `api call`. Can we get a cache? It's not like that schema is changing very often. Just caching for a day would make a huge difference.

steipete:
Cache successful Discovery documents for `api call` and `api describe` for 24 hours, retaining at most 32 documents/64 MiB under gog's configured cache root. Endpoint/version/fallback keys are isolated; private atomic writes and serialized eviction keep the cache bounded. Expired or corrupt entries and unavailable cache storage fall back to the network. `--no-cache` bypasses both reads and writes.

API responses and `api list` remain uncached. Account authorization, method policy and read-only/send guards still run for every API call.

Built-CLI proof across separate processes verified cold/warm request counts, bypass without cache changes, expiry and corruption recovery. Actual Google Discovery outputs matched cold/warm/bypass, and two authenticated Drive API calls agreed. One schema-only run measured 250/68/483 ms respectively. Tests cover fallback isolation, failures, cancellation, concurrent writes and disk limits; Codex autoreview is clean through P2.

Closes #1106. Thanks @gurgeous. Changelog context is reserved for the final notes PR; exact-head CI proof will be posted before handoff.

### DIFF
--- a/docs/automation.md
+++ b/docs/automation.md
@@ -187,6 +187,20 @@ fi
 
 New classifications may be added. Keep a generic non-zero fallback.
 
+## Discovery document cache
+
+`gog api call` and `gog api describe` share a 24-hour on-disk cache of
+Discovery documents under the configured cache directory's `discovery`
+subdirectory. It keeps at most 32 documents and 64 MiB, evicting the oldest
+documents; individual documents are limited to 16 MiB. API versions, endpoint
+overrides and service-hosted fallback behavior use separate cache keys.
+
+Pass `--no-cache` to either command to fetch without reading or writing this
+cache. Missing, expired or corrupt entries are fetched again. Cache failures
+do not prevent network access, and failed fetches are not cached. `api list`
+and actual API responses remain uncached; authorization and command-policy
+checks still run on every call.
+
 ## MCP discovery
 
 MCP uses its standard `tools/list` request for client-side tool discovery. To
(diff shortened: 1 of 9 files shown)


steipete:
Maintainer verification at 16739f5d8a77aaf0c70a967a75312f7c6dc82de8:

- Separate built-CLI processes made one schema fetch across cold/warm calls. Bypass fetched again without changing the cache; corruption and expiry each caused a fresh fetch. Cached method-policy denial was still enforced.
- Actual Google schema outputs matched across cold/warm/bypass. The final source run measured 342/68/279 ms respectively; two authenticated no-match Drive queries also agreed. Only Discovery documents are cached, never API responses.
- Full local make ci and race-enabled Discovery tests passed. Coverage includes HTTP/decode failures, unavailable cache storage, key/fallback isolation, cancellation, atomic concurrent writes and disk limits.
- Local and committed-branch Codex autoreviews are clean through P2.
- Exact-head CI passed: https://github.com/openclaw/gogcli/actions/runs/34654683220

Prepared for orchestrator-approved squash; closes #1106 on merge. The changelog entry and thanks are reserved for the final batch notes PR.

PR #1122, opened 2026-09-11
fix(slides): preserve leading style when replacing text

### LINKED ISSUE #1111: bug(slides): insert-text --replace discards the run's text style — 67 elements silently fell back to Calibri
### What happens

`insert-text --replace` is a `deleteText` followed by an `insertText`, and the inserted run inherits the style in force at the insertion point rather than the run it replaced. Emptying the shape destroys that run, so the new text lands on the placeholder default.

Filling 67 text elements on a deck copied from a themed template moved every one of them from the template's two brand faces to **Calibri**, and dropped bold and the foreground colour at the same time. Font **size** survived.

That last detail is what makes this expensive: size surviving means the slide's proportions look right in a thumbnail, so the damage is invisible until someone opens the deck. Recovering it meant reading the source presentation, collecting the first run style per `objectId`, and re-applying `updateTextStyle` across all 67 elements.

### Repro

```
# element carries Geist Mono 21.75pt, bold, brand foreground
gog slides insert-text <deck> <objectId> "NEW TEXT" --replace
gog slides raw <deck> | grep -A3 fontFamily     # now Calibri, bold and colour gone, 21.75pt kept
```

Running `insert-text --replace` a second time on an element whose style has since been fixed **does** preserve it, because by then the insertion point carries the right style. So the failure only shows on the first write to a freshly copied template, which is exactly when a caller is least likely to check.

### Suggested shape

Capture the style of the first run before the delete and re-apply it to the inserted range, either by default or behind a flag:

```
gog slides insert-text <presentationId> <objectId> <text> --replace [--keep-style]
```

Default-on seems right. A caller replacing the text of a styled shape almost never wants the placeholder default, and the current behaviour has no use case that a subsequent `style-text` could not express deliberately.

### Notes

Where a shape holds several differently styled runs, preserving the first run's style for the whole insert is still wrong, but it is much closer than falling back to the theme. A range-scoped edit that keeps surrounding runs intact is the general answer; a whole-shape replace only needs the first-run style.

Related: #1097, since a deck fill is already many separate `presentations.batchUpdate` calls, and a style restamp currently doubles that count.

steipete:
Replacing copied-template text could turn branded text into the final newline's fallback font, weight and color. DeleteText left that newline behind, then InsertText inherited its style. Insert the new text before removing the old suffix so it inherits the leading text style, including master/layout inheritance.

Read the target and pin its revision first: empty shapes and cells require insertion alone because suffix deletion would be an invalid empty range. Normalize the control/private-use characters Google strips before calculating the UTF-16 boundary. Empty or fully stripped replacements clear existing text or return a successful no-op. Dry runs stay auth-free, and unrelated notes-replacement ordering is unchanged.

Reproduced on main with inherited Roboto Mono/bold/brand-color text and a Calibri/regular/black final newline. The built fixed CLI passed 17 real Google cases covering inherited and explicit controls, Unicode, stripped characters, empty shapes/cells and clear-only behavior. All disposable presentations were trashed and verified. Tests cover request ordering, exact offsets, revision protection, page kinds and empty boundaries; final Codex autoreview is clean through P2.

Closes #1111. Thanks @sebsnyk. Changelog context is reserved for the final notes PR; exact-head CI proof will be posted before handoff.


steipete:
Maintainer proof at 0b2001645af5f453f8d6042cb71a235acd793413:

- Reproduced on main with a copied placeholder whose visible text had direct style {} and inherited Roboto Mono/bold/#993366/21.75pt from its parent, while the final newline explicitly used Calibri/regular/black. Clear-then-insert changed the visible replacement to that newline's fallback style.
- The built fixed CLI passed 17 real Google cases: inherited title/body, uniform and mixed explicit controls, emoji/tab/newline/stripped characters, empty shapes/cells and empty/stripped-only replacements. Inserting first preserved the leading style. Empty targets used insertion alone; no-op replacements returned without an empty API batch. All fixtures were trashed and verified.
- Full local make ci and final local/branch Codex autoreview through P2 passed. Updated x/tools v0.50 formatting and Darwin/Linux deadcode checks also passed on this branch.
- The generated command reference is retained in the final diff; a staging overlap during regeneration was corrected before this final head.
- Exact-head CI passed: https://github.com/openclaw/gogcli/actions/runs/34655573584

Ready for orchestrator-approved squash; closes #1111 on merge. Release notes are in the final batch notes PR.

PR #1123, opened 2026-09-11
feat(slides): add paragraph styling for shapes and cells

### LINKED ISSUE #1098: feat(slides): no paragraph-style command — alignment, spacing and indentation unreachable (updateParagraphStyle)
### The gap

`gog slides` has no paragraph-level command. `style-text` covers character style, `bullets` toggles list bullets, and `table cell style --content-align` covers vertical alignment inside a table cell. Nothing reaches `updateParagraphStyle`, so there is no way to set:

- `alignment` (START / CENTER / END / JUSTIFIED)
- `lineSpacing`
- `spaceAbove` / `spaceBelow`
- `indentStart` / `indentEnd` / `indentFirstLine`
- `direction`

### Why it bites

Corporate templates commonly set body placeholder paragraphs to centred. Text written into a shape with `insert-text` inherits the paragraph style in force, so a left-reading list of bullet lines lands centred and reads as a ragged column. A deck built as native shapes, which is the only route on tenants where image insertion is refused, is therefore unfinished after `create-shape` + `style` + `insert-text` + `style-text`: the text is the right size, weight and colour, and still in the wrong place.

There is also no way back. `element style` does not touch text, and re-inserting the text inherits the same paragraph style again.

The practical consequence is that a script otherwise entirely driven by `gog` has to open a Google API client library for one `updateParagraphStyle` request, purely because the CLI cannot express it. That is a single-purpose second auth path for one field.

### Manual bullets are the second symptom

With no `indentFirstLine` / `indentStart`, a manually written bullet line that wraps returns flush left on its second line and reads as a new item. `bullets` creates native list bullets but gives no control over the hanging indent, so hand-drawn bullet glyphs inside a shape cannot be made to hang.

### Proposed surface

Mirror `style-text`, which already takes an object id and a UTF-16 range:

```
gog slides paragraph-style <presentationId> <objectId> [flags]
      --range=STRING             UTF-16 text range as start:end; defaults to all text
      --row=INT-64 --col=INT-64  Target a table cell instead of a shape
      --align=START|CENTER|END|JUSTIFIED
      --line-spacing=FLOAT-64    Percent, 100 = single
      --space-above=FLOAT-64     Points
      --space-below=FLOAT-64     Points
      --indent-start=FLOAT-64    Points
      --indent-end=FLOAT-64      Points
      --indent-first-line=FLOAT-64
      --direction=LEFT_TO_RIGHT|RIGHT_TO_LEFT
```

One `UpdateParagraphStyleRequest`, `fields` built from whichever flags were passed, same as `style-text` does for `textStyle`. Table cells need the `cellLocation` on the request, which is why `--row` / `--col` appear here as they do on `table cell style`.

`read-slide --detail` already reports `alignment` on each text element, so the CLI can show that a placeholder is centred and has no command to change it. That asymmetry is the clearest statement of the gap.

steipete:
Add `slides paragraph-style` for alignment, line spacing, paragraph spacing, indentation and direction on shapes or individual table cells. Optional UTF-16 ranges affect intersecting paragraphs; omitted ranges select all text. Only supplied fields are changed, including explicit zero dimensions.

Reuse the existing mutation helpers for complete auth-free dry-run plans and normal command guards. Cell targets are validated against the current table and protected by the presentation revision. The supplied baked profiles leave the new mutation disabled.

The built CLI passed live Google checks for range-scoped alignment while preserving the second paragraph, spacing reset to zero, hanging indentation, table-cell alignment/line spacing/direction and invalid-row refusal. The disposable presentation was trashed and verified. Tests cover field masks, wire-level zeros, ranges, validation, cell targeting/revision checks and complete dry-run values. Codex autoreview is clean through P2.

Closes #1098. Thanks @sebsnyk. The changelog entry is reserved for the final notes PR; exact-head CI proof will be posted before handoff.

### DIFF
--- a/.agents/skills/gog-slides/SKILL.md
+++ b/.agents/skills/gog-slides/SKILL.md
@@ -46,6 +46,7 @@ gog --readonly --account user@example.com slides --help
 | `locate` | Locate text in shapes and table cells with object IDs and UTF-16 ranges |
 | `move-slide` | Move a slide to a zero-based insertion index |
 | `new-slide` | Create a native themed slide |
+| `paragraph-style` | Set paragraph alignment, spacing, indentation, or direction |
 | `raw` | Dump raw Google Slides API response as JSON (Presentations.Get; lossless; for scripting and LLM consumption) |
 | `read-slide` | Read slide content: speaker notes, text elements, and images |
 | `replace-slide` | Replace an existing slide image from a local file or public URL |
--- a/docs/commands.generated.md
+++ b/docs/commands.generated.md
@@ -696,6 +696,7 @@ Generated from `gog schema --json`.
     - [`gog slides (slide) locate (find-element) <presentationId> <text> [flags]`](commands/gog-slides-locate.md) - Locate text in shapes and table cells with object IDs and UTF-16 ranges
     - [`gog slides (slide) move-slide --to-index=TO-INDEX <presentationId> <slideId>`](commands/gog-slides-move-slide.md) - Move a slide to a zero-based insertion index
     - [`gog slides (slide) new-slide <presentationId> [flags]`](commands/gog-slides-new-slide.md) - Create a native themed slide
+    - [`gog slides (slide) paragraph-style <presentationId> <objectId> [flags]`](commands/gog-slides-paragraph-style.md) - Set paragraph alignment, spacing, indentation, or direction
     - [`gog slides (slide) raw <presentationId> [flags]`](commands/gog-slides-raw.md) - Dump raw Google Slides API response as JSON (Presentations.Get; lossless; for scripting and LLM consumption)
     - [`gog slides (slide) read-slide <presentationId> <slideId> [flags]`](commands/gog-slides-read-slide.md) - Read slide content: speaker notes, text elements, and images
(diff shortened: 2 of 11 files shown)


steipete:
Maintainer proof at 53b37d547a6224720b80a28927d395a7d382a634:

- The built CLI set shape paragraph alignment and spacing, changed only the first intersecting paragraph, reset spacing to explicit zero, and applied a hanging indent. Raw Google readback verified each result and preserved the second paragraph's alignment.
- A real table-cell update verified alignment, line spacing and direction; invalid-row input was refused. The disposable presentation was trashed and verified.
- Tests verify selective field masks, serialized zero dimensions, UTF-16 ranges, table revision protection, invalid inputs, and complete auth-free dry-run plans containing the actual values and ranges.
- Full local make ci, final local/branch Codex autoreview through P2, and x/tools v0.50 formatting plus Darwin/Linux deadcode checks passed.
- Exact-head CI passed: https://github.com/openclaw/gogcli/actions/runs/34655542638

Ready for orchestrator-approved squash; closes #1098 on merge. Release notes are in the final batch notes PR.

Issue #1058, opened 2026-09-01
regression(gmail): plain-text drafts still hard-wrap after Gmail web sends them


daveotero:
## Summary

The plain-text hard-wrapping reported in #475 still occurs in `v0.38.1` for drafts that a human opens and sends from the Gmail web UI.

PR #476 fixed Gog's initial MIME construction: a fresh `--body` draft now uses quoted-printable encoding, and its decoded raw body remains one logical line. However, after that draft is opened and sent from Gmail web, the sent message contains real CRLF breaks around 70 characters. Gmail's draft editor looks normal before send, so the damage is only visible in the sent message/recipient view.

This appears to be an incomplete fix or integration regression specific to the Gog draft -> Gmail web editor -> send path.

## Environment

- Gog: `v0.38.1 (324f656a 2026-08-26T04:37:04Z)`
- Install: standalone Windows `amd64` binary
- OS: Windows 11
- Browser: Chrome / Gmail web
- Draft created through Gmail API by Gog, then sent manually in Gmail web

## Steps to reproduce

1. Supply one long paragraph with no `\r` or `\n` characters:

   ```text
   This is one deliberately long plain-text paragraph supplied to gog without any newline characters. It should remain a single logical line in the MIME body so Gmail can wrap it responsively, but the current draft serializer inserts hard line breaks.
   ```

2. Create the draft:

   ```bash
   gog --account me@example.com gmail drafts create \
     --to recipient@example.com \
     --subject "Plain-text wrapping reproduction" \
     --body "<the single-line paragraph above>" \
     --json --no-input
   ```

3. Open the draft in Gmail web. It appears correctly reflowed in the editor.
4. Send it using Gmail web.
5. Fetch the sent message with `gog gmail get <messageId> --format=raw --json`, base64url-decode the RFC822 message, then quoted-printable-decode the body.

## Before-send control

A fresh `v0.38.1` draft created with the text above has:

```text
Content-Type: text/plain; charset="utf-8"
Content-Transfer-Encoding: quoted-printable
```

After base64url and quoted-printable decoding:

```text
INPUT_HAS_NEWLINE False
DECODED_NONEMPTY_LINE_LENGTHS [248]
```

So #476 is working at initial draft construction time.

## Actual result after Gmail web sends the draft

The sent message still has:

```text
Content-Type: text/plain; charset="UTF-8"
Content-Transfer-Encoding: quoted-printable
```

[...]

PR #1126, opened 2026-09-12
feat(gmail): support raw RFC822 drafts

darkamenosa:
Stage a prebuilt RFC822 message with `gmail drafts create/update --raw-file PATH|-`, then retrieve the stored MIME with `gmail drafts get --format raw`. Draft creation permits absent recipients; raw update replaces the whole message and never merges old recipients, attachments or reply state. Caller MIME bytes are submitted unchanged, with sender and recipient-header validation shared with raw send.

No-send policies continue to permit draft staging. Read-only mode blocks writes before authentication; offline dry-runs expose only source, byte count, SHA-256 and IDs. Raw send and drafts use a bounded 35 MiB reader, including stdin; import retains its existing input contract. Text retrieval emits decoded RFC822 without an added newline, while JSON retains the draft envelope. Compose conflicts, sender aliases, direct-token account selection, input bounds and safe diagnostics have regression coverage.

The contributor's implementation was brought onto current main, retaining the concise architecture guide and updating the current changelog. Human credit: Tuyen <hxtxmu@gmail.com> (@darkamenosa).

Validation: full `make ci`, generated command references, and isolated Codex review through P2. A signed compiled CLI created a recipient-free multipart draft in an existing authorized Google account, verified its custom display name, alternatives and CID bytes, compared plain raw output with decoded JSON, replaced it through stdin, and verified stable draft ID/new message ID with old attachment and reply state absent. The draft was deleted. No messages were sent; every operation used `--gmail-no-send`. Gmail may normalize stored MIME; exact upload bytes are verified by request-level regressions. This does not claim to fix Gmail web-send wrapping in #1058.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.40.1 - Unreleased
 
+- Gmail: stage and replace raw RFC822 drafts, retrieve their MIME, and preview safely offline; bound raw draft/send input to 35 MiB while preserving no-send policies. (#1126) — thanks @darkamenosa.
 - Apps Script: keep file names within one TSV field in `appscript content`, escaping line breaks instead of splitting output rows. (#1091) — thanks @haosdent.
 - Search Console: add `searchconsole inspect` for per-URL index status via the URL Inspection API (coverage state, indexing/page-fetch/robots.txt state, canonical, sitemaps, last crawl time), using the existing `webmasters` OAuth scope. (#1094) — thanks @laihenyi.
 - Search Console: preserve permission-denied exit codes when adding API setup or scope guidance. (#1094)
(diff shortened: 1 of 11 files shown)


steipete:
Thanks @darkamenosa — raw draft creation, complete replacement and retrieval landed in [33d0d010](https://github.com/openclaw/gogcli/commit/33d0d010c5b7523fcc0ed59bc8bdd5ad0a468b8a), with Tuyen's contributor credit preserved. Full local checks, isolated P2 review and [exact-head CI](https://github.com/openclaw/gogcli/actions/runs/34721517933) passed. The signed CLI round-tripped a recipient-free multipart/CID draft through Google, replaced it via stdin and deleted it; no messages were sent. Exact submission bytes, safe previews and no-send behavior are covered by regressions. This does not resolve Gmail web-send wrapping in #1058.

PR #1138, opened 2026-09-17
fix(gmail): return first thread-detail fetch error

SebTardif:
When `gog gmail search QUERY` lists threads, it loads each thread with a concurrent `Threads.Get`. If any of those Gets failed, the follow-up sequential retry ignored the recovered thread body (`_, err :=`). A later success then returned only the first-pass hits with a nil error. The search printed a partial list, or `No results` / `--fail-empty` exit 3, even though the retry had succeeded.

This matches `fetchMessageDetails`: keep the first concurrent error and stop. Empty IDs are still skipped, and only items with a non-empty ID are returned.

## What Problem This Solves

`gog gmail search QUERY` (default `--max 10`) calls `fetchThreadDetails` after `Threads.List`. One failed concurrent `Threads.Get` used to hide recovered threads instead of failing the command.

Call chain:

1. `gog gmail search QUERY`
2. `internal/cmd/gmail_search.go` calls `fetchThreadDetails`
3. Concurrent `Users.Threads.Get` for each listed ID
4. Old sequential retry discarded a successful Get body
5. Compact of `item.ID != ""` then dropped the recovered row

A user sees a short list, or `No results` with `--fail-empty` exit 3, after a transient thread-detail error that later succeeded.

The discard-body retry has been in `fetchThreadDetails` since at least [`4a840b5`](https://github.com/openclaw/gogcli/commit/4a840b5) (2026-03-08 file split, 193 days on main). Sibling `fetchMessageDetails` already returns `firstErr` and does not retry.

## Evidence

Before this change, one `Threads.Get` that failed once and then succeeded on the sequential retry returned two first-pass threads and a nil error:

```console
$ go test ./internal/cmd -count=1 -run TestFetchThreadDetails_ConcurrentError
--- FAIL: TestFetchThreadDetails_ConcurrentError (0.03s)
    gmail_concurrent_test.go:262: expected first concurrent error, got 2 items
FAIL
FAIL    github.com/openclaw/gogcli/internal/cmd 0.243s
FAIL
```

After this change, the same helper returns the first concurrent error and does not run the sequential retry:

```console
$ go test ./internal/cmd -count=1 -v -run TestFetchThreadDetails_ConcurrentError
=== RUN   TestFetchThreadDetails_ConcurrentError
--- PASS: TestFetchThreadDetails_ConcurrentError (0.00s)
PASS
ok  github.com/openclaw/gogcli/internal/cmd     0.168s
```

Full prefix after the patch:

```console
$ go test ./internal/cmd -count=1 -run TestFetchThreadDetails_
ok      github.com/openclaw/gogcli/internal/cmd 0.309s
```

Race run of the same prefix also passed (`go test -race ./internal/cmd -count=1 -run TestFetchThreadDetails_`). `make fmt-check` and `make lint` were clean locally.

## Real behavior proof

- **Behavior or issue addressed:** Concurrent `Threads.Get` failure in `gog gmail search` no longer drops recovered threads behind a nil error. The helper now returns that first error, same as `fetchMessageDetails`.
- **Real environment tested:** Windows 11, Go from `go.mod`, worktree at `C:/Users/sebta/AppData/Local/Temp/oc-pr-2026-09-17/gogcli-f025` on `fix/f025-thread-detail-retry` (`f717acd`).

  ```console
  go test ./internal/cmd -count=1 -v -run TestFetchThreadDetails_ConcurrentError
  go test ./internal/cmd -count=1 -run TestFetchThreadDetails_
  ```

- **Evidence after fix:** terminal output from the patched helper:

  ```console
  $ go test ./internal/cmd -count=1 -v -run TestFetchThreadDetails_ConcurrentError
  === RUN   TestFetchThreadDetails_ConcurrentError
  --- PASS: TestFetchThreadDetails_ConcurrentError (0.00s)
  PASS
  ok  github.com/openclaw/gogcli/internal/cmd     0.168s
  ```

  One failing `Threads.Get` now returns an error and nil items. The sequential discard-body retry is gone (three Gets for three IDs, empty ID still skipped).

- **Observed result after fix:** `fetchThreadDetails` returns the first concurrent Get error instead of a two-item partial list with a nil error.
- **What was not tested:** A live `gog gmail search` against a real Gmail account. That path needs operator OAuth and a real `Threads.Get` failure from Google.


steipete:
Landed as 0fcaeb949fd1b3572dcd2d65cd9c04a8be18271f. The helper and command-level HTTP 500 regressions failed against the old implementation; `make ci` passed on AWS Crabbox, independent review was clean through P2, and Linux, minimum-Go, macOS, Windows, worker, and Docker CI passed on `e5c7214b6e32ae0b2cca3c9db3c03084a90de729`. The command now returns the detail-fetch error without emitting partial JSON. No live Gmail account was used. Thanks @SebTardif!

PR #1142, opened 2026-09-20
fix(sheets): reject append responses without update metadata

SebTardif:
## What Problem This Solves

`gog sheets append` panics when Google returns HTTP 200 with no `updates` object. The copy-validation path already treats a missing `updates` field as an error, and `gog sheets tables append` already returns `append response missing update metadata`. Values append still printed `resp.Updates` on the success path, so an empty JSON body crashed the CLI.

## Evidence

httptest served `{}` for `spreadsheets.values.append`. Kong ran `sheets append s1 Sheet1!A1 --values-json [["a"]]`.

Before this patch, on `origin/main` `469b823`, the command process panicked:

```
go test ./internal/cmd -count=1 -run TestSheetsAppendMissingUpdates
--- FAIL: TestSheetsAppendMissingUpdates (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered, repanicked]
github.com/openclaw/gogcli/internal/cmd.(*SheetsAppendCmd).Run
    internal/cmd/sheets.go:630
```

After this patch, the same empty body returns a diagnostic and exit non-zero:

```
go test ./internal/cmd -count=1 -v -run TestSheetsAppendMissingUpdates
=== RUN   TestSheetsAppendMissingUpdates
--- PASS: TestSheetsAppendMissingUpdates (0.00s)
PASS
ok  github.com/openclaw/gogcli/internal/cmd
```

The recovered error text is `append response missing update metadata`. `--copy-validation-from` with `updates.updatedRange=""` still returns `append response missing updated range for validation copy`.

## Real behavior proof

- **Behavior or issue addressed:** Values append crashed on a successful API response that omitted `updates`. The CLI now returns `append response missing update metadata` instead of panicking, matching table-append.
- **Real environment tested:** macOS, Go 1.27.1, patched tree on `fix/f028-sheets-append-nil-updates`. Google was not called; httptest served the Sheets values.append JSON the CLI already uses in this package.

  Ran `go test ./internal/cmd -count=1 -run TestSheetsAppendMissingUpdates` on `origin/main` with the new probe copied in (panic at `sheets.go:630`). Then ran the same command on this branch (diagnostic, no panic). Also ran `TestSheetsAppendCopyValidationMissingRange` so `--copy-validation-from` still uses the updated-range message.

  ```
  === RUN   TestSheetsAppendMissingUpdates
  --- PASS: TestSheetsAppendMissingUpdates (0.00s)
  === RUN   TestSheetsAppendCopyValidationMissingRange
  --- PASS: TestSheetsAppendCopyValidationMissingRange (0.00s)
  PASS
  ```

  Empty `{}` now fails with `append response missing update metadata`. Empty `updatedRange` with `--copy-validation-from` still fails with `append response missing updated range for validation copy`.

- **Observed result after fix:** Kong recovered a nil `Updates` pointer as an error instead of SIGSEGV at `sheets.go:630`. Copy-validation still uses its original message when `updatedRange` is empty.
- **What was not tested:** A live Google Sheets append against a real spreadsheet. The panic is in post-response decoding, which httptest already drives through `SheetsAppendCmd.Run`.


steipete:
Landed as 3c4a7108f2ceec3eae7fe3cc92e19dd3a6370a7c. The old code panicked on a successful response with missing metadata. The expanded regression passes for absent/null metadata in human and JSON modes, with empty stdout and exactly one append. `make ci` passed on AWS Crabbox; independent review was clean through P2; all cross-platform CI passed on `d9732a842ee188e0b69518b57fded12dd051dfe0`. No live spreadsheet was changed. Thanks @SebTardif!

PR #1143, opened 2026-09-20
fix(docsbatch): print batch id in revision-mismatch errors

SebTardif:
## What Problem This Solves

When a Docs batch append fails because the document revision changed, the error says `batch=%s` but fills that field with `RequiredRevisionID` (the old revision), not the batch id. Operators looking up the queued transaction get the revision string instead of the UUID they need.

## Evidence

Create a batch, append with revision `rev1`, then append again with `rev2`.

Before this patch, on `origin/main` `469b823`, the mismatch error labeled `batch=` with the old revision:

```
append batch: document revision changed since the first request was queued (batch=rev1 current=rev2): document revision changed
```

That string does not contain the batch id `018f47b5-7b5e-7cc0-9a78-4a5bb1886251`.

After this patch, the same sequence prints the batch id in `batch=`:

```
go test ./internal/docsbatch -count=1 -run TestRepositoryAppendValidation
ok  github.com/openclaw/gogcli/internal/docsbatch  0.264s
```

The error now contains `batch=018f47b5-7b5e-7cc0-9a78-4a5bb1886251` and still wraps `ErrRevisionChanged`. `current=` remains the incoming revision (`rev2`).

## Real behavior proof

- **Behavior or issue addressed:** Docs batch revision-mismatch errors put the old revision in `batch=`. They now print `state.BatchID` so the queued transaction can be found.
- **Real environment tested:** macOS, Go 1.27.1, patched tree on `fix/f027-docsbatch-batch-id`. No Google Docs call; `Repository.Append` is the production queue path used by `gog docs batch`.

  Created a batch, appended with `rev1`, appended again with `rev2`. On main the error was `batch=rev1 current=rev2`. On this branch the same append prints `batch=` plus the batch UUID.

- **Evidence after fix:** terminal output from the patched append path:

  ```
  go test ./internal/docsbatch -count=1 -run TestRepositoryAppendValidation
  ok  github.com/openclaw/gogcli/internal/docsbatch
  ```

  The mismatch error includes `batch=` plus the batch UUID and no longer contains `batch=rev1`.

- **Observed result after fix:** `batch=` is the queued batch id. `current=` is still the incoming revision. The error still unwraps to `ErrRevisionChanged`.
- **What was not tested:** A live `gog docs batch` against a real Google Doc whose revision changed between appends. The format string is filled in `Repository.Append` before any Docs API write.


steipete:
Landed as b2411fa43e33187ff2cc31b45297f280df6361b1. The old implementation failed the revision-mismatch regression with `batch=rev1`; the fix reports the queued UUID and preserves the sentinel error. `make ci` passed on AWS Crabbox, independent review was clean through P2, and all cross-platform CI passed on `f4b1450dee3eb39f11557c5e5710196a30096a29`. No live document was changed. Thanks @SebTardif!

PR #1144, opened 2026-09-20
fix(calendar): surface Zoom rollback errors after failed writes

SebTardif:
## What Problem This Solves

`gog calendar create --with-zoom` (and update paths that create a Zoom meeting, then patch Calendar) can leave a Zoom meeting behind when the Calendar write fails. The CLI already calls `cancelZoomMeeting`, but it discards that error with `_ =`, so the leftover meeting is silent behind the Calendar 500.

`--remove-zoom` still warns and continues. That path is unchanged.

## Evidence

httptest returned HTTP 500 for calendar insert/patch after Zoom create. The fake Zoom client failed delete with `zoom delete denied`.

Before this patch, on `origin/main` `469b823`, Kong only reported the Calendar failure:

```
error = googleapi: got HTTP response code 500 with body: calendar insert failed
        , want rollback failure wrapping zoom delete denied
error = googleapi: got HTTP response code 500 with body: calendar update failed
        , want rollback failure wrapping zoom delete denied
```

After this patch, the same 500 plus failed Zoom delete is joined:

```
=== RUN   TestCalendarCreateCmd_WithZoomInsertFailureSurfacesRollbackError
--- PASS: TestCalendarCreateCmd_WithZoomInsertFailureSurfacesRollbackError (0.00s)
=== RUN   TestCalendarUpdateCmd_WithZoomPatchFailureSurfacesRollbackError
--- PASS: TestCalendarUpdateCmd_WithZoomPatchFailureSurfacesRollbackError (0.00s)
PASS
```

The error text includes `Zoom meeting rollback failed` and `zoom delete denied`. Zoom create count is 1 and delete was attempted once.

## Real behavior proof

- **Behavior or issue addressed:** Calendar insert/update discarded Zoom rollback errors. If Calendar fails after Zoom create, the operator now sees both the Calendar error and the Zoom delete failure.
- **Real environment tested:** macOS, Go 1.27.1, patched tree on `fix/f029-zoom-rollback-errors`. Calendar and Zoom HTTP were httptest fakes used by the existing calendar Zoom command path (`runKong` on `CalendarCreateCmd` / `CalendarUpdateCmd`).

  Ran `calendar create --with-zoom` against a Calendar POST that returns 500, with Zoom delete returning `zoom delete denied`. Repeated for `calendar update --with-zoom` where GET succeeds and PATCH returns 500.

  ```
  === RUN   TestCalendarCreateCmd_WithZoomInsertFailureSurfacesRollbackError
  --- PASS: TestCalendarCreateCmd_WithZoomInsertFailureSurfacesRollbackError (0.00s)
  === RUN   TestCalendarUpdateCmd_WithZoomPatchFailureSurfacesRollbackError
  --- PASS: TestCalendarUpdateCmd_WithZoomPatchFailureSurfacesRollbackError (0.00s)
  PASS
  ```

  Both errors contain `Zoom meeting rollback failed` and `zoom delete denied`.

- **Observed result after fix:** Create and update wrap `errors.Join(calendarErr, cancelErr)` instead of returning only the Calendar 500. `--remove-zoom` still warns and continues.
- **What was not tested:** A live Zoom API plus live Google Calendar write. The production `cancelZoomMeeting` call is the same helper used by regenerate/delete; this change only stops discarding its error on the failed-write rollback path.


steipete:
Landed as 4341ecead6119f4a8654cb0da68fd6c50abf0b5c. Completed the fix through the final CLI formatter: both Calendar and Zoom rollback failures are now visible while causes remain inspectable. The strengthened tests failed on the submitted implementation, then passed with the wrapper repair. Full `make ci` passed on AWS Crabbox, final staged review was clean through P2, and cross-platform CI passed on `72c01293fdcfa4fef06baa85c49856193b1d9aa2`. Synthetic failures only; no live event or meeting was changed. Thanks @SebTardif!

PR #1145, opened 2026-09-20
fix(api): pass request context through leftover Google Do calls

SebTardif:
## What Problem This Solves

Authenticated Google HTTP uses a 30s `ResponseHeaderTimeout` and no `Client.Timeout`. Command cancel and SIGINT only abort the round trip when the Google client call is bound with `.Context(ctx)`. YouTube still called `.Do()` without that, and Chat `GetSpaceReadState` / `Messages.Create` plus People `Get` did the same next to siblings that already pass context.

## Evidence

httptest handlers blocked on `<-r.Context().Done()`. Kong ran `youtube playlists list --mine`, `chat messages list spaces/aaa --unread`, and `people get people/123` with an already-canceled context.

Before this patch, on `origin/main` `469b823`, the YouTube list never saw the canceled context (`gensupport.SendRequest` with a nil context) and the command hung until the 8s deadline:

```
panic: test timed out after 8s
```

After this patch, the same canceled context returns immediately:

```
go test ./internal/cmd -timeout 8s -count=1 -run 'TestYouTubePlaylistsListHonorsCanceledContext|TestChatMessagesListUnreadHonorsCanceledContext|TestPeopleProfileGetHonorsCanceledContext'
ok  github.com/openclaw/gogcli/internal/cmd  0.700s
```

Each command error unwraps to `context.Canceled`. All 12 YouTube `.Do()` sites now chain `.Context(ctx)`.

## Real behavior proof

- **Behavior or issue addressed:** YouTube, Chat unread/send, and People get/relations ignored command cancel because `.Do()` had no request context. They now abort the HTTP round trip when the context is canceled.
- **Real environment tested:** macOS, Go 1.27.1, patched tree on `fix/f026-google-api-context`. Google was not called; httptest blocked until the request context ended, which is the same client path Kong uses for these commands.

  Started httptest handlers that wait on `r.Context().Done()`. Canceled the command context, then ran `youtube playlists list --mine`, `chat messages list --unread`, and `people get people/123`.

  ```
  go test ./internal/cmd -timeout 8s -count=1 -run 'TestYouTubePlaylistsListHonorsCanceledContext|TestChatMessagesListUnreadHonorsCanceledContext|TestPeopleProfileGetHonorsCanceledContext'
  ok  github.com/openclaw/gogcli/internal/cmd
  ```

  Each command returned `context.Canceled` instead of hanging until the 8s deadline.

- **Observed result after fix:** Canceled context reaches `gensupport.SendRequest`. YouTube, Chat GetSpaceReadState/Messages.Create, and People.Get stop waiting on a stalled peer.
- **What was not tested:** A live YouTube/Chat/People call against Google with Ctrl-C. Remaining `.Do()` sites outside this first PR are a follow-up sweep.


steipete:
Landed as 38dbdaac6dd91e3ccfe011433e9696994c63040a. Strengthened the regressions to cancel only after each YouTube/Chat/People request reaches the HTTP server. Old source waited two seconds for each fallback 504; the fixed calls return context.Canceled immediately. Full `make ci` passed on AWS Crabbox, independent review was clean through P2, and cross-platform CI passed on `0362d1a11460cf466e71d3077be94ee4d2dd6c8a`. Synthetic HTTP only; no Google account used. Thanks @SebTardif!

